PR #11, opened 2015-07-12
Geo import export

mgax:
@georust/core Conversion between GDAL (well, OGR actually) Geometry objects and our Rust-native `geo` types. This means we can read, and soon write, vector data in any format supported by GDAL.


frewsxcv:
Wow, this is great; LGTM :+1: :sparkles:

PR #14, opened 2015-07-14
Write geometries to new dataset

mgax:
Here's basic support for creating a new dataset and writing features. Just the geometries, no properties yet.


frewsxcv:
LGTM :shipit:

PR #17, opened 2016-05-05
Some work on the raster part

jdroenner:
Some time ago i did some work on the raster part of rust-gdal:

- added enums reflecting the c/c++ enums of GDAL
- added a GdalType trait
- added generic read/write methods (for all types implementing GdalType)
- replaced Point in the raster part with tuples (raster coordinates are always integer tuples)
- changed geo_transform to return an Option<[c_double; 6]>

( i also replaced all *const () with *const c_void to silence all the compiler warnings)


frewsxcv:
@mgax Any thoughts here? Otherwise I can review it later today

jdroenner:
I know this are many changes in one pull request. I did this for a project and forgot to create pull requests earlier. The next ones will be smaller ;)

mgax:
Cool! @frewsxcv, go ahead, otherwise I'll take a look over the weekend.

frewsxcv:
Just looked through all the commits, looks great! Thanks!

The only one I have reservations about is 780530968fd093355c0b3d1607993a6b52d8719a. In general, I'm in favor of making it explicit which coordinate we're dealing with. That said, I don't feel strongly. I think something like this could be middle-ground:

```rust
use std::convert::From;

pub struct Point<T> {
    pub x: T,
    pub y: T
}

impl<T> From<(T, T)> for Point<T> {
    fn from(t: (T, T)) -> Point<T> {
        Point {x: t.0, y: t.1}
    }
}
```

https://doc.rust-lang.org/std/convert/trait.From.html

So that one can do something like:

```rust
let rv = dataset.read_raster(
    1,
    (20, 30).into(),
    (2, 3).into(),
    (2, 3).into(),
);
```

Though it's not entirely necessary, just pointing out another option. If I don't hear anything from anyone by tomorrow night (EST), I'll merge this and deploy a new version.

Thanks again!

jdroenner:
You are right that the tuple has no explicit x and y axis which would be nice to have.
I removed the point because the vector part also uses tuples (e.g. "Geometry::get_point") or Point from rust-geo (which is not generic). Having a different Point type just for the raster part seemed a bit strange to me. Also raster read/write requires a point and two sizes so the name "Point" could be a bit misleading.

frewsxcv:
Sounds great. Thanks again!

frewsxcv:
Version 0.2.0 has just been released which includes these changes.

PR #19, opened 2016-05-17
vector tests: add "id" to ok_names

jdroenner:
The schema test fails for me with GDAL 2.x. It seams that "id" is now also reported as a part of the schema.

### DIFF
--- a/src/vector/tests/mod.rs
+++ b/src/vector/tests/mod.rs
@@ -129,7 +129,7 @@ fn test_schema() {
         .map(|f| f.name())
         .collect();
     let ok_names: Vec<String> = vec!(
-        "kind", "sort_key", "is_link", "is_tunnel",
+        "id", "kind", "sort_key", "is_link", "is_tunnel",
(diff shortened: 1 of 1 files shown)


jdroenner:
Ah i guess travis still uses GDAL 1.x :(

frewsxcv:
It looks like the latest version of GDAL that Ubuntu Trusty has is 1.x (Trusty is the latest version of Ubuntu that Travis CI supports)

http://packages.ubuntu.com/search?keywords=gdal&searchon=names&suite=trusty&section=all

Maybe it's time we create a `gdal-sys` crate that also compiles GDAL in the process which can be accomplished with a tool like https://github.com/crabtw/rust-bindgen . Not sure how feasible this is, haven't looked into it at all.

jdroenner:
I guess the switch from GDAL 1.x to 2.x will be a bit tricky anyway. On Windows (OSGeo4W) 2.x is also very new.

Maybe we can adapt the test to pass with or without "id"...

Issue #20, opened 2016-05-17
Accessing metadata (GDALMayorObject)


jdroenner:
`GDALMayorObject` is the base class for everything in GDAL: `GDALDataset`, `GDALRasterBand`and `OGRLayer`. Also since GDAL 2.0 the separation between raster and vector datasets was removed. Every dataset is now a `GDALDataset` (`OGRDataSource` is deprecated).

[GDALMayorObject ](http://www.gdal.org/classGDALMajorObject.html) provides methods to access metadata e.g. `GetMetadataItem`. Sadly there is no way to access metadata from rust-gdal.

My current idea to provide access to metadata for all `GDALMayorObject` based types is the following:
- Add a (non exported)  trait to access the c-pointer:

```
trait MajorObject {
    unsafe fn get_gdal_object_ptr(&self) -> *const c_void;
}
```
- Add a trait with default implementations (using the pointer) to access the metadata:

```
pub trait Metadata: MajorObject {
    fn get_metadata_item(&self, key: &str, domain: &str) -> Option<String> { ... }
    ...
}
```
- Implement MajorObject and Metadata for all matching types:

```
pub struct Dataset {
    c_dataset: *const c_void,
}

impl MajorObject for Dataset {
    unsafe fn get_gdal_object_ptr(&self) -> *const c_void {
        self.c_dataset
    }
}

impl Metadata for Dataset {}
```

I have a working implementation of this approach, but I would like to hear other ideas :)

frewsxcv:
This sounds great! Would love to see some metadata retrieval implemented for all the types. I can't think of any better ideas at the time for how to structure this, so feel free to open a PR for your changes. If we think of something better, we can always release a new version :)

PR #21, opened 2016-05-22
first version of metadata handling

jdroenner:
This is a first step to introduce metadata access: #20.

TODO: RasterBand metadata. This probably needs a `struct RasterBand`.

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -1,0 +1,41 @@
+use libc::{c_char, c_void, c_int};
+use utils::{_string};
+use std::ffi::{CString};
+use gdal_major_object::MajorObject;
+
+#[link(name="gdal")]
+extern {
+    fn GDALGetMetadataItem(hGdalMayorObject: *const c_void, pszName: *const c_char, pszDomain: *const c_char) -> *const c_char;
+    fn GDALSetMetadataItem(hGdalMayorObject: *const c_void, pszName: *const c_char, pszValue: *const c_char, pszDomain: *const c_char ) -> c_int;
+}
+
+pub trait Metadata: MajorObject {
+
+    fn get_metadata_item(&self, key: &str, domain: &str) -> Option<String> {
+        let c_key = CString::new(key.to_owned()).expect("Could not transform String to CString"); // FIXME: no unwrap
+        let c_domain = CString::new(domain.to_owned()).expect("Could not transform String to CString");
+
+        let c_res = unsafe { GDALGetMetadataItem(self.get_gdal_object_ptr(), c_key.as_ptr(), c_domain.as_ptr())};
+        if c_res.is_null() {
+            None
+        }
+        else {
+            Some(_string(c_res))
+        }
+    }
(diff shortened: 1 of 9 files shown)


frewsxcv:
> TODO: RasterBand metadata. This probably needs a struct RasterBand.

Are you planning to address this TODO in this PR or in a follow-up one? Either way is fine

jdroenner:
I'm not yet sure how much work this will require. If you want to merge this early i can do a follow-up ;)

src/metadata.rs
```diff
@@ -0,0 +1,41 @@
...
+extern {
+    fn GDALGetMetadataItem(hGdalMayorObject: *const c_void, pszName: *const c_char, pszDomain: *const c_char) -> *const c_char;
+    fn GDALSetMetadataItem(hGdalMayorObject: *const c_void, pszName: *const c_char, pszValue: *const c_char, pszDomain: *const c_char ) -> c_int;
+}
+
+pub trait Metadata: MajorObject {
+
+    fn get_metadata_item(&self, key: &str, domain: &str) -> Option<String> {
+        let c_key = CString::new(key.to_owned()).expect("Could not transform String to CString"); // FIXME: no unwrap
+        let c_domain = CString::new(domain.to_owned()).expect("Could not transform String to CString");
```
frewsxcv: could these both just `return None` if the `CString::new` fails?

src/metadata.rs
```diff
@@ -0,0 +1,41 @@
...
+        }
+        else {
+            Some(_string(c_res))
+        }
+    }
+
+    fn set_metadata_item(&mut self, key: &str, value: &str, domain: &str) -> Result<(), ()> {
+        let c_key = CString::new(key.to_owned()).expect("Could not transform String to CString"); // FIXME: no unwrap
+        let c_domain = CString::new(domain.to_owned()).expect("Could not transform String to CString");
+        let c_value =  CString::new(value.to_owned()).expect("Could not transform String to CString");
```
frewsxcv: Same thing here. `return Err(())` instead of `expect`?

src/metadata.rs
jdroenner: Good catch. A `&str` which can not be converted to a `CString ` can't be a valid key.

src/metadata.rs
jdroenner: Yes you are right. I missed this after changing the method to use `&str` instead of `CStr`.

src/metadata.rs
frewsxcv: Need any help making this changes? I can make them for you if you want, just let me know

jdroenner: I think i can do it till Friday :)

frewsxcv: Sounds good, no rush!

jdroenner:
I also added `GDALGetDescription ` and `impl Metadata for Driver`

frewsxcv:
Looks great, thanks!

Issue #25, opened 2016-12-13
split into gdal-sys and gdal crate


jdroenner:
Split rust-gdal into a gdal-sys and a (rust-) gdal crate.

(I'm currently working on this)

frewsxcv:
Sounds great! I've been meaning to do this for a while.

jdroenner:
@frewsxcv should i add the rust-sys repository to georust? ;)

frewsxcv:
I'd recommend having both `gdal` and `gdal-sys` library crates exist within the same repository. This way, if there are ever updates to `gdal-sys`, `gdal` will get updated to reflect the changes at the same time.

There a couple strategies for doing this:

### `gdal-sys` within `gdal`

```
/Cargo.toml (config for 'gdal')
/src/ (source files for 'gdal')
/gdal-sys/ (crate for 'gdal-sys')
/gdal-sys/Cargo.toml (config for 'gdal-sys')
/gdal-sys/src/ (source files for 'gdal-sys')
```

An example: https://github.com/alexcrichton/git2-rs

### `gdal-sys` and `gdal` separate w/ cargo workspace

cargo workspaces are relatively new: https://github.com/rust-lang/rfcs/blob/master/text/1525-cargo-workspace.md

```
/Cargo.toml (workspace config)
/gdal/ (crate for 'gdal')
/gdal/Cargo.toml (config for 'gdal')
/gdal/src/ (source files for 'gdal')
/gdal-sys/ (crate for 'gdal-sys')
/gdal-sys/Cargo.toml (config for 'gdal-sys')
/gdal-sys/src/ (source files for 'gdal-sys')
```

An example: https://github.com/servo/webrender

----------

Nowadays I prefer the second option

jdroenner:
I've been wondering how to do this and the examples are really helpful, thanks! Cargo workspaces look really nice.

PR #26, opened 2016-12-18
Move ffi into a seperate crate.

jdroenner:
#25:  moved ffi to a new crate: gdal-sys.
I also added a cargo workspace where the rust-gdal crate is the root and gdal-sys is a member.


frewsxcv:
[It looks like Travis found some errors](https://travis-ci.org/georust/rust-gdal/builds/184985550) when running `cargo test`. Some of the imports might need to be changed within `mod test`?

jdroenner:
It appears that the problem is `use gdal_sys::gdal;` in the raster mod and `use gdal_sys::ogr;` in the vector mod.
I'm using `use raster::{gdal, Driver, RasterBand};` in the other source files of the module which works with the nightly version of rust but not in stable and beta. What is the correct behavior? Using `pub use ...` in the module works fine with beta and stable but i don't think that the ffi should be public in rust-gdal.

I will put `use gdal_sys::gdal;` and `use gdal_sys::ogr;` directly into the source files to fix this.

PR #29, opened 2017-01-02
Allow to transform ogr geometries to other SRS

mthh:
Hi!

Thanks for providing an access to GDAL/OGR api in rust!
I was interested to transform ogr geometries to an other SRS so I wrapped some functions of the [OGR spatial reference](http://www.gdal.org/ogr__srs__api_8h.html) part of the gdal api (in order to be able to use *OGRSpatialReferenceH* and *OGRCoordinateTransformationH* types from C in rust).

It only uses a few major functions, as my main objective was simply to do something like :
```rust
let spatial_ref1 = SpatialRef::new_from_epsg(4326);
let spatial_ref2 = SpatialRef::new_from_epsg(3035);
let htransform = CoordTransform::new(&spatial_ref2, &spatial_ref1);
// an ogr Geometry :
geometry.transform(&htransform);
// or :
// geometry.transform_to(&spatial_ref1)
```

Could you be interested to merge these changes ? It also includes a simple example and some tests.

Also don't hesitate to tell me if there is some modifications requested in order to merge this (i'm not really experienced in rust) or if there is some other features of that API that you think should be wrapped.

Cheers!

### DIFF
--- a/src/spatial_ref/srs.rs
+++ b/src/spatial_ref/srs.rs
@@ -65,19 +65,25 @@ impl PartialEq for SpatialRef {
 }
 
 impl SpatialRef {
-    pub fn new() -> SpatialRef {
+    pub fn new(definition: &str) -> Result<SpatialRef> {
         let c_obj = unsafe { osr::OSRNewSpatialReference(ptr::null()) };
-        SpatialRef(c_obj)
+        if c_obj.is_null() {
+            return Err(ErrorKind::NullPointer("OSRNewSpatialReference").into());
+        }
+        let rv = unsafe { osr::OSRSetFromUserInput(c_obj, CString::new(definition).unwrap().as_ptr()) };
+        if rv != ogr_enums::OGRErr::OGRERR_NONE {
+            return Err(ErrorKind::OgrError(rv, "OSRSetFromUserInput").into());
+        }
+        Ok(SpatialRef(c_obj))
     }
 
     pub fn new_from_wkt(wkt: &str) -> Result<SpatialRef> {
         let c_str = CString::new(wkt).unwrap();
         let c_obj = unsafe { osr::OSRNewSpatialReference(c_str.as_ptr()) };
         if c_obj.is_null() {
-            Err(ErrorKind::NullPointer("OSRNewSpatialReference").into())
-        } else {
-            Ok(SpatialRef(c_obj))
+            return Err(ErrorKind::NullPointer("OSRNewSpatialReference").into());
         }
+        Ok(SpatialRef(c_obj))
     }
 
     pub fn new_from_epsg(epsg_code: u32) -> Result<SpatialRef> {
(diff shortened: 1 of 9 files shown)


frewsxcv:
Have time to review this @jdroenner? If not, I can

jdroenner:
I will look into it tomorrow.

jdroenner:
@mthh thank you for your great contribution.
There is just one thing: I'm going to merge #28 today or tomorrow. This will replace panicking and returning `Option` with `Result`. Would you adapt your pull request to this before we merge it?

mthh:
Sure, you can count on it! (today or tomorrow as well)

mthh:
I updated the code of this PR following your latests changes regarding error handling.

src/spatial_ref/srs.rs
```diff
@@ -0,0 +1,139 @@
...
+        } else {
+            Ok(SpatialRef(c_obj))
+        }
+    }
+
+    pub fn to_wkt(&self) -> Result<String> {
+        let mut c_wkt: *const c_char = ptr::null_mut();
+        let _err = unsafe { osr::OSRExportToWkt(self.0, &mut c_wkt) };
+        if _err != ogr_enums::OGRErr::OGRERR_NONE {
+            Err(ErrorKind::OgrError(_err, "OSRImportFromProj4").into())
```
jdroenner: The method name of the error should be "OSRExportToWkt".

src/spatial_ref/srs.rs
```diff
@@ -0,0 +1,139 @@
...
+        } else {
+            Ok(_string(c_wkt))
+        }
+    }
+
+    pub fn to_pretty_wkt(&self) -> Result<String> {
+        let mut c_wkt: *const c_char = ptr::null_mut();
+        let _err = unsafe { osr::OSRExportToPrettyWkt(self.0, &mut c_wkt, false as c_int) };
+        if _err != ogr_enums::OGRErr::OGRERR_NONE {
+            Err(ErrorKind::OgrError(_err, "OSRImportFromProj4").into())
```
jdroenner: "OSRExportToPrettyWkt"

src/spatial_ref/srs.rs
```diff
@@ -0,0 +1,139 @@
...
+        } else {
+            Ok(_string(c_wkt))
+        }
+    }
+
+    pub fn to_proj4(&self) -> Result<String> {
+        let mut c_proj4str: *const c_char = ptr::null_mut();
+        let _err = unsafe { osr::OSRExportToProj4(self.0, &mut c_proj4str) };
+        if _err != ogr_enums::OGRErr::OGRERR_NONE {
+            Err(ErrorKind::OgrError(_err, "OSRImportFromProj4").into())
```
jdroenner: "OSRExportToProj4"

src/spatial_ref/srs.rs
```diff
@@ -65,19 +65,25 @@ impl PartialEq for SpatialRef {
 }
 
 impl SpatialRef {
-    pub fn new() -> SpatialRef {
+    pub fn new(definition: &str) -> Result<SpatialRef> {
```
jdroenner: Is there no use-case for `new()` without parameters? Maybe the wrapped `OSRSetFromUserInput` should be a separate method like `new_from_definition(definition: &str)` or something like this?

mthh: You're right, there is use-cases for `new()` without parameters (I did this as a shortcut as currently the other exposed functions only allow to "import" a SRS, but it might be better to keep a method to create an empty Spatial Reference).

As you advised, it keeps now a `new()` method to create a spatial reference without parameters and a `new_from_definition()` using `OSRSetFromUserInput`.

src/spatial_ref/srs.rs
```diff
@@ -0,0 +1,171 @@
...
+impl SpatialRef {
+    pub fn new() -> Result<SpatialRef> {
+        let c_obj = unsafe { osr::OSRNewSpatialReference(ptr::null()) };
+        if c_obj.is_null() {
+            return Err(ErrorKind::NullPointer("OSRNewSpatialReference").into());
+        }
+        Ok(SpatialRef(c_obj))
+    }
+
+    pub fn new_from_definition(definition: &str) -> Result<SpatialRef> {
```
frewsxcv: This is not a big deal, but if you wanted to match the conventions in `std`, this would be renamed from `new_from_definition` to `from_definition`.

mthh: Thanks, I didn't know (but it clearly makes sense).
I guess it also apply to `from_wkt`, `from_epsg` and `from_proj4` ?
I will rename theses functions.

frewsxcv: To back up my previous comment, you can see some examples here:

https://doc.rust-lang.org/std/?search=from

(scroll down below the `from` method/function names)

mthh: This is indeed a significant list, thx (by looking around i also came across the RFC 430 which mention this `from_some_other_type` kind of constructor).

jdroenner:
I merged this with master. @mthh thanks again for your contribution :)

Issue #30, opened 2017-01-03
Allow to write ogr data


mthh:
I guess it would be nice to be able to write OGR data.

(I started to work on this if it's ok for you ?)

frewsxcv:
Sounds good to me! cc @jdroenner

jdroenner:
Sounds great :)

Issue #4, opened 2015-01-28
Borrowed vs. owned geometries


mgax:
The [OGR C API](http://www.gdal.org/ogr__api_8h.html) has two ways of accessing geometries:

* take a geometry pointer from a feature using `OGR_F_GetGeometryRef`, and you're not supposed to modify or deallocate it,
* clone the above geometry, or create one from scratch e.g. using `OGR_G_CreateFromWkt`, and you can modify it, and you're responsible for deallocating; you can also pass ownership back to OGR by attaching it to a feature, using `OGR_F_SetGeometryDirectly`.

In both cases, you have a pointer to `OGRGeometryH`. We should support operations on both, e.g. transforming them to JSON or WKT, and it makes sense to share the implementation.

A naive approach would be to always copy geometries, and only work with owned geometries from Rust, but that seems wasteful. But I'm probably implementing this initially, to have a working API.

Another option is to create different types for owned and borrowed geometries and share the implementation by a common trait. I've experimented with this in the [feature-geometry branch](https://github.com/georust/rust-gdal/compare/feature-geometry).

A third option is to have an owned geometry type, which wraps a borrowed geometry type, similar to `String` and `str`.

@georust/core, any input is much appreciated :)

skj3gg:
What would be the duplication/harm/opinions in having separate sub-packages? one for native or borrowed.

mgax:
> What would be the duplication/harm/opinions in having separate sub-packages? one for native or borrowed.

You mean, having completely separate types, that share nothing? It would be duplicating the implementation of a bunch of methods (exporting to json/wkt/wkb, testing for overlap/containment/intersection, computing buffer/centroid/hull/area/etc), and not allowing the API user to use them interchangeably, even though in OGR they're actually the same type.

sunng87:
For now we cannot avoid some package specific types. I think it's OK to keep these types and export them publicly. (Perhaps in a `raw` module?) 

We only need to provide To/From traits and impls to convert them into our geo types for integration.

frewsxcv:
> clone the above geometry, or create one from scratch e.g. using OGR_G_CreateFromWkt, and you can modify it, and you're responsible for deallocating; you can also pass ownership back to OGR by attaching it to a feature, using OGR_F_SetGeometryDirectly.

Admittedly I don't know too much about the OGR C API, but this approach sounds the most Rust-like

metasim:
See also: https://github.com/georust/gdal/discussions/360

PR #31, opened 2017-01-12
Extend the existing possibilities of writing ogr datasets.

mthh:
This PR (related to #30) aims to extend the existing possibilities of writing ogr datasets.

- It allows to set a definition schema (name, type, precision and width for each field) on a layer.
- It still allows to create a feature directly from geometry but now also with the various values to set on that feature.
- It adds three functions on the `Field` struct, allowing to retrieve the width, type and precision when reading a layer.
- It adds an `IntergerValue` to the `FieldValue` enum (allowing to get/set integer values too).
- It implements the `Clone` trait on the `Geometry` struct

I honestly don't know if I did it the right way or if there is too much redundancy in the way i implemented that (there is two commented examples (`read_write_ogr.rs` and `write_ogr.rs`) to facilitate your potential reviews).
Let me know if you have advices/requirements on how it should be designed (I should have asked for some guidance before doing it..!) and/or if it combines badly with the existing functionalities allowing to create/write ogr features.

For example in *src/vector/test/mod.rs* a dataset is currently written like this :
```rust
{
	let driver = Driver::get("GeoJSON").unwrap();
	let mut ds = driver.create(fixture!("output.geojson")).unwrap();
	let mut layer = ds.create_layer().unwrap();
	layer.create_feature(Geometry::from_wkt("POINT (1 2)").unwrap()).unwrap();
	// dataset is closed here
}
```

Now it could be possible to write :

```rust
{
	let driver = Driver::get("GeoJSON").unwrap();
	let mut ds = driver.create(fixture!("output.geojson")).unwrap();
	let mut layer = ds.create_layer().unwrap();
	layer.create_defn_fields(&[("Name", OFT_STRING), ("Value", OFT_REAL)]);
	layer.create_feature_fields(
		Geometry::from_wkt("POINT (1 2)").unwrap(),
		&["Name", "Value"],
		&[FieldValue::StringValue("Feature 1".to_string()), FieldValue::RealValue(45.78)]
		);
	// dataset is closed here
}
```


gdal-sys/src/ogr.rs
```diff
@@ -42,10 +49,17 @@ extern {
...
+    pub fn OGR_Fld_GetWidth(hDefn: *const c_void) -> c_int;
+    pub fn OGR_Fld_GetPrecision(hDefn: *const c_void) -> c_int;
+    pub fn OGR_Fld_SetWidth(hDefn: *const c_void, nNewWidth: c_int) -> c_void;
+    pub fn OGR_Fld_SetPrecision(hDefn: *const c_void, nNewPrecision: c_int) -> c_void;
+    pub fn OGR_Fld_Destroy(hDefn: *mut c_void) -> c_void;
     pub fn OGRFree(ptr: *mut c_void);
     pub fn VSIFree(ptr: *mut c_void);
 }
 
+pub const OFT_INTEGER:            c_int = 0;
```
jdroenner: I think it would be nice to change the OFT types into an enum and move it into ogr_enums.rs. This way all the FFI methods could take OgrFieldType instead of c_int as parameter. ([link to the OGR enum](http://www.gdal.org/ogr__core_8h.html#a787194bea637faf12d61643124a7c9fc))

src/vector/feature.rs
```diff
@@ -55,6 +71,52 @@ impl<'a> Feature<'a> {
...
+    pub fn create(&self, lyr: &Layer) -> Result<()> {
+        let rv = unsafe { ogr::OGR_L_CreateFeature(lyr.gdal_object_ptr(), self.c_feature) };
+        if rv != ogr_enums::OGRErr::OGRERR_NONE {
+            return Err(ErrorKind::OgrError(rv, "OGR_L_CreateFeature").into());
+        }
+        Ok(())
+    }
+
+    pub fn set_field_string(&self, field_name: &str, value: &str){
+        let c_str_field_name = CString::new(field_name).unwrap();
```
jdroenner: If it is possible please try not to use `unwrap` in library methods. Methods like this should return `Result<()>`.
I think we should add the returned error in errors.rs if we don't handle it already.

jdroenner:
@mthh this looks really nice :) I can have a closer look this weekend.

src/vector/defn.rs
```diff
@@ -26,6 +29,11 @@ impl Defn {
             total: total
         };
     }
+
+    pub fn new_from_layer(lyr: &Layer) -> Defn {
```
frewsxcv: This should be just `from_layer` if you wanted to match conventions in `std`.

src/vector/defn.rs
```diff
@@ -66,4 +74,16 @@ impl<'a> Field<'a> {
         let rv = unsafe { ogr::OGR_Fld_GetNameRef(self.c_field_defn) };
         return _string(rv);
     }
+
+    pub fn get_type(&'a self) -> i32 {
```
frewsxcv: This should be just `type` if you wanted to match conventions in `std`.

src/vector/defn.rs
```diff
@@ -66,4 +74,16 @@ impl<'a> Field<'a> {
...
+        unsafe { ogr::OGR_Fld_GetType(self.c_field_defn) }
+    }
+
+    pub fn get_width(&'a self) -> i32 {
```
frewsxcv: This should be just `width` if you wanted to match conventions in `std`.

src/vector/defn.rs
```diff
@@ -66,4 +74,16 @@ impl<'a> Field<'a> {
...
+        unsafe { ogr::OGR_Fld_GetWidth(self.c_field_defn) }
+    }
+
+    pub fn get_precision(&'a self) -> i32 {
```
frewsxcv: This should be just `precision` if you wanted to match conventions in `std`.

src/vector/feature.rs
```diff
@@ -87,4 +150,12 @@ impl FieldValue {
             _ => panic!("not a RealValue")
         }
     }
+
+    /// Interpret the value as `i32`. Panics if the value is something else.
+    pub fn as_int(self) -> i32 {
```
frewsxcv: Methods that consume `self` are usually named `into_<type>`, so in this case, it would be `into_int`, but I notice above that the other methods on this structure use `as_<type>`.

src/vector/defn.rs
mthh: I guess I can't use `type` as a function name as it is keyword in rust (that's why I chose `get_type` in replacement and thus to be consistent over the methods of this object I also used `get_width` and `get_precision`). Any suggestions ?

src/vector/defn.rs
mthh: Sure, I will change it.

gdal-sys/src/ogr.rs
mthh: I was also thinking about moving OFT types into an enum, so no problem, I will do it.

src/vector/feature.rs
mthh: Alright for returning `Result` on these methods.
However is it possible here to not use `unwrap()` ?

src/vector/defn.rs
jdroenner: As this will return a `OGRFieldType` you could change the method name to `field_type`.

src/vector/feature.rs
mthh: Should I also change the name of the two previous methods (`as_real` -> `to_real` and `as_string` `to_string` ?)

src/vector/defn.rs
mthh: Ok!

src/vector/feature.rs
jdroenner: Error handling in rust is a bit difficult. Where is a lot about [it in the book](https://doc.rust-lang.org/stable/book/error-handling.html).

To make this easier rust-gdal uses the [error-chain crate](https://docs.rs/error-chain/0.7.2/error_chain/). We already have some wrapping errors in errors.rs:
```
FfiNulError(::std::ffi::NulError);
StrUtf8Error(::std::str::Utf8Error);
```

To handle the `Ok `or return the (wrapped)  `Err` you can use the `try!{}` macro or `?`. In This case all you need to do is to replace `unwrap()` with `?`:  `let c_str_field_name = CString::new(field_name)?;` 
 `CString::new` returns `Result<CString, ::std::ffi::NulError>` in the error case `?` will convert `NulError` into the wrapping error (`FfiNulError`) and return it.

src/vector/feature.rs
jdroenner: i guess this would be the right thing to do. Also i think these methods should return an Option instead of panicking...

mthh: I made the other changes but not this last one (*returning an `Option` instead of panicking*), because I didn't know how to handle this when using a `FieldValue` to set a field :
```rust
pub fn set_field(&self, field_name: &str, type_value: OGRFieldType, value: FieldValue) -> Result<()> {
    match type_value {
        OGRFieldType::OFTReal => self.set_field_double(field_name, value.to_real()),
```
like here, when using  `value.to_real()`. I guess there is a way to handle this nicely without using `unwrap`, i will look into the `error_chain` crate documentation.

src/vector/feature.rs
```diff
@@ -35,15 +48,19 @@ impl<'a> Feature<'a> {
...
+            OGRFieldType::OFTReal => {
                 let rv = unsafe { ogr::OGR_F_GetFieldAsDouble(self.c_feature, field_id) };
                 return Some(FieldValue::RealValue(rv as f64));
             },
-            _ => panic!("Unknown field type {}", field_type)
+            OGRFieldType::OFTInteger => {
+                let rv = unsafe { ogr::OGR_F_GetFieldAsInteger(self.c_feature, field_id) };
+                return Some(FieldValue::IntegerValue(rv as i32));
+            },
+            _ => panic!("Unknown field type {:?}", field_type)
```
jdroenner: There should be an error for this :) . Maybe `UnhandledFieldType` ?

src/vector/feature.rs
```diff
@@ -55,6 +72,54 @@ impl<'a> Feature<'a> {
...
+    pub fn set_field_integer(&self, field_name: &str, value: i32) -> Result<()> {
+        let c_str_field_name = CString::new(field_name)?;
+        let idx = unsafe { ogr::OGR_F_GetFieldIndex(self.c_feature, c_str_field_name.as_ptr())};
+        unsafe { ogr::OGR_F_SetFieldInteger(self.c_feature, idx, value as c_int) };
+        Ok(())
+    }
+
+    pub fn set_field(&self, field_name: &str, type_value: OGRFieldType, value: FieldValue) -> Result<()> {
+        match type_value {
+            OGRFieldType::OFTReal => self.set_field_double(field_name, value.to_real()),
```
mthh: Shouldn't the `Result` from `set_field_double` (and others) just be passed to the caller of `set_field` like this?

```rust
pub fn set_field(&self, field_name: &str,  value: FieldValue) -> Result<()> {
      match value {
         FieldValue::RealValue(value) => self.set_field_double(field_name, value),
         FieldValue::StringValue(value) => self.set_field_string(field_name, value.as_str()),
         FieldValue::IntegerValue(value) => self.set_field_integer(field_name, value)
     }
 }
```
This way, with an invalid field name we get the `ErrKind::InvalidFieldName` error from the `set_field_double/string/..` function, and we can't enter an invalid FieldValue due to the type checking.

src/vector/geometry.rs
```diff
@@ -197,3 +217,11 @@ impl Drop for Geometry {
         }
     }
 }
+
+impl Clone for Geometry {
+    fn clone(&self) -> Geometry {
+        let c_geometry = self.c_geometry_ref.borrow();
+        let new_c_geom = unsafe { ogr::OGR_G_Clone(c_geometry.unwrap())};
```
mthh: Should I just check `self.has_gdal_ptr()` is true before allowing to get the `c_geometry` and performing the cloning ? I guess `clone` cannot return a `Result` so i don't know the preferred way to handle errors here, i will look into it.

mthh:
Many thanks for the detailed and instructive review!! I made most of the changes but i'm still unsure about 1 or 2 things.
I'll push it on github it in the day.

src/vector/feature.rs
jdroenner: Yes you are right this should be perfectly fine. I was thinking to complicated.

src/vector/geometry.rs
jdroenner: I think you are right that `clone` can not return a `Result`.  @frewsxcv can you help here?

mthh:
I guess I modified the various things discussed, except the clone method for which I did not know how to handle errors that might occur.

jdroenner:
This is great!
I think we could merge this and create an issue for `Geometry::clone`. We may need to change some things in the `Geometry `implementation to solve this.

mthh:
Great that it can be merged !
As I plan to dig a bit into rust (and to use rust-gdal) I will keep an eye on this future issue and see if can figure something out to improve some points of the `Geometry` implementation.

jdroenner:
Merged.
Actually there already is an issue for Geometry handling:  #4

PR #32, opened 2017-02-04
remove unwraps and add more error msgs.

jdroenner:
This removes most `.unwrap()` calls from library methods and updates dependency versions.
Additionally `ErrorKind::NullPointer` now tries to get a GDAL error msg.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -14,9 +14,9 @@ error_chain! {
             description("GDAL internal error")
             display("CPL error class: '{:?}', error number: '{}', error msg: '{}'", class, number, msg)
         }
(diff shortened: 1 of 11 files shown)


frewsxcv:
Any interest in reviewing this @mthh? Otherwise I can look at it sometime over the next couple days

jdroenner:
I guess we should also release a new version of rust-gdal after this is merged :)

mthh:
Sure, I will look on this today!
To be true, while I appreciate (really) being in the loop on this package and reviewing this, I might not be able to make some of the sharpest observations (speaking from a purely rust point of view, as I'm more coming from the GIS field and still not an accomplished rustacean!).

PR #33, opened 2017-02-25
Add build.rs to gdal-sys

jdroenner:
I added a `build.rs` to gdal-sys to enable building on windows. Building on windows requires  `$GDAL_HOME` to point to the base of the GDAL installation. The gnu toolchain works if `gdal_i.lib` is found in` $GDAL_HOME\lib` or if a `gdal{version}.dll` is found in `$GDAL_HOME\bin`. Msvc works only when the lib is available.

Additionally i moved the last FFI call to gdal-sys and silenced some warnings..


gdal-sys/build.rs
```diff
@@ -0,0 +1,50 @@
+
+
+fn main() {
+
+    let link_type = "dylib";
+    let lib_name = "gdal";
+    
+    #[cfg(target_family="windows")]
```
frewsxcv: https://doc.rust-lang.org/reference.html#conditional-compilation

You should be able to make this just `#[cfg(windows)]`

gdal-sys/build.rs
```diff
@@ -0,0 +1,50 @@
...
+                println!("cargo:rustc-link-search={}", bin_path.to_string_lossy());
+                println!("cargo:rustc-link-lib={}={}", link_type, format!("{}{}",lib_name, version));
+            }
+            else {
+                panic!("windows-gnu requires either gdal_i.lib in $GDAL_HOME\\lib OR gdal{version}.dll in $GDAL_HOME\\bin.");
+            }
+        }             
+    }
+
+    #[cfg(target_family="unix")]
```
frewsxcv: `#[cfg(unix)]`

PR #35, opened 2017-08-11
API extensions

pka:
Here are some API extensions I needed for my application. I'm working with GDAL 2.1, but it should still work with GDAL >= 1.11. Two tests are failing with 2.1 though.

### DIFF
--- a/gdal-sys/src/ogr.rs
+++ b/gdal-sys/src/ogr.rs
@@ -1,6 +1,8 @@
 use libc::{c_int, c_char, c_double, c_void};
 use ogr_enums::*;
 
+// http://gdal.org/ogr__api_8h.html
+
 extern {
     pub fn OGRRegisterAll();
     pub fn OGRGetDriverByName(pszName: *const c_char) -> *const c_void;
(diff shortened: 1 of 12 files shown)


frewsxcv:
do we need a newer version of gdal to test against? maybe we should bump travis to use ubuntu trusty and maybe it has a newer gdal?

jdroenner:
So far this looks good to me. @pka thanks for your contribution :)

The failures in "spatial_ref::tests::transform_ogr_geometry" and "vector::tests::test_schema" are caused by changes in GDAL 2.0:
- spatial_ref::tests::transform_ogr_geometry -> The precision of coordinates were extended, so the results are not equal.
- vector::tests::test_schema -> The "id" field is now (always?) part of the schema. 

@frewsxcv i think we should focus the development of rust-gdal on GDAL>= 2.0. There are many internal changes but the C API, which we use, wasn't changed much.
However we will only get GDAL >= 2.0 from the [ubuntugis ppa](https://launchpad.net/~ubuntugis/+archive/ubuntu/ppa). Zesty is the first Ubuntu version with GDAL 2.x...

frewsxcv:
here's a relevant thread for ubuntugis w/ travis: https://github.com/travis-ci/travis-ci/issues/2401

looks like we may need to modify our travis config to incorporate ubuntugis

jdroenner:
I changed the travis.yml to use GDAL 2.1 from ubuntugis-stable #36. However there is still one issue with one of the failing tests...

jdroenner:
@pka I adapted travis to use GDAL 2.1. Could you rebase/adapt your changes?

PR #36, opened 2017-08-13
adapted travis.yml to GDAL 2.x from ubuntugis-stable on trusty

jdroenner:


jdroenner:
I'm a bit lost with the fix for the "transform_ogr_geometry" test. I guess the issue is the precision of GDALs <OGR_G_ExportToWkt()>. As far as i know GDAL 2.1 always rounds the coordinates. With [GDAL 2.2 ](https://trac.osgeo.org/gdal/wiki/Release/2.2.0-News) there will be an option "OGR_WKT_PRECISION" to set the precision. So i guess we have to round the expected coordinates to make the test green...

frewsxcv:
i'm not too familiar with gdal and don't keep up with the versioning/changes, so I can't provide much help here. if you manage to get it working how you think it should, feel free to merge

PR #41, opened 2017-10-11
Add a few area-related wrappers

turboladen:
My first official foray into FFI with Rust, so I may have missed some stuff.

FWIW, I wrote the Ruby gem [ffi-gdal](https://bitbucket.org/agrian/ffi-gdal) (which may or may not be well done, but either way, I've spent hours with the stuff)--I dunno if that could serve as an aid in wrapping some of stuff that's not yet done, but I thought I'd post just in case it helps in some way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,8 @@
 /target
 /.vscode
 /gdal-sys/target
+
+# gtags
+GPATH
+GTAGS
+GRTAGS
--- a/gdal-sys/build.rs
+++ b/gdal-sys/build.rs
@@ -4,7 +4,7 @@ fn main() {
 
     let link_type = "dylib";
     let lib_name = "gdal";
-    
+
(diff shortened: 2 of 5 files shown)


turboladen:
Hmm... sorry for the ignorance, but I see tests are failing to find the new `gdal-sys` functions: what's the best way to tell `gdal` to use the new stuff?

jdroenner:
Hi, 
thanks for your contribution.

I changed Cargo.toml to use gdal-sys from crates.io for the latest release. I will change it back to use  the sub-crate directly.

turboladen:
@jdroenner sounds good, thanks!

jdroenner:
@turboladen something was wrong with the version string in Cargo.toml. Now gdal-sys from the repository is used again.

turboladen:
@jdroenner aha! Ok, thank you!

PR #43, opened 2017-10-11
Add a non panicing version of CoordTransform::transform_coord

weiznich:


weiznich:
> Also could we use the OGRError for this and/or get more information from GDAL similar, to the methods in src/utils.rs?

I did not found any function to gain more informations from gdal why the transformation is not possible.

jdroenner:
I will look a bit further into this. It should be possible to use `utils::_last_cpl_error` to get more information from GDAL. But i'm not sure if this is always the case.

This code:
```
    let wgs84 = SpatialRef::from_epsg(4326).unwrap();
    let webmercator = SpatialRef::from_epsg(3857).unwrap();

    let mut x = [1000000.0];
    let mut y = [1000000.0];
    let mut z = [0.0, 0.0];

    let trafo = CoordTransform::new(&wgs84, &webmercator).unwrap();
    let r = trafo.transform_coord(&mut x, &mut y, &mut z);
    println!("{:?}", r);
    let err = _last_cpl_err(CPLErr::CE_Failure);
    println!("{}", err);
```

creates:

```
CPL error class: 'CE_Failure', error number: '1', error msg: 'latitude or longitude exceeded limits'
```

However, your example creates no CPL errors :(

weiznich:
Any news on this?

jdroenner:
Let's merge this and adapt it if needed :)

PR #44, opened 2017-10-25
Add GDAL library config module

warwickm:


jdroenner:
Hi, thanks for your contribution.
Could you replace the `unwrap()` calls with `try`/`?` ?

warwickm:
I am finding that running cargo test locally fails (with a segfault) quite frequently and I see the Travis CI build also has problems.  I get the same failures without my changes too.

Running the tests using valgrind indicates a read-after-free problem in the last line of `Geometry::test_spatial_reference` where it calls `geom.spatial_reference()`. Valgrind is suggesting that some memory is being read after it has been freed after the previous line.

If I comment out the `test_spatial_reference` test, all the other tests pass and there are no segfaults.

I will continue to investigate, but I am confident that this PR is not the cause.

jdroenner:
This is interesting. Travis started to fail often some time ago but i had no idea why. Could you create a new issue for this?

Issue #45, opened 2017-10-28
Unit tests frequently fail with a segfault


warwickm:
Running ``cargo test`` frequently fails with a segfault.

Running the tests using valgrind indicates a read-after-free problem in the last line of ``Geometry::test_spatial_reference`` where it calls ``geom.spatial_reference()``. Valgrind is suggesting that some memory is being read after it has been freed after the previous line.

Commenting out the ``test_spatial_reference`` test makes all the other tests pass reliably and there are no segfaults.

Perhaps ``srs`` is being dropped on exit from ``set_spatial_reference`` because it is being moved into the function.

Full valgrind output follows:

```
$ valgrind target/debug/deps/gdal-70427d28ae09def0 --test-threads 1
==14812== Memcheck, a memory error detector
==14812== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==14812== Using Valgrind-3.10.1 and LibVEX; rerun with -h for copyright info
==14812== Command: target/debug/deps/gdal-70427d28ae09def0 --test-threads 1
==14812== 

running 56 tests
test raster::tests::test_create ... ok
test raster::tests::test_create_copy ... ok
test raster::tests::test_create_with_band_type ... ok
test raster::tests::test_geo_transform ... ok
test raster::tests::test_get_band_type ... ERROR 5: GDALDataset::GetRasterBand(2) - Illegal band #

ok
test raster::tests::test_get_dataset_driver ... ok
test raster::tests::test_get_description ... ok
test raster::tests::test_get_driver_by_name ... ok
test raster::tests::test_get_metadata_item ... ok
test raster::tests::test_get_no_data_value ... ok
test raster::tests::test_get_offset ... ok
test raster::tests::test_get_projection ... ok
test raster::tests::test_get_raster_count ... ok
test raster::tests::test_get_raster_size ... ok
test raster::tests::test_get_rasterband ... ok
test raster::tests::test_get_scale ... ok
test raster::tests::test_open ... ERROR 4: No such file or directory
ok
test raster::tests::test_read_full_raster_as ... ok
test raster::tests::test_read_raster ... ok
test raster::tests::test_read_raster_as ... ok
test raster::tests::test_set_metadata_item ... ok
test raster::tests::test_write_raster ... ok
test spatial_ref::tests::authority ... ok
test spatial_ref::tests::comparison ... ok
test spatial_ref::tests::from_epsg_to_wkt_proj4 ... ok
[...]

PR #47, opened 2017-11-02
A better fix for issue #45

warwickm:
This reverts the previous fix and changes the impl of Drop for SpatialRef
to use OSGRelease instead of OSRDestroySpatialReference. This will just
decrement the reference count for the SpatialRef and destroy it if the
count reaches 0.

This resolves a valgrind error that was still being reported after the previous
fix. This fix also has the advantage that it does not change the API for
existing users.


jdroenner:
looks good. thanks!

PR #49, opened 2018-02-05
RasterBand to Ndarray

austinstig:
I added a few functions mostly to RasterBand to support a read_band_as_array method on the RasterBand API. The user must still pass read_band_as_array a type 'U' argument that will return an ndarray::Array2<U> after unwrapping the result. I do not know how to handle the C types on the GDALDataType. I also added a few libraries to support this effort and extended the errors.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,6 +24,9 @@ extern crate libc;
 extern crate geo;
 extern crate gdal_sys;
 extern crate num_traits;
+extern crate ndarray;
+extern crate num;
+extern crate byteorder;
 
 #[macro_use]
 extern crate error_chain;
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -6,6 +6,14 @@ use gdal_major_object::MajorObject;
 use metadata::Metadata;
 use gdal_sys::{gdal, cpl_error};
 use utils::{_last_cpl_err};
+use ndarray::Array2;
+use num::FromPrimitive;
+use std::any::TypeId;
+use byteorder::LittleEndian;
+use byteorder::ReadBytesExt;
+use super::gdal_enums::GDALDataType;
+use errors::ErrorKind::ConversionError;
+use std::io::Cursor;
 
 use errors::*;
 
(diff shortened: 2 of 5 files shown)


src/lib.rs
```diff
@@ -24,6 +24,9 @@ extern crate libc;
 extern crate geo;
 extern crate gdal_sys;
 extern crate num_traits;
+extern crate ndarray;
+extern crate num;
+extern crate byteorder;
```
jdroenner: do we really need num and byteorder for this?

src/raster/rasterband.rs
```diff
@@ -169,3 +183,85 @@ impl<'a> MajorObject for RasterBand<'a> {
 }
 
 impl<'a> Metadata for RasterBand<'a> {}
+
+/// # Read To Array
+/// A helper function that takes a raster band as input. It will
+/// read the rasterband into an ndarray.
+fn read_to_array<T: 'static + Copy + FromPrimitive>(rband: &RasterBand) -> Result<Array2<T>> {
```
jdroenner: Since GDALDataType is now public, i think `T: GDALDataType`  could be enough here.

src/raster/rasterband.rs
```diff
@@ -6,6 +6,14 @@ use gdal_major_object::MajorObject;
 use metadata::Metadata;
 use gdal_sys::{gdal, cpl_error};
 use utils::{_last_cpl_err};
+use ndarray::Array2;
+use num::FromPrimitive;
```
jdroenner: if only FromPrimitive is needed, you can import num-traits.

src/raster/rasterband.rs
```diff
@@ -169,3 +183,85 @@ impl<'a> MajorObject for RasterBand<'a> {
...
 impl<'a> Metadata for RasterBand<'a> {}
+
+/// # Read To Array
+/// A helper function that takes a raster band as input. It will
+/// read the rasterband into an ndarray.
+fn read_to_array<T: 'static + Copy + FromPrimitive>(rband: &RasterBand) -> Result<Array2<T>> {
+    // get the data type of the dataset
+    let gt: GDALDataType = rband.band_type();
+    // get the byte buffer of the full dataset
+    let buffer: Buffer<u8> = rband.read_band_as::<u8>()?;
```
jdroenner: why do we need to read the data as `Buffer<u8>`?  `T: GDALDataType` allows reading the correct type.

src/raster/rasterband.rs
```diff
@@ -169,3 +183,85 @@ impl<'a> MajorObject for RasterBand<'a> {
...
+    // get the data from the byte buffer
+    let shape: (usize, usize) = buffer.size;
+    let data = buffer.data;
+    // convert the bytes of the buffer to type T
+    return match gt {
```
jdroenner: i don't think the extract method is needed... With `T: GDALDataType` mapping the `Buffer<T>` into an Array directly should be possible. Maybe `GDALDataType` can even stay private similar to the read_as_buffer methods.

PR #51, opened 2018-02-14
Wrap OGR_L_GetExtent

lnicola:
I'm not sure whether it's all right to expose `OGREnvelope` directly, any preferences?


jdroenner:
Thanks for your contribution :)
I guess it is ok to expose `OGREnvelope` since it is exposed by gdal-sys anyway.

PR #52, opened 2018-02-14
Wrap OGR_L_GetSpatialRef, OSRAutoIdentifyEPSG, fix return type

lnicola:


gdal-sys/src/ogr_enums.rs
```diff
@@ -36,3 +36,6 @@ pub enum OGRFieldType {
     OFTInteger64List = 13,
     OFTMaxType = 14
 }
+
+pub const FALSE: i32 = 0;
+pub const TRUE: i32 = 1;
```
jdroenner: Is the `i32 `always correct or is this a `c_int`?
Maybe we cloud call it `C_TRUE `and `C_FALSE` ?

lnicola: Good point. Bikeshed: maybe `OGR_TRUE`?

jdroenner: well it's not really a OGR constant... So i would prefer the C prefix.

lnicola: `CPL_` would have been the best :smile:

jdroenner: GDAL prefixes are a bit confusing sometimes 😄

lnicola:
@jdroenner Can we mergethis as-is?

PR #55, opened 2018-02-16
Use bindgen for gdal-sys

### LINKED ISSUE #50: Consider using bindgen for gdal-sys
What the title says. This might make adding the missing APIs easier, but will probably require a large rework of the crate.

Do you think it might be a good idea?

I'm not necessarily volunteering to implement this 😄 .

lnicola:
Closes #50 
Closes #56 

TODO: 

- [x] figure out how to find the GDAL include path
- [x] clean up `bindgen` config
- [x] choose between `libc` and `std::os::raw`
- [x] see if it makes sense to split the bindings into sub-modules
- [x] decide whether to use the modules for enums
- [x] use modules for the remaining enums
- [x] pick again between `1` and `TRUE`
- [x] clean up the pointer casts
- [x] bring back re-exports in submodules
- [x] see where the generated file should live
- [x] figure out the Windows story
- [x] decide whether to use `bindgen` or commit the generated file
- [x] decide whether to use the system GDAL headers
- [x] figure out why my headers are missing parameter names
- [x] try to use the generated handle types instead of `*mut c_void`
- [x] consider replacing `WkbType` with `OGRwkbGeometryType`
- [x] decide whether we _really_ want to do this
- [x] do something about `OGRErr`
- [x] look into `i32` vs. `u32` for enums
- [x] allow setting the include and link paths
- [x] allow static GDAL builds
- [x] update `gdal-sys` build instructions
- [x] bump version


jdroenner:
This is really great work :)

It was not easy to get the crate building on Windows with mingw as well as the MS tool-chain. So please make sure that it still builds on Windows. If you don't have a Windows system around, just tell me and i will test it.

lnicola:
Thanks, but I'm pretty sure it's no longer working. Unfortunately, I don't have a way to test it, neither on MSVC, nor in MinGW. So it would be good if you could look into it.

jdroenner:
No problem. I will test it later today  :)

jdroenner:
Building works but the tests won't run. Will try to fix it tomorrow....

jdroenner:
It appears to be a problem with my machine or windows in general. The test are not running on master, too :(

lnicola:
Are you getting any error, either as a message box or in Event Viewer? Maybe it can't find the GDAL DLL?

In the meanwhile, do you have any thoughts about the remaining TODO items?

jdroenner:
> see if it makes sense to split the bindings into sub-modules

I don't know if it makes sense to split the bindings. What modules would that be? Since GDAL 2.x  OGR and GDAL are merging...

> decide whether to use bindgen or commit the generated file

This is a bit more complicated. I guess we can do a new release for new GDAL releases and running bindgen is not really required for each build...

> consider replacing WkbType with OGRwkbGeometryType

This would be the way to go. I think we already use all other enums.

> decide whether we really want to do this

we will never be able to cover all of GDAL if we create the bindings by hand

lnicola:
> I don't know if it makes sense to split the bindings. What modules would that be? Since GDAL 2.x OGR and GDAL are merging...

I didn't realize this.

> I guess we can do a new release for new GDAL releases and running bindgen is not really required for each build...

Or we can do both with a feature. But I'm worried abut compatibility. Will old binding work with a newer GDAL? Will new bindings work with an older GDAL as long as the application isn't calling any missing method? In the latter case, won't the OS loader complain, at least on Windows?

I looked at `libsqlite3-sys`, and it does this: https://github.com/jgallagher/rusqlite/blob/master/libsqlite3-sys/Cargo.toml#L13.

jdroenner:
Maybe we should go with the features approach. [PyGDAL]( https://github.com/nextgis/pygdal) has a binding for every release but i'm not yet convinced that this is required. GDAL releases are very conservative in my experience.

jdroenner:
surprisingly it works with the gnu toolchain on windows on different machine.
It requires installing clang (that was expected) and adding `.clang_arg("-IC:\\OSGeo4W64\\include")` to the bindgen builder.

The MSVC toolchain does not work but i think i can figure it out with a bit more time.
`C:\OSGeo4W64\include/cpl_port.h:137:10: fatal error: 'stdio.h' file not found`

lnicola:
> surprisingly it works with the gnu toolchain on windows on a different machine.
It requires installing clang (that was expected) and adding .clang_arg("-IC:\\OSGeo4W64\\include") to the bindgen builder.

Maybe we could bundle at least _some_ bindings, so people don't have to deal with that?

> The MSVC toolchain does not work but i think i can figure it out with a bit more time.

Ouch. So `clang` is not finding the VS include path.

jdroenner:
I just reinstalled the VS build tools and now it also works with the MSVC tool chain 😄

jdroenner:
Uuh this was really obvious...

You have to run the ` #[cfg(windows)]` part regardless if you are using bindgen or use pre-generated bindings.
We should move it into a special method `prepare_windows` and call it first in the `main` method ` #[cfg(windows)] prepare_windows();`.

lnicola:
I've been trying to make this work for multiple versions of GDAL, but I ran into https://github.com/rust-lang-nursery/rust-bindgen/issues/1267.

jdroenner:
great! could you add the configurable include paths? I guess this will also allow to build on windows.

lnicola:
Yes, I'll try to look into it tomorrow.

lnicola:
@jdroenner, @weiznich, can you take a look at https://github.com/georust/rust-gdal/pull/55/commits/4ae3da628968d26dd2d49b15d0b5bb75ebcf0ba7?

jdroenner:
i will try it later today when i have access to a windows pc.

jdroenner:
all i get is "error: could not find native static library `gdal_i.lib`, perhaps an -L flag is missing?"

lnicola:
Can you add a print or `panic!` [here](https://github.com/georust/rust-gdal/pull/55/commits/8985421a751a4d0f03dc84ac812f8d5a6838f023#diff-9a32350d33992b5b57b08e6c07b0691cR145) to see what path it picks?

lnicola:
Ah, I think it uses `gdal_i.lib` instead of `gdal_i`.

jdroenner:
Its working now 👍 Great work!

lnicola:
Okay, so I've gone through my TODO list.

jdroenner:
This is really great 👍 Thanks a lot!
I guess we need to release a new version of rust-gdal as well :)

lnicola:
As a test, I tried to port https://github.com/t-rex-tileserver/t-rex to the new version. It was trivial, but I had to add a reference to `gdal-sys` because `gdal_sys::OGRwkbGeometryType` isn't re-exported.

Should we re-export these? It might be nicer for the users.

lnicola:
Oh, it actually is. Please ignore my last comment.

lnicola:
For the record, the pre-built files can be obtained with:

```
bindgen --constified-enum-module ".*" --ctypes-prefix libc --whitelist-function "(CPL|GDAL|OGR|OSR|OCT|VSI).*" wrapper.h
```

PR #56, opened 2018-02-23
Small buildscript improvements

weiznich:
* Allow to static link libgdal on linux
* Allow to set libary path on linux


jdroenner:
We are currently switching gdal-sys to bindgen. This will have to wait until we merged #55.

PR #63, opened 2018-07-17
Add method to set no data values for raster bands

davebrent:
Currently you can retrieve a no data value for a raster, but there doesn't exist a method to set it. Thought this might be useful for someone else other than myself.

### DIFF
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -125,6 +125,14 @@ impl <'a> RasterBand<'a> {
         None
     }
 
+    pub fn set_no_data_value(&self, no_data: f64) -> Result<()> {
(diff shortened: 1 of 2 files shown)


src/raster/rasterband.rs
```diff
@@ -125,6 +125,11 @@ impl <'a> RasterBand<'a> {
         None
     }
 
+    pub fn set_no_data_value(&self, no_data: f64) -> Result<()> {
+        unsafe { gdal_sys::GDALSetRasterNoDataValue(self.c_rasterband, no_data) };
```
jdroenner: GDALSetRasterNoDataValue can return a CPL_Error. Please check if the call failed and if it does return an error.
I guess you can look at the ` pub fn write<T: GdalType+Copy>` method and copy the error handling from there :)

davebrent: Yup will do :+1:

PR #67, opened 2019-03-03
Allow reading block-size of Rasters

rmanoka:
This PR allows efficient streaming of raster by reading (using already existing `read_as`) along block boundaries.  To get block size, call `block_size` on a `RasterBand`.  

Additionally, we also add feature flag `gdal_2_2` that enables the corresponding feature in `gdal-sys` crate to pull in features from GDAL >= 2.2.  Using this, we provide `actual_block_size` function behind this feature gate to be used if required.

List of what is implemented:

1. implemented GetBlockSize
2. added feature flag "gdal_2_2" to enable recent GDAL functions.
3. implemented GetActualBlockSize (behind "gdal_2_2" flag)


pka:
Thanks for your contribution, @rmanoka! Code looks good to me. Would you mind adding some tests for the new functions and resolve the conflicting files?

rmanoka:
@pka Sure, I'll add tests and resolve `Cargo.toml` conflict over the weekend.

There are a couple of minor issue with #68 that has been merged recently:

1. The PR introduces `block_size` in RasterBand, similar to this PR.  The code however is identical, so I will fix it in when updating this PR.

2. However, `read_block` function added by the PR seems to have a bug.  Suppose the raster dimensions are not perfectly divisible by the block-size, then the right, and bottom edge blocks will not have the size as provided by `block_size` (as it will be smaller).   This was the reason why we had to include `actual_block_size` in this PR.  To the best of my understanding, the `read_block` function would error when reading these edge blocks.  The fix is to use `actual_block_size` in the beginning of `read_block` implementation;  however, the two functions are under different feature gates, so I don't know how to proceed here.

Any ideas?

rmanoka:
Actually, the [GDAL docs](https://www.gdal.org/classGDALRasterBand.html#ad80cecc562fd48f5783677de625360ac) say it already yields an array of block-size dimensions; so the user only needs to use `actual_block_size` to correctly use the output from `read_block`.  Seems we should be in good shape with present implementation.  Sorry about the confusion.

src/raster/rasterband.rs
```diff
@@ -42,6 +40,31 @@ impl <'a> RasterBand<'a> {
         (size_x as usize, size_y as usize)
     }
 
+    /// Get x-size of the band
+    pub fn x_size(&self) -> usize {
+        let out;
```
lnicola: Nit: `let out = unsafe { gdal_sys::GDALGetRasterBandXSize(self.c_rasterband); }`

lnicola:
Ping @pka?

pka:
`cargo doc` fails on nightly. Will hopefully work again in a few days?

lnicola:
Works for me. Let's wait for the `master` [build](https://travis-ci.org/georust/gdal/builds/532299563) to finish.

PR #68, opened 2019-03-10
RasterBand to Ndarray, with failure.

rochamatcomp:


jdroenner:
That's really great! I wanted to add something like that some time ago.
Maybe the ndarray functions could be placed inside a feature?

rochamatcomp:
I placed the imports, functions and tests inside the feature `array`. It's ok like that?

jdroenner:
looks good. thanks :)

PR #72, opened 2020-02-11
read/write support for date und datetime fields

nmandery:
This PR adds read and write support for Date and Datetime fields in the OGR API. An example is provided.

This also brings in the `chrono` crate as an additional dependency.


jdroenner:
That looks great! Thank you for this contribution.
Could you make the feature and therefore the chrono dependency optional? Like for ndarray.

nmandery:
Sure, I added one more commit containing the feature gate.

PR #73, opened 2020-04-04
add raster::Driver create options

hamersaw:
adding an options parameter allows passing creation options. For example "COMPRESS=LZW" enables LZW compression for GTiff driver files.

I haven't looked into the rest of the repository, but it seems this construct could replace null_mut() in a few places. The only downside, the c_options vector lifetime needs to extend through the gdal_sys function call since it's a mutable reference. If it is dropped before the call the result is a segmentation fault.

### DIFF
--- a/src/raster/driver.rs
+++ b/src/raster/driver.rs
@@ -62,13 +62,15 @@ impl Driver {
         filename: &str,
         size_x: isize,
         size_y: isize,
-        bands: isize
+        bands: isize,
+        options: Option<Vec<&str>>,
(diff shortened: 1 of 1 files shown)


frewsxcv:
@hamersaw did you mean to close this?

hamersaw:
@frewsxcv yes, i found an issue. it still seg faults in a random case that I'm trying to figure out the cause of. Is there a better way to handle this?

Additionally, I think it should be applied to all situations instead of passing null_mut() to instances of **char.

PR #78, opened 2020-06-25
Define Geometry Intersection trait and implement it for Geometry

boydjohnson:
I built off of #75 since it is already approved. 

gdal-sys has vector methods for both layers and geometries. It would be nice if these operations could use one trait, but unfortunately the intersection function signature for layers and geometries is much different.

Seeking comments on the location for these traits. I went with `crate::vector::ops::geometry::intersection`.

### DIFF
--- a/src/vector/geometry.rs
+++ b/src/vector/geometry.rs
@@ -43,7 +43,7 @@ impl Geometry {
         *(self.c_geometry_ref.borrow_mut()) = Some(c_geometry);
     }
 
-    unsafe fn with_c_geometry(c_geom: OGRGeometryH, owned: bool) -> Geometry {
+    pub(crate) unsafe fn with_c_geometry(c_geom: OGRGeometryH, owned: bool) -> Geometry {
         Geometry {
             c_geometry_ref: RefCell::new(Some(c_geom)),
             owned,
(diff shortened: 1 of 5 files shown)


src/vector/geometry.rs
```diff
@@ -43,7 +43,7 @@ impl Geometry {
         *(self.c_geometry_ref.borrow_mut()) = Some(c_geometry);
     }
 
-    unsafe fn with_c_geometry(c_geom: OGRGeometryH, owned: bool) -> Geometry {
+    pub unsafe fn with_c_geometry(c_geom: OGRGeometryH, owned: bool) -> Geometry {
```
jdroenner, line 46: I'm not sure if we want to pub this method. Could you change it to pub(crate) ?

boydjohnson: Updated.

src/vector/ops/geometry/intersection.rs
```diff
@@ -0,0 +1,74 @@
...
+    ///
+    /// Generates a new geometry which is the region of intersection of
+    /// the two geometries operated on. Call intersects (Not yet implemented)
+    /// to check if there is a region of intersection.
+    /// Geometry validity is not checked. In case you are unsure of the
+    /// validity of the input geometries, call IsValid() before,
+    /// otherwise the result might be wrong.
+    ///
+    /// # Returns
+    /// Ok(Geometry) if the two geometries intersect
```
michaelkirk, line 19: method returns an `Option<Geometry>`, but this documentation shows as if it returns a `Result<Geometry>`

boydjohnson: Updated. And added another unit test.

PR #80, opened 2020-07-07
Add support for Integer64

### LINKED ISSUE #79: OFTInteger64 is not supported

lnicola:
Fixes #79

### DIFF
--- a/src/vector/feature.rs
+++ b/src/vector/feature.rs
@@ -3,7 +3,11 @@ use crate::vector::geometry::Geometry;
 use crate::vector::layer::Layer;
 use crate::vector::Defn;
 use gdal_sys::{self, OGRErr, OGRFeatureH, OGRFieldType};
+#[cfg(feature = "gdal_2_2")]
+use libc::c_longlong;
 use libc::{c_double, c_int};
+#[cfg(feature = "gdal_2_2")]
+use std::convert::TryInto;
 use std::ffi::CString;
 
 #[cfg(feature = "datetime")]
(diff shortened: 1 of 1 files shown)


src/vector/feature.rs
```diff
@@ -68,6 +71,11 @@ impl<'a> Feature<'a> {
                 let rv = unsafe { gdal_sys::OGR_F_GetFieldAsInteger(self.c_feature, field_id) };
                 Ok(FieldValue::IntegerValue(rv as i32))
             },
+            #[cfg(feature = "gdal_2_2")]
```
lnicola, line 74: We only need 2.0, but this is a mess anyway :-(.

jdroenner: i guess we have to check if the feature is >= gdal_2_x. Maybe we could do a not "gdal_1_11"? Could you look into this?

My suggestion for the future is to rebuild the crate into feature gated modules.

lnicola: > Maybe we could do a not "gdal_1_11"? Could you look into this?

That might end up as `any(feature = "gdal_2_2", feature = "gdal_2_4", feature = "gdal_3_0")`.

> My suggestion for the future is to rebuild the crate into feature gated modules.

Sigh. I'm not sure this works too well -- it's an enum variant, after all.

Maybe we could always compile it in, but fail (or panic at runtime) if the feature wasn't enabled?

jdroenner:
@lnicola i created a PR for my idea with the versioned features https://github.com/georust/gdal/pull/81 (this would also drop support for GDAL 1.*)

lnicola:
I wonder if it's better to drop support for older versions. It's easier to get an up-to-date GDAL these days.

src/vector/feature.rs
michaelkirk: In #86, I faced a similar issue and made a "superset" feature for the major version, maybe it would make sense to do something similar here?

```
gdal_2_2 = ["gdal_2", "gdal-sys/min_gdal_version_2_2"]
gdal_2_4 = ["gdal_2", "gdal-sys/min_gdal_version_2_4"]
gdal_2 = []
```

``` 
#[cfg(feature = "gdal_2")]
...
```

jdroenner:
Would you port this to the feature_gate_versions branch? I think it solves the issues with different versions and i like to merge it into master soon.

lnicola:
@jdroenner I'll port it if it gets merged, but it feels like a lot of complexity TBH. Did you pick a solution for the enum versioning?

jdroenner:
With the change we online support gdal 2 and 3. So i guess the enum can stay in common for now. My idea was to split enums also by version if needed. So we can have different enums for different versions.

PR #87, opened 2020-08-30
Copy vector layers to another dataset + test case

yorickdewid:
When accepted I'll write the same impl for the raster dataset


jdroenner:
Hi, sorry for the delay. Please look at the CI failures and adapt the PR to the recent changes to the crate. We changed Dataset to reflect the GDAL 2.0 design. You can now return a Layer from the method.

PR #88, opened 2020-09-11
added additional metadata retrieval options.

hamersaw:
Allows listing all domains for the specified Metadata and retrieving the strings for a given domain.

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -13,7 +13,7 @@ pub trait Metadata: MajorObject {
         Ok(_string(c_res))
     }
 
-    fn metadata_domain_list(&self) -> Vec<String> {
+    fn metadata_domains(&self) -> Vec<String> {
         let mut domains = Vec::new();
         let c_res = unsafe {
             gdal_sys::GDALGetMetadataDomainList(self.gdal_object_ptr())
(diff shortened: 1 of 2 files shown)


jdroenner:
Hi, sorry for the delay, Could you add a test for each method?

jdroenner:
thanks for the tests. Sadly, GDAL 2 and GDAL 3 have a different behavior. The CI fails for GDAL 2 with: 

 ```
 thread 'raster::tests::test_get_metadata_domains' panicked at 'assertion failed: `(left == right)`

  left: `["IMAGE_STRUCTURE", "xml:XMP", "DERIVED_SUBDATASETS", "COLOR_PROFILE"]`,

 right: `["", "IMAGE_STRUCTURE", "xml:XMP", "DERIVED_SUBDATASETS", "COLOR_PROFILE"]`', src/raster/tests.rs:133:5
```

could you provide two lists to compare the result with? you can use `#[cfg(major_is_2)]` and `#[cfg(major_ge_3)]` for conditional compilation.

hamersaw:
It looks like it may be a compilation flag. Tests for version 2 on Mac seem to work in travis-ci. I tested with gdal versions 2.4.4, 3.0.4, and 3.1.3 installed from source on my debian testing install and they all include the leading "" domain.

Is it enough to remove the leading "" if it exists and ensure the remaining 4 domains exist? I don't believe it will be possible to exhaustively conditionally compile tests for full coverage given functionality difference for the same gdal versions.

jdroenner:
this is strange. i guess removing the "" is the  best way to go.

PR #86, opened 2020-08-28
gdal3 ci

michaelkirk:
Tests are currently failing for gdal 3.1 on both mac and linux.

## Crate changes

- adds gdal_3_1 features
- adds prebuilt 3.1 bindings
- changes some test expactations for gdal_3_1
- extracts gdal_3 feature, which is shared by gdal_3_0 and gdal_3_1 for sharing test behavior.

## CI Changes

- Adds a stable-rust gdal 3.0 target for linux
- Adds a stable-rust gdal 2.4 target for macos
- Adds a stable-rust gdal 3.1 target for macos

For now, I've maintained all the existing stable, beta, nightly gdal 2.0 test runners, though it might make sense to remove some of these older ones and emphasize more modern combinations if that better reflects the user base. The work in this PR makes it fairly easy to shuffle around the test matrix.

PR #92, opened 2020-09-25
Detect GDAL version at build time / remove version features

jdroenner:
This PR changes the build.rs file for gdal-sys and introduces a build.rs for gdal. It makes conditional compilation for version handling more convenient. However, it is still not very pretty.

In gdal-sys, the lib version is now used to select the correct binding file. On Linux it is provided by pkg-config if the gdal lib is detected.
In gdal, the features for versions are removed. The actual GDAL version is requested from gdal-sys and the config parameters are generated.

The changes for **gdal-sys** are:
- version specification as environmental variable  or auto-detection using pkg-config. (i guess windows will require the `GDAL_VERSION` variable to be set).
- no more features to select versions

The changes for **gdal** are: 
- Features for versions are removed. 
- The Current GDAL  version is requested from gdal-sys.
- Build.rs generates config parameters based on the GDAL version.

The following parameters are generated in gdals build.rs:
```
gdal_x, gdal_x_y, gdal_x_y_z with x = major, y = minor, and z= patch. (this provides backward capability to the features)
mayor_is_x, minor_is_y, patch_is_z
mayor_ge_x, minor_ge_y, patch_ge_z  for all version numbers up to the actual version. (major starts with 2).
```
For example GDAL 3.1.2 will have: --cfg major_is_3 --cfg minor_is_1 --cfg patch_is_2 --cfg major_ge_2 --cfg major_ge_3 --cfg minor_ge_0 --cfg minor_ge_1 --cfg patch_ge_0 --cfg patch_ge_1 --cfg patch_ge_2

This allows to enable methods based on versions:
```
#[all(cfg("major_ge_3"), cfg(minor_ge_1" ))]
fn method()
```

### DIFF
--- a/gdal-sys/build.rs
+++ b/gdal-sys/build.rs
@@ -192,7 +192,7 @@ fn main() {
                 version.major, version.minor
             ));
             if !binding_path.exists() {
-                panic!("No pre-build binding available for this GDAl version.");
+                panic!("No pre-built bindings available for GDAL version {}.{}. Use `--features bindgen` to generate your own bindings.", version.major, version.minor);
             }
 
             std::fs::copy(&binding_path, &out_path)
--- a/gdal-sys/build.rs
+++ b/gdal-sys/build.rs
@@ -197,6 +197,8 @@ fn main() {
 
             std::fs::copy(&binding_path, &out_path)
                 .expect("Can't copy bindings to output directory");
+        } else {
+            panic!("No GDAL version detected");
         }
     }
 }
--- a/gdal-sys/build.rs
+++ b/gdal-sys/build.rs
@@ -1,6 +1,6 @@
 #[cfg(feature = "bindgen")]
 use bindgen;
-use pkg_config;
+use semver::Version;
 
 use pkg_config::Config;
 use std::env;
@@ -59,9 +59,11 @@ fn find_gdal_dll(lib_dir: &Path) -> io::Result<Option<String>> {
 
 fn main() {
     println!("cargo:rerun-if-env-changed=GDAL_STATIC");
+    println!("cargo:rerun-if-env-changed=GDAL_DYNAMIC");
     println!("cargo:rerun-if-env-changed=GDAL_INCLUDE_DIR");
     println!("cargo:rerun-if-env-changed=GDAL_LIB_DIR");
     println!("cargo:rerun-if-env-changed=GDAL_HOME");
+    println!("cargo:rerun-if-env-changed=GDAL_VERSION");
 
     let mut need_metadata = true;
     let mut lib_name = String::from("gdal");
@@ -72,6 +74,9 @@ fn main() {
     let mut include_dir = env_dir("GDAL_INCLUDE_DIR");
     let mut lib_dir = env_dir("GDAL_LIB_DIR");
     let home_dir = env_dir("GDAL_HOME");
+    let mut version = env::var_os("GDAL_VERSION")
+        .map(|vs| vs.to_string_lossy().to_string())
+        .and_then(|vs| Version::parse(&vs).ok());
 
     let mut found = false;
     if cfg!(windows) {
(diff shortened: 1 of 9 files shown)


jdroenner:
I guess there are some other tasks which should be addressed as separate issues:
- rework gdal-sys build.rs and separate linux and windows as well as bindgen and predefined binding processing.
- improve gdal-sys build on windows
- check if gdal-sys was build against a lib with the same version

michaelkirk:
> This allows to enable methods based on versions:
>
> ```
> #[any(cfg("major_ge_3"), cfg(minor_ge_1" ))]
> fn method()
> ```

I think that's a typo, shouldn't it be:

> This allows to enable methods based on versions:
>
> ```diff
> -#[any(cfg("major_ge_3"), cfg(minor_ge_1" ))]
> +#[all(cfg("major_ge_3"), cfg(minor_ge_1" ))]
> fn method()
> ```

Otherwise it seems like `method` would be enabled for gdal 1.1+, 2.1+, 3.1+, etc.


To make sure I understand, this is equivalent to (but sometimes more ergonomic than) something like:
```
#[all(cfg("gdal_3"), not(cfg("gdal_3_0")))]
fn foo() {
   do_something()
}
```

Is my understanding correct?

It might be helpful to see them used in practice to evaluate their fitness. It looks like the code is still checking for the no-longer existing features, e.g.:

> #[cfg(feature = "gdal_3_0")]

jdroenner:
> I think that's a typo, shouldn't it be:
> 
> > This allows to enable methods based on versions:
> > ```diff
> > -#[any(cfg("major_ge_3"), cfg(minor_ge_1" ))]
> > +#[all(cfg("major_ge_3"), cfg(minor_ge_1" ))]
> > fn method()
> > ```
> 
> Otherwise it seems like `method` would be enabled for gdal 1.1+, 2.1+, 3.1+, etc.

yes this is a typo. `#[all(cfg("major_ge_3"), cfg(minor_ge_1" ))]` is correct.

> To make sure I understand, this is equivalent to (but sometimes more ergonomic than) something like:
> 
> ```
> #[all(cfg("gdal_3"), not(cfg("gdal_3_0")))]
> fn method()
> ```
> 
> Is my understanding correct?

This is correct. I don't think there is a more maintainable way to handle the versioning.


> It might be helpful to see them used in practice to evaluate their fitness. It looks like the code is still checking for the no-longer existing features, e.g.:
> 
> > #[cfg(feature = "gdal_3_0")]

oh I must have missed that

jdroenner:
i replaced all the ` cfg(feature = gdal_ ...` with the new ones

michaelkirk:
```
$ cargo test
...
  thread 'main' panicked at 'No pre-build binding available for this GDAl version.', gdal-sys/build.rs:195:17
```

Which is accurate - there are no bindings for 3.1 currently. It might be nice to include the version number in the error message and some potential next steps. 

Maybe something like:

>  No pre-build binding available for this GDAL version 3.1. Build with `--features bindgen` to generate your own bindings.

And when I run with bindgen:

```
$ cargo test --features bindgen
...
---- raster::tests::test_get_offset stdout ----
thread 'raster::tests::test_get_offset' panicked at 'assertion failed: `(left == right)`
  left: `None`,
 right: `Some(0.0)`', src/raster/tests.rs:347:5

---- raster::tests::test_get_scale stdout ----
thread 'raster::tests::test_get_scale' panicked at 'assertion failed: `(left == right)`
  left: `None`,
 right: `Some(1.0)`', src/raster/tests.rs:339:5


failures:
    raster::tests::test_get_offset
    raster::tests::test_get_scale
```

Which are exactly the tests fixed by #86 (which I can rebase after this is merged).

So, LGTM

Nice work!

gdal-sys/build.rs
```diff
@@ -166,30 +171,34 @@ fn main() {
...
+                "cargo:rustc-cfg=gdal_sys_{}_{}_{}",
+                version.major, version.minor, version.patch
+            );
+
+            let binding_path = PathBuf::from(format!(
+                "prebuilt-bindings/gdal_{}.{}.rs",
+                version.major, version.minor
+            ));
+            if !binding_path.exists() {
+                panic!("No pre-build binding available for this GDAl version.");
```
michaelkirk, line 195:
```suggestion
                panic!("No pre-built bindings available for GDAL version {}.{}. Use `--features bindgen` to generate your own bindings.", version.major, version.minor);
```

PR #96, opened 2020-09-28
Upgrade to gdal2

jdroenner:
This PR changes to a more GDAL 2 conform structure
- Merged Dataset
- Merged Driver
- Open with GdalOpenEx
- Layer and Rasterband both have a &Dataset to the owning dataset
- remove all checks for GDAL 2, which is the minimum version now


src/dataset.rs
```diff
@@ -1,60 +1,112 @@
...
-}
 
-impl Dataset {
     pub fn open(path: &Path) -> Result<Dataset> {
+        Self::open_ex(path, None, None, None, None)
+    }
+
+    pub fn open_ex(
+        path: &Path,
+        open_flags: Option<u32>,
```
ChristianBeilschmidt, line 46:
```suggestion
        open_flags: Option<GDALAccess::Type>,
```

src/dataset.rs
```diff
@@ -1,60 +1,112 @@
...
         let c_filename = CString::new(filename.as_ref())?;
-        let c_dataset = unsafe { gdal_sys::GDALOpen(c_filename.as_ptr(), GDALAccess::GA_ReadOnly) };
+        let c_open_flags = open_flags.unwrap_or(GDALAccess::GA_ReadOnly); // This defaults to GdalAccess::GA_ReadOnly
+
+        let c_dataset = unsafe {
+            gdal_sys::GDALOpenEx(
+                c_filename.as_ptr(),
+                c_open_flags,
+                ptr::null(),
+                ptr::null(),
```
ChristianBeilschmidt, line 61: `papszOpenOptions` aka `_open_options` would be the most important improvement I guess

src/dataset.rs
```diff
@@ -1,60 +1,112 @@
...
+        _string(rv)
+    }
+
+    pub fn set_projection(&self, projection: &str) -> Result<()> {
+        let c_projection = CString::new(projection)?;
+        unsafe { gdal_sys::GDALSetProjection(self.c_dataset, c_projection.as_ptr()) };
+        Ok(())
+    }
+
+    pub fn create_copy(&self, driver: &Driver, filename: &str) -> Result<Dataset> {
```
ChristianBeilschmidt, line 86:
```suggestion
    pub fn create_copy(&self, driver: &Driver, filename: &Path) -> Result<Dataset> {
```

Maybe use  a `Path` here like in `open_ex`

src/dataset.rs
```diff
@@ -67,42 +119,67 @@ impl Dataset {
...
+    pub fn layer_by_name(&mut self, name: &str) -> Result<Layer> {
+        let c_name = CString::new(name)?;
+        let c_layer = unsafe { gdal_sys::OGR_DS_GetLayerByName(self.c_dataset(), c_name.as_ptr()) };
+        if c_layer.is_null() {
+            Err(_last_null_pointer_err("OGR_DS_GetLayerByName"))?;
+        }
+        Ok(self.child_layer(c_layer))
+    }
+
+    pub fn raster_count(&self) -> isize {
```
ChristianBeilschmidt, line 148: What happens if the dataset is no raster? Same as `layer_count` if the dataset is a raster.

ChristianBeilschmidt (commented):
Yay, finally `openEx` :+1:.

src/dataset.rs
jdroenner: I think all "paths" in the lib should use  Path instead of &str. I guess thats a new issue.

src/dataset.rs
jdroenner: Since GDAL 2 there is only one Dataset: https://gdal.org/doxygen/classGDALDataset.html . The count will be 0 in the respective case.

PR #98, opened 2020-10-01
replace unused back refs with phantom lifetime

jdroenner:


jdroenner:
I guess the lifetime issue is more complex as i thought.

The GDAL docs states for "~GDALDataset() override"
`
Destroy an open GDALDataset.
This is the accepted method of closing a GDAL dataset and deallocating all resources associated with it. 
Equivalent of the C callable GDALClose(). Except that GDALClose() first decrements the reference count, and then closes only if it has dropped to zero.`
So do we have to keep a lifetime or reference around at all?

rmanoka:
@jdroenner The lifetime reference may be important for the `Send` implementation.  Otherwise, with #99 this will allow the dataset to be send to another thread, while a constituent rasterband is in the current thread.  The lifetime reference (even if using PhantomData) prevents this.

Also, I'm not 100% sure if creating a rasterband from a dataset increments the ref. count on the dataset at all.  Internally, it is just storing a pointer to the underlying dataset (`poDS` in [`gdalrasterband.cpp`](https://github.com/OSGeo/gdal/blob/master/gdal/gcore/gdalrasterband.cpp#L2837)).  I understand that the ref. count is incremented if same file is opened in shared mode more than once.

jdroenner:
Thanks for your feedback @rmanoka . GDAL is really very complex... I'm also not sure if the refcount gets increased. Also, the "send" argument is very convincing. Then I would go with the PhantomData approach.

PR #99, opened 2020-10-01
add send to dataset

jdroenner:


lnicola:
Should `RasterBand` be `Send` too?

PR #100, opened 2020-10-01
Ability to read into preallocated slice for rasterband

### LINKED ISSUE #82: Provide read_into_slice ability in RasterBand
The library currently provides a `RasterBand::read_as` function, that allocates a buffer on the heap.  It would be very convenient, if we could re-use the same buffer, and allow functionality to read data into an already allocated slice.  This would save many heap allocations if reading chunks of large rasters in a loop.

A possible implementation would the following addition to `RasterBand` in `src/raster/rasterband.rs`.  It is almost verbatim, the same code as `read_as`, except without an extra `Vec` allocation.

```
    /// Read into a slice `[T]` from a `Dataset`. T implements `GdalType`
    /// # Arguments
    /// * band_index - the band_index
    /// * window - the window position from top left
    /// * window_size - the window size (GDAL will interpolate data if window_size != buffer_size)
    /// * size - the desired size of the `Buffer`
    /// * buffer - the slice to copy into (length is checked with `assert`)
    pub fn read_into_slice<T: Copy + GdalType>(
        &self,
        window: (isize, isize),
        window_size: (usize, usize),
        size: (usize, usize),
        buffer: &mut [T],
    ) -> Result<()>
    {
        let pixels = (size.0 * size.1) as usize;
        assert_eq!(buffer.len(), pixels);

        //let no_data:
        let rv = unsafe {
            gdal_sys::GDALRasterIO(
                self.c_rasterband,
                GDALRWFlag::GF_Read,
                window.0 as c_int,
                window.1 as c_int,
                window_size.0 as c_int,
                window_size.1 as c_int,
                buffer as *mut [T] as *mut T as GDALRasterBandH,
                size.0 as c_int,
                size.1 as c_int,
                T::gdal_type(),
                0,
                0
            )
        };
        if rv != CPLErr::CE_None {
            Err(_last_cpl_err(rv))?;
        }

        Ok(())
    }
```

Happy to provide a PR if it makes sense.

rmanoka:
+ add `read_into_slice`
+ make `read_as` and `read_as_array` to use it
+ minor doc fixes
+ add test for `read_into_slice`

closes #82

### DIFF
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -64,20 +64,23 @@ impl<'a> RasterBand<'a> {
         (self.x_size(), self.y_size())
     }
 
-    /// Read a 'Buffer<T>' from a 'Dataset'. T implements 'GdalType'
+    /// Read data from this band into a slice. T implements 'GdalType'
+    ///
     /// # Arguments
-    /// * band_index - the band_index
     /// * window - the window position from top left
     /// * window_size - the window size (GDAL will interpolate data if window_size != buffer_size)
-    /// * buffer_size - the desired size of the 'Buffer'
-    pub fn read_as<T: Copy + GdalType>(
+    /// * size - the desired size to read
+    /// * buffer - a slice to hold the data (length must equal product of size parameter)
+    pub fn read_into_slice<T: Copy + GdalType>(
         &self,
         window: (isize, isize),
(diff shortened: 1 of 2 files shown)


jdroenner (approved):
Looks good to me. could you resolve the merge conflict?

PR #103, opened 2020-10-17
replace failure with thiserror

### LINKED ISSUE #76: Replace "failure" error handling
Handling gdal crate errors in other crates requires the "failure" crate which is deprecated. We should use a crate / error strategy which does not require additional crates.

jdroenner:
this resolves  #76


michaelkirk:
Nice cleanup!

PR #105, opened 2020-10-21
update geo-types and ndarray

jdroenner:


src/vector/geo_to_gdal.rs
```diff
@@ -167,6 +167,12 @@ where
             geo_types::Geometry::MultiLineString(ref c) => c.to_gdal(),
             geo_types::Geometry::MultiPolygon(ref c) => c.to_gdal(),
             geo_types::Geometry::GeometryCollection(ref c) => c.to_gdal(),
+            geo_types::Geometry::Triangle(_) => Err(GdalError::UnsupportedGeometryType {
+                geometry_type: "Triangle",
```
michaelkirk, line 171: Would it be crazy to convert these to Polygons?

jdroenner: is this the behavior a user would expect?

michaelkirk: Yeah, it's a good question, and I'm not sure. 

On the one hand, people generally expect things to "just work". For example if i had a GeometryCollection with a mix of polygons and triangles, it would be nice if gdal could make use of it.

On the other hand, I can see it being surprising that the operation would be lossy - e.g. I put in a triangle and get back a three point polygon in the results.

ttencate: Heh, I spotted this too late. If conversion is desirable, #106 has the code already :)

jdroenner: merged #107 will update ndarray separatly

PR #115, opened 2020-11-04
Add Feature::fid() and Layer::feature(fid)

ttencate:
To test this, the fids in roads.geojson were changed from string to
numeric, because GDAL only supports numeric fids. As a result it no
longer shows up in the schema.


ttencate:
Hmm, any idea why Travis dislikes this PR only on Trusty? It seems to be seeing differences in geometry where there should be none, as I didn't change any of the geometry in `roads.geojson`, only `id`s (just double-checked).

jdroenner:
maybe it uses a different order for the elements in the geojson?
```
thread 'vector::tests::tests::test_wkt' panicked at 'assertion failed: `(left == right)`

  left: `"LINESTRING (26.1009423 44.4318179,26.1008957 44.4315723,26.1008417 44.431218,26.1007684 44.4309207)"`,

 right: `"LINESTRING (26.1019276 44.4302748,26.1019382 44.4303191,26.1020002 44.4304202)"`', src/vector/tests/mod.rs:188:13
```

ttencate:
You're right, it seems the features are sorted by `id` if `id` is numeric, but returned in input order otherwise. The order is apparently dependent on the GDAL version and probably also on the driver.

I have added a function to get a feature by its fid, which I could use in the tests but is also useful in general.

ttencate:
Travis is happy now.

PR #119, opened 2020-11-12
[docs] min_gdal_version_* features removed as of 87c59f54b701bb3138eb…

michaelkirk:
Those features flags no longer exist - we now detect the version automatically.


michaelkirk:
As a side question, do you have any input on "the right way" to generate bindings (asking for #https://github.com/georust/proj/issues/44)? 

Are you doing it in a docker container or anything, or just running it on your local with the `bindgen` feature and copying over from the built crate?

jdroenner:
Thanks, i already prepared something like this for the next release but then got lost in too much work. Would it be ok if i copy your readme text over to #120 ?

PR #120, opened 2020-11-13
update changes, readme

jdroenner:


gdal-sys/Cargo.toml
```diff
@@ -11,6 +11,6 @@ edition = "2018"
 libc = "0.2"
 
 [build-dependencies]
-bindgen = { version = "0.54", optional = true }
+bindgen = { version = "0.56", optional = true }
```
michaelkirk, line 14: We recently hit some surprising behavior with a bindgen update 

https://github.com/georust/proj/pull/53

Note sure if this affects you, or maybe you've already dealt with it, but bindgen recently changed how size_t gets translated to rust. 

It used to map `size_t` to `usize`. In rust, the byte size of `usize` changes depending on your platform to match the pointer size. On all "first class" rust platforms the size of `usize` in rust is equal to sizeof `size_t` in C. 

But there's nothing in the c standard that says size_t needs to equal pointer width. So in theory, technically, they *could* be different.

...so the (controversial) decision was made to output a concretely sized type (based on the size of `size_t` on the person running bindgen).

So, long story short, if you are packaging and relying on pre-builting bindings and want them to work on 32bit and 64bit machines, you need to enable the `--size_t-is-usize` option for bindgen when you're packing your pre-built bindings.

You can trace the thread back from: https://github.com/rust-lang/rust-bindgen/pull/1688

jdroenner: oh wow i did not know. So if i add `.size_t_is_usize(true)` to bindgen in `build.rs` it should work?

jdroenner: a created a PR for this: #127

michaelkirk: >  if i add .size_t_is_usize(true) to bindgen in build.rs it should work?

Yes, my understanding is that this will maintain the behavior from 0.53 and before.

I notice you're updating from bindgen 0.54, so the ship may have sailed, but it would have only affected your pre-built bindings if you'd regenerated the bindings when you last updated bindgen.

jdroenner: i think all the bindings are older then the last bindgen update :)

PR #121, opened 2020-11-13
Add Dataset::spatial_ref

lnicola:


src/dataset.rs
```diff
@@ -164,6 +164,10 @@ impl Dataset {
         Ok(())
     }
 
+    pub fn spatial_ref(&self) -> Result<SpatialRef> {
+        unsafe { SpatialRef::from_c_obj(gdal_sys::GDALGetSpatialRef(self.c_dataset)) }
+    }
```
michaelkirk, line 169: Would it make sense to add something for GDALSetSpatialRef as well?

lnicola: I suppose so. And I think the CI will fail for GDAL 2.

jdroenner:
i guess the CI did not fail. So do we merge this now? :D

lnicola:
Wow, I'd forgotten about this PR :smile:.

CI failed, but I fixed it with some `#[cfg]`s.

PR #123, opened 2020-11-26
Fix i8 ptr instead of c_char ptr passed to OSRImportFromESRI()

### LINKED ISSUE #122: Fails to build on arm64 due to wrong pointer type
Not sure why, but it seems like something in gdal_sys expects a different pointer type for ESRI when compiling for arm64. This is the only error I get:

```rust
error[E0308]: mismatched types
   --> /app/.cargo/registry/src/github.com-1ecc6299db9ec823/gdal-0.6.0/src/spatial_ref/srs.rs:157:62
    |
157 |         let rv = unsafe { gdal_sys::OSRImportFromESRI(c_obj, ptrs.as_mut_ptr()) };
    |                                                              ^^^^^^^^^^^^^^^^^ expected `u8`, found `i8`
    |
    = note: expected raw pointer `*mut *mut u8`
               found raw pointer `*mut *mut i8`
```

I'm going to dig further into this.

devyn:
c_char is not set to i8 on every architecture, and indeed, this fixes
the build on arm64 and possibly other architectures as well.

Closes #122.

### DIFF
--- a/src/spatial_ref/srs.rs
+++ b/src/spatial_ref/srs.rs
@@ -1,6 +1,6 @@
 use crate::utils::{_last_cpl_err, _last_null_pointer_err, _string};
 use gdal_sys::{self, CPLErr, OGRCoordinateTransformationH, OGRErr, OGRSpatialReferenceH};
-use libc::c_int;
+use libc::{c_int, c_char};
 use std::ffi::{CStr, CString};
 use std::ptr;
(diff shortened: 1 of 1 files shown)


devyn:
Btw, I can confirm that all tests pass on arm64 on linux with this change.

jdroenner:
yep the documentation also states that this should be a char. thanks!

PR #128, opened 2020-12-24
fix docs.rs build for gdal-sys (hopefully)

michaelkirk:
The gdal-sys docs are currently failing to build:
https://docs.rs/crate/gdal-sys/0.3.0/builds/326757

I think this is because the build script requires libgdal to actually be installed. This adds a short circuit to skip the normal build process and hardcodes a version of the prebuilt bindings to use when generating docs.

### DIFF
--- a/gdal-sys/build.rs
+++ b/gdal-sys/build.rs
@@ -57,6 +57,31 @@ fn find_gdal_dll(lib_dir: &Path) -> io::Result<Option<String>> {
 }
 
 fn main() {
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap()).join("bindings.rs");
+
+    // Hardcode a prebuilt binding version while generating docs.
+    // Otherwise docs.rs will explode due to not actually having libgdal installed.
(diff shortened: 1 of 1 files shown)


michaelkirk:
Build failure is same as the build failure in master:

```
error: no matching package named `gdal-sys` found
location searched: /home/travis/build/michaelkirk/gdal/gdal-sys
prerelease package needs to be specified explicitly
gdal-sys = { version = "0.3.1-alpha.0" }
```

I think it's because we're using a semantic qualifier in the constraints, which isn't compatible with "alpha" (and other prerelease) version strings.

jdroenner:
Ah yes, i guess we should remove the alpha. Cargo Release did that. Maybe we could also remove the Version from the gdal-sys dependancy.

michaelkirk:
> Ah yes, i guess we should remove the alpha. Cargo Release did that. Maybe we could also remove the Version from the gdal-sys dependancy.

I'm not familiar with `cargo release`,  but yeah I was surprised by the "alpha" tag.

Issue #132, opened 2021-01-03
Feature::field should return Result<Option<FieldValue>>


dead10ck:
Currently, [`Feature::field`](https://github.com/georust/gdal/blob/fb85a0cd26b12fa1480321293296212df857fcc1/src/vector/feature.rs#L70) returns `Result<FieldValue>`, but fields can be null. When this is the case, the default value for the `FieldValue` variant is returned. This makes it difficult to distinguish e.g. a legitimate Integer 0 value from the absence of a value.

I am not very familiar with the GDAL lib, but just taking a quick one through, it looks like maybe the `OGR_F_IsFieldNull` function could be used to see if a field is null first, and if so, return `None`.

PR #134, opened 2021-01-03
Feature::field returns Result<Option<FieldValue>>

dead10ck:
Fixes https://github.com/georust/gdal/issues/132

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,37 @@
 # Changes
+
 ## Unreleased
+* **Breaking**: Change `Feature::field` return type from
+  `Result<FieldValue>` to `Result<Option<FieldValue>>`. Fields
+  can be null. Before this change, if a field was null, the value
+  returned was the default value for the underlying type.
+  However, this made it impossible to distinguish between null
+  fields and legitimate values which happen to be default value,
+  for example, an Integer field that is absent (null) from a 0,
+  which can be a valid value. After this change, if a field is
+  null, `None` is returned, rather than the default value.
+
+  If you happened to rely on this behavior, you can fix your code
+  by explicitly choosing a default value when the field is null.
+  For example, if you had this before:
+
+  ```rust
+  let str_var = feature.field("string_field")?
+      .into_string()
+      .unwrap();
+  ```
+
+  You could maintain the old behavior with:
+
+  ```rust
+  use gdal::vector::FieldValue;
+
+  let str_var = feature.field("string_field")?
+      .unwrap_or(FieldValue::StringValue("".into()))
+      .into_string()
+      .unwrap();
+  ```
+    * <https://github.com/georust/gdal/pull/134>
 * Add basic support to read overviews
 * BREAKING: update geo-types to 0.7.0. geo-types Coordinate<T> now implement `Debug`
   * <https://github.com/georust/gdal/pull/146>
--- a/examples/read_write_ogr.rs
+++ b/examples/read_write_ogr.rs
@@ -15,9 +15,10 @@ fn run() -> Result<()> {
         .collect::<Vec<_>>();
 
     // Create a new dataset:
-    let _ = fs::remove_file("/tmp/abcde.shp");
+    let path = std::env::temp_dir().join("abcde.shp");
+    let _ = fs::remove_file(&path);
     let drv = Driver::get("ESRI Shapefile")?;
-    let mut ds = drv.create_vector_only("/tmp/abcde.shp")?;
+    let mut ds = drv.create_vector_only(path.to_str().unwrap())?;
     let lyr = ds.create_layer_blank()?;
 
     // Copy the origin layer shema to the destination layer:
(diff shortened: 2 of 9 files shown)


michaelkirk:
This seems like a useful change, though it's likely to break a lot of existing code right? Maybe worth a note in CHANGES.md to help people migrating?

dead10ck:
Yes, it's likely to break a lot of code. I do think it's beneficial, however. I myself thought of this because I found myself writing code like

```rust
fn get_property_string(
    key: &str,
    feature: &mut gdal::vector::Feature,
) -> Result<Option<String>, Box<dyn Error + Send + Sync>> {
    let s: String = feature
        .field(key)?
        .into_string()
        .ok_or(format!("missing key: {}", key))?;

    if &s[..] == "" {
        Ok(None)
    } else {
        Ok(Some(s))
    }
}
```

to get around it. This just so happened to work for me, but I imagine there are use cases where `0` and `""` are valid values that need to be distinguished from null.

I'll add a note to the change log.

CHANGES.md
```diff
@@ -1,4 +1,16 @@
...
+
+## Unreleased
+* Change `Feature::field` return type from `Result<FieldValue>` to
+  `Result<Option<FieldValue>>`. Fields can be null. Before this change, if a
+  field was null, the value returned was the default value for the underlying
+  type. However, this made it impossible to distinguish between null fields and
+  legitimate values which happen to be default value, for example, an Integer
+  field that is absent (null) from a 0, which can be a valid value. After this
+  change, if a field is null, `None` is returned, rather than the default
+  value.
```
michaelkirk, line 11: Are there likely to be people for whom the old behavior, of getting the default value, is desirable?

If so, after this change, how will they restore that behavior?
Will they need to use snippets like: `.unwrap_or_default(0)` `. unwrap_or_default("")`? or is there a way to say something like "field.default_value()"?

michaelkirk: (To be clear, I think your change is good overall, just trying to think of any pain we can save a minority of people)

dead10ck: Yes, that's right, if they had `dataset.field("some_field")?.into_int()` before, they could do `dataset.field("some_field")?.unwrap_or(FieldValue::Integer(0)).into_int()`. Although it would surprise me what kind of use case there would be where people wouldn't be surprised by nulls turning into `0` or `""`. Probably the more likely case is people aren't aware that nulls silently get turned into real values.

michaelkirk: > Probably the more likely case is people aren't aware that nulls silently get turned into real values.

Yeah, that seems likely. :+1:

> it would surprise me what kind of use case there would be where people wouldn't be surprised by nulls turning into 0 or "".

I've never personally used this feature, but apparently you can get/set other defaults for fields: https://gdal.org/doxygen/classOGRFieldDefn.html#ac4210fa7c6f10ed090a5558224447cfa

Maybe out of scope for this PR, but since you're thinking about it, what's your opinion on exposing an API leveraging that functionality like:
```
let foo: Result<FieldValue> = dataset.field_default("some_field")?;
```

Then people could exactly mirror the previous behavior with:
```
let foo: Result<FieldValue> = dataset.field("some_field")?.unwrap_or_else(dataset.field_default("some_field")?)
``` 

and maybe offered succinctly as: 
```
let foo: Result<FieldValue> = dataset.field_value_or_default("some_field")?;
```

dead10ck: That makes a lot of sense to me. Having a `field_value_or_default` function is clear and descriptive, and exposes useful functionality from OGR.

And since you ask, I was actually starting to think about whether it would be possible to do away with the `FieldValue` type completely and just use generics. I haven't fully fleshed out the details, but I was thinking it should be possible to just make the `field` generic on the return type, so you could just do

```rust
let str_var: String = dataset.field("string_field")?.unwrap();
let int_var: i32 = dataset.field("int_field")?.unwrap();
etc...
```

I think it could be done if we made a `Field` type that mirrors the `OGRFieldDfn` type, then `impl TryFrom<str> for Field<String/i32/etc>`, and then the signature of `field` could be something like

```rust
fn field<T, D>(field_dfn: D) -> Result<Option<T>> where D: Into<Field<T>>
```

The `Field` type could have that `set_default` method on it too.

dead10ck: Was there anything else you wanted for this PR? Did you want me to add that example code snippet for getting a default value?

jdroenner: i think the PR looks fine. We will wait a few days for other PRs with breaking changes before we release a new version.
Could you add docs to the `field_from_id` method and include that example snippet?

dead10ck: Done 🙂

jdroenner:
uh clippy is not happy :( could you check what is wrong?

dead10ck:
Oh sorry, I did not run the tests with --all-features, so I missed a breakage in the examples. I've fixed it, and a couple other issues with the examples. I ran all of the example binaries and made sure they worked.

PR #136, opened 2021-01-04
maintenance

jdroenner:


build.rs
```diff
@@ -1,13 +1,19 @@
...
+    unsafe {
         let res_ptr = gdal_sys::GDALVersionInfo(c_key.as_ptr());
         let c_res = std::ffi::CStr::from_ptr(res_ptr);
         c_res.to_string_lossy().into_owned()
-    };
-    version_string
+    }
 }
 
 fn main() {
```
michaelkirk, line 19: Stubbing the gdal implementation like this could work.

Alternatively would it be simpler to instead short-circuit the entire build script like in gdal-sys?

jdroenner: i'm happy if this works. Parsing the version number is all the build script does so i guess this is more or less the short-circuit :see_no_evil:

michaelkirk: well then, excellent!

PR #141, opened 2021-01-06
Select resampling algo

### LINKED ISSUE #140: Any way of selecting GDALRIOResampleAlg?
At the moment, to read a raster into a slice you are using [GDALRasterIO](https://github.com/georust/gdal/blob/1ee19ed1ea3da385dfc1dd47090e5a42b7b3bffc/src/raster/rasterband.rs#L89), which doesn't provide a way to select a resampling algorithm.

If I'm not mistaken, [GDALRasterIOEx](https://github.com/georust/gdal/blob/2fadecc39c84c317ca023cfaf67fd6fbd0c568a5/gdal-sys/prebuilt-bindings/gdal_2.4.rs#L6156) has an extra option which could be used to do that.

Is there any way to pass a sampling algorithm ([GDALRIOResampleAlg](https://github.com/georust/gdal/blob/2fadecc39c84c317ca023cfaf67fd6fbd0c568a5/gdal-sys/prebuilt-bindings/gdal_2.4.rs#L5285)) to `read_as` or `read_into_slice`?

I tried to implement my own `read_into_slice` function but `RasterBand.c_rasterband` is private and I'm not sure how to proceed.

Cheers!

spadarian:
---

This PR changes the functions used to read rasters by adding an extra parameter to select the resampling algorithm used when `buffer_size != window size` (GDAL needs to interpolate the values) (closes #140)

### DIFF
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -4,8 +4,8 @@ use crate::metadata::Metadata;
 use crate::raster::{GDALDataType, GdalType};
 use crate::utils::{_last_cpl_err, _last_null_pointer_err, _string};
 use gdal_sys::{
-    self, CPLErr, GDALColorInterp, GDALMajorObjectH, GDALRIOResampleAlg, GDALRWFlag,
-    GDALRasterBandH, GDALRasterIOExtraArg,
+    self, CPLErr, GDALColorInterp, GDALMajorObjectH, GDALRWFlag, GDALRasterBandH,
+    GDALRasterIOExtraArg,
 };
 use libc::c_int;
 use std::ffi::CString;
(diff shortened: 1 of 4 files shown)


spadarian:
I've just started using Rust, so I'm not sure this is the best implementation (it introduces some breaking changes).

I'm happy to change it if there is a better way!

lnicola:
Looks like this needs rebasing (`Dataset` moved IIRC). I'm not terribly familiar to GDAL, does this API support reading a window and resampling at the same time?

spadarian:
It does. When reading a window, if `window_size != buffer_size` then GDAL performs an interpolation. By default, it uses the nearest neighbour. This PR allows you to set a different resampling method.

src/raster/rasterband.rs
```diff
@@ -12,6 +15,68 @@ use ndarray::Array2;
 
 use crate::errors::*;
 
+/// Extra options used to read a raster.
+///
+/// For documentation, see `gdal_sys::GDALRasterIOExtraArg`.
+#[derive(Debug)]
+pub struct RasterIOExtraArg {
+    pub n_version: libc::c_int,
+    pub e_resample_alg: GDALRIOResampleAlg::Type,
```
jdroenner, line 24: i think we should have an enum for the resample algorithms. That will make it easier for most users.

src/raster/rasterband.rs
```diff
@@ -12,6 +15,68 @@ use ndarray::Array2;
...
+    pub pfn_progress: gdal_sys::GDALProgressFunc,
+    pub p_progress_data: *mut libc::c_void,
```
jdroenner, line 26: We should hide the pointer stuff for now. In the long run we could add a callback and/or a future/channel to access the state.

src/raster/rasterband.rs
```diff
@@ -12,6 +15,68 @@ use ndarray::Array2;
...
+    pub b_floating_point_window_validity: libc::c_int,
```
jdroenner, line 27: can we change this to i32 or another integer from core?

jdroenner (changes requested):

src/raster/rasterband.rs
spadarian: `GDALRIOResampleAlg::Type` are already constants. Should I create a mod inside rasterband to map it?

src/raster/rasterband.rs
spadarian: I guess an `usize` should be ok as well.

Just out of curiosity, what would be the difference? Just a "readability" thing? Or is there any technical reason?

jdroenner: just readability

src/raster/rasterband.rs
spadarian: @jdroenner I added an enum with the names of the algorithms and a function to map them to integers (GDAL uses integers).

spadarian:
I added the suggested changes and did a rebase.

PR #142, opened 2021-01-06
Read overviews

spadarian:
---

This PR adds basic support to read overviews.


jdroenner:
Thanks for this great enhancement. I guess we should move the methods to rasterband. In the GDAL C++ API it is also a method of rasterband.

spadarian:
I had it there initially but it felt weird when I was writing the test.

To return the overlay as a band I use `RasterBand::from_c_rasterband`, which requires a `&Dataset` as a first argument. To use it had to:
* Get the band (`Dataset.rasterband`)
* Use the band to get the overview, passing the `&Dataset` as an argument.

It does make sense to get the overview from a band... but also to get the overview from the dataset and get a single band of it (like this PR).

If you want to be completely consistent with the GDAL C++ API, I can change it tomorrow.

michaelkirk:
Hi, I'm not a maintainer. Just a drive by reviewer and fan of the gdal crate.

> I had it there initially but it felt weird when using it to write the test.

I'm trying to understand what would have been weird about it. Would it be like this:

```diff
 #[test]
 fn test_read_overviews() {
     let dataset = Dataset::open(fixture!("tinymarble.tif")).unwrap();
     let rasterband = dataset.rasterband(1).unwrap();
-    let overview_count = dataset.overview_count(1).unwrap();
+    let overview_count = rasterband.overview_count().unwrap();
     assert_eq!(overview_count, 2);

-     let overview_2 = dataset.overview_get(1, 0).unwrap();
-     let overview_4 = dataset.overview_get(1, 1).unwrap();
+     let overview_2 = dataset.rasterband(1).unwrap().overview(0).unwrap();
+     let overview_4 = dataset.rasterband(1).unwrap().overview(1).unwrap();
     assert_eq!(rasterband.size(), (100, 50));
     assert_eq!(overview_2.size(), (50, 25));
     assert_eq!(overview_4.size(), (25, 13));
 }
```

That seems reasonable to me, but maybe I'm missing something.

I generally assume that the GDAL people have more time and experience in designing a good GDAL API - by sticking close to it we can benefit from their design expertise and also be familiar to people with other GDAL experience. 

Of course sometimes concessions have to be made for things to work with Rust. And sometimes it's worth breaking those rules if we have a _substantially_ better API to offer due to some unique rust feature.

src/dataset.rs
```diff
@@ -399,6 +399,30 @@ impl Dataset {
...
+        unsafe {
+            let c_band = gdal_sys::GDALGetRasterBand(self.c_dataset, band_index as c_int);
+            if c_band.is_null() {
+                return Err(_last_null_pointer_err("GDALGetRasterBand"));
+            }
+            Ok(gdal_sys::GDALGetOverviewCount(c_band))
+        }
+    }
+
+    pub fn overview_get(&self, band_index: isize, overview_index: isize) -> Result<RasterBand> {
```
michaelkirk, line 413: 🤓 idiomatic rust omits the `get` from getters: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

spadarian:
The code would actually be something like:

```diff
 #[test]
 fn test_read_overviews() {
     let dataset = Dataset::open(fixture!("tinymarble.tif")).unwrap();
     let rasterband = dataset.rasterband(1).unwrap();
-    let overview_count = dataset.overview_count(1).unwrap();
+    let overview_count = rasterband.overview_count().unwrap();
     assert_eq!(overview_count, 2);

-     let overview_2 = dataset.overview_get(1, 0).unwrap();
-     let overview_4 = dataset.overview_get(1, 1).unwrap();
+     let overview_2 = dataset.rasterband(1).unwrap().overview(dataset, 0).unwrap();
+     let overview_4 = dataset.rasterband(1).unwrap().overview(dataset, 1).unwrap();
     assert_eq!(rasterband.size(), (100, 50));
     assert_eq!(overview_2.size(), (50, 25));
     assert_eq!(overview_4.size(), (25, 13));
 }
```

Unless there is a reference to the Dataset stored in the RasterBand? The only reference I see is `RasterBand.phantom` but I'm not sure if it is possible to do something with it...

spadarian:
I'm just starting to use Rust, so I'm not sure if this is possible...

`RasterBand::from_c_rasterband` needs to carry a lifetime. At the moment it uses `Dataset` to do that.

If I make `from_c_rasterband` generic, I can use the lifetime of a `Dataset` or a `RasterBand`. In the case where the overview carries the lifetime of the `RasterBand`, we would have a chain `Dataset` -> `RasterBand` -> overview (which is also a `RasterBand`). I'm not sure if it makes sense that the overview depends on the `RasterBand`...

* Is it possible to "copy" the lifetime used by `RasterBand` so we have `Dataset` -> `RasterBand` AND `Dataset` -> overview?
* I'm not sure how to make `from_c_rasterband` generic but restricted to a `Dataset` or `RasterBand`.

michaelkirk:
> The code would actually be something like:

Ah, I see it's not so simple. Thanks for explaining.

> I'm just starting to use Rust, so I'm not sure if this is possible...

I'm having a hard time following what you're trying to do. If you could push up a different branch with the code close to what you *want* to write (even if it doesn't compile) I could take a look.

spadarian:
Sure. Check https://github.com/spadarian/gdal/tree/generic_from_c_rasterband

jdroenner:
At some point before v0.7.0 we had a reference to the `Dataset` in `Rasterband`. I guess we could put that back it there is no other way to solve this. The generic approach could be an alternative since we only care for the lifetime.
I will also think about it.

michaelkirk:
> overview depends on (lifetime) rasterband and dataset, so I can't drop either.

Ah, I think I understand what you're trying to do.

This is probably the first time I've written code that carries a lifetime on phantom data (I'm also pretty new to rust), so it might not be correct, but take a look: https://github.com/michaelkirk/gdal/tree/mkirk/generic_from_c_rasterband

spadarian:
That does carry the lifetime, so I think it is ok.

The "problem" is that the overview shares the lifetime with the initial `RasterBand`:

 `'a Dataset` -> `'a RasterBand` -> `'a overview`.

In my head, a `RasterBand` should be only linked to a `Dataset`. The overview is a `RasterBand`, so the same rule should apply.

 `'a+'b Dataset` -> `'a RasterBand` -> `'b overview`... if that makes any sense

Is the overview internally linked to the `RasterBand`? Is it possible to point directly to the `Dataset`? I'm way out of my depth here :smile: Maybe what I'm thinking is not even possible.

If it is possible, this should compile:
```rust
#[test]
fn test_rasterband_lifetime() {
    let dataset: Dataset = Dataset::open(fixture!("tinymarble.tif")).unwrap();
    let rasterband = dataset.rasterband(1).unwrap();
    let overview = rasterband.overview(0).unwrap();

    drop(rasterband);
    assert_eq!(overview.size(), (50, 25));
}

```

spadarian:
Actually, is there a problem if the `RasterBand` outlives the `Dataset`? I imagine that is why you added `PhantomData`

jdroenner:
i think i have to read the gdal documentation on overviews first. will try to do this on the weekend :)

michaelkirk:
> The "problem" is that the overview shares the lifetime with the initial RasterBand:
>
> 'a Dataset -> 'a RasterBand -> 'a overview.

I guess I _am_ confused, because I thought that was exactly the point. 😅

michaelkirk:
> Is the overview internally linked to the RasterBand?

My understanding is that, if we build an `overview`, the lifetime of that `overview` is ultimately tied to the same borrow of `dataset` as was used by `rasterband`, which is what I thought we wanted.

spadarian:
In your branch (and mine), the `overview` is tied to `rasterband` and by transitivity to `dataset`, where it should be tied directly to `dataset` (no dependency to `rasterband` at all).

spadarian:
Any thoughts on this @jdroenner ?

I'm not sure if it is possible to have that `Dataset` -> `Rasterband` and `Dataset` -> `overview` dependence if we create `overview` within `Rasterband`. At least not with my basic knowledge of rust :smile:

jdroenner:
sorry for the late reply. I was unsure why the example above does not work with the `dataset: &'aDataset` approach.
I played with the lifetime a bit and i guess the solution is quite simple. We need to add 'a to the result Rasterband like this:     `pub fn overview(&self, overview_index: isize) -> Result<RasterBand<'a>> {` . Basically it was a lucky guess :see_no_evil: 

I created a branch which includes your test: https://github.com/georust/gdal/tree/overview-lifetime

Is this what you had in mind?

spadarian:
Yes, exactly what I had in mind!

I need to practice all this lifetime business... :facepalm:

It looks ready to me.

jdroenner:
i'm going to merge https://github.com/georust/gdal/pull/149 which contains the lifetime changes.

PR #144, opened 2021-01-07
Add RasterBand's colour interpretation

shuternay:

---

This PR adds methods for reading and setting `RasterBand`'s colour interpretation. It also adds methods for getting interpretation value name and getting interpretation value by name.

P.S. I'm not sure how to achieve this, but shouldn't mutable borrow of `RasterBand` require mutable borrow of it's `Dataset`?


src/raster/mod.rs
```diff
@@ -4,7 +4,9 @@ mod rasterband;
 mod types;
 mod warp;
 
-pub use rasterband::{Buffer, ByteBuffer, RasterBand};
+pub use rasterband::{
+    get_color_interpretation_by_name, get_color_interpretation_name, Buffer, ByteBuffer, RasterBand,
```
jdroenner, line 8: i don't think we need to pub use the methods since we pub use RasterBand. However, i guess we need to pub use GDALColorInterp?

src/raster/rasterband.rs
```diff
@@ -260,6 +261,18 @@ impl<'a> RasterBand<'a> {
         Ok(())
     }
 
+    pub fn get_color_interpretation(&self) -> GDALColorInterp::Type {
```
jdroenner, line 264: in rust the "get_" prefix is usually skipped. So please remove it from the getter methods.

src/raster/tests.rs
```diff
@@ -1,8 +1,8 @@
 use crate::dataset::Dataset;
 use crate::metadata::Metadata;
-use crate::raster::ByteBuffer;
+use crate::raster::{get_color_interpretation_by_name, get_color_interpretation_name, ByteBuffer};
```
jdroenner, line 3: the methods should work if rasterband is imported.

jdroenner (changes requested):
please look at the exports / imports and remove the "get_" prefix. Also please resolve the merge conflicts.

src/raster/mod.rs
shuternay: I'm a bit confused here. `get_color_interpretation_{by_}name` are not `RasterBand`'s methods (and I'm not sure if it is possible to pub use struct's methods), so pub using `RasterBand` doesn't make these functions visible. And `rasterband` module is private, so I need to pub use these functions here.

src/raster/rasterband.rs
```diff
@@ -320,3 +333,13 @@ impl<T: GdalType> Buffer<T> {
 }
 
 pub type ByteBuffer = Buffer<u8>;
+
+pub fn get_color_interpretation_name(color_interp: GDALColorInterp::Type) -> String {
```
michaelkirk, line 337: I'm just a drive by reviewer, but my two cents:

Typically we've been creating wrapper structs for the underlying gdal-sys types, rather than adding free methods. Would it be more in line with the rest of the crate to have an API more like:

```
struct ColorInterpretation(GDALColorInterp::Type)

impl ColorInterpretation {
    fn name(&self) -> &str { ... }
    fn from_name(&str) -> Self  { ... }
}

src/raster/mod.rs
jdroenner: you are right. somehow i thought they are Rasterband methods.

src/raster/rasterband.rs
shuternay: Sounds reasonable. I guess it should be an enum with interpretations then

shuternay:
I've added an enum for color interpretations (should resolve problems with free methods and imports) and resolved merge conflicts

PR #145, opened 2021-01-19
Add more SpatialRef methods

dmarteau:
---

This PR add more bridge functions to gdal::spatial_ref::srs::SpatialRef implementation.

Struct added

* AreaOfUse

Type alias

* AxisOrientationType 

Functions added:

* get_name
* get_angular_units
* get_angular_units_name
* get_linear_units
* get_linear_units_name
* is_geographic
* is_derived_geographic
* is_local
* is_projected
* is_compound
* is_geocentric
* is_vertical
* get_axis_orientation
* get_axis_name
* get_axes_count
* get_area_of_use

Added tests.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,9 +1,8 @@
 # Changes
 
-## unreleased
-* Add more functions to SpatialRef implementation
-
 ## Unreleased
+* Add more functions to SpatialRef implementation
+    * <https://github.com/georust/gdal/pull/145>   
 * **Breaking**: Change `Feature::field` return type from
   `Result<FieldValue>` to `Result<Option<FieldValue>>`. Fields
   can be null. Before this change, if a field was null, the value
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 # Changes
 
+## unreleased
+* Add more functions to SpatialRef implementation
+
 ## Unreleased
 * **Breaking**: Change `Feature::field` return type from
   `Result<FieldValue>` to `Result<Option<FieldValue>>`. Fields
--- a/src/spatial_ref/srs.rs
+++ b/src/spatial_ref/srs.rs
@@ -78,6 +78,18 @@ impl CoordTransform {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct AreaOfUse {
+    pub west_lon_degree: f64,
+    pub south_lat_degree: f64,
+    pub east_lon_degree: f64,
+    pub north_lat_degree: f64,
+    pub name: String,
+}
+
+pub type AxisOrientationType = gdal_sys::OGRAxisOrientation::Type;
+
+
 #[derive(Debug)]
 pub struct SpatialRef(OGRSpatialReferenceH);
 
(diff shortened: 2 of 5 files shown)


CHANGES.md
```diff
@@ -1,5 +1,8 @@
 # Changes
 
+## unreleased
+* Add more functions to SpatialRef implementation
```
michaelkirk, line 4: Looks like there's already an Unreleased heading, you can add this new item beneath it.

dmarteau: Done.

src/spatial_ref/srs.rs
```diff
@@ -303,6 +314,111 @@ impl SpatialRef {
...
+    pub fn get_angular_units_name(&self) -> Result<String> {
+        let mut c_ptr = ptr::null_mut();
+        unsafe { gdal_sys::OSRGetAngularUnits(self.0, &mut c_ptr) };
+        if c_ptr.is_null() {
+            return Err(_last_null_pointer_err("OSRGetAngularUnits"));
+        }
+        Ok(_string(c_ptr))
+    }
+
+    pub fn get_angular_units(&self) -> f64 {
```
michaelkirk, line 335: So if I want to get both the name and value I need to make two calls like this right?

```
let name = srs.get_angular_units_name();
let value = srs.get_angular_units();
```

Alternatively, you could consider making the api something like:
```
let (name, value) = srs.get_angular_units()
let name = srs.get_angular_units_name();
let value = srs.get_angular_units_value();
```

(I don't have any authority in whether or not this gets merged, I'm just a gdal crate user that watches PR's come by)

src/spatial_ref/srs.rs
```diff
@@ -303,6 +314,111 @@ impl SpatialRef {
...
+    }
+
+    #[inline]
+    pub fn is_vertical(&self) -> bool {
+        unsafe { gdal_sys::OSRIsVertical(self.0) == 1 }
+    }
+
+    pub fn get_axis_orientation(&self, target_key: &str, axis: i32) -> AxisOrientationType {
+        // We can almost safely assume that if we fail to build a CString then the input
+        // is not a valide key.
```
michaelkirk, line 390: Should this be an Option/Result type? Or is it reasonable for users to infer that an error occurred when they get `gdal_sys::OGRAxisOrientation::OAO_Other`?

jdroenner:
Thank you @michaelkirk for reviewing and of cause thanks @dmarteau for the contribution. I have a lot to do at the moment so sometimes there is just no time for more code reviews.

src/spatial_ref/srs.rs
jdroenner: uh well i guess both ways are fine. I would suggest to keep it this way to avoid mixing Results.

src/spatial_ref/srs.rs
jdroenner: you are right is should be a result type. The GDAL docs state that it will return "  the name of the axis or NULL on failure. "

dmarteau: That was not obvious to me since  "No CPLError is issued on routine failures (such as not finding the AXIS)".

The error would then be a 'AxisNotFound' error value ?
In this case `get_axis_name` should also return a Result too not an `Option<String>` imho,

src/spatial_ref/srs.rs
dmarteau: It tooks me some time to end up with this solution but  many of the use case I have faced up involved returning either only the name or only the result for calculation, since returning the name implies allocation it considered that it was better to have a zero cost way to get the units factor value (note: this is also the choice of the python bindings).

src/spatial_ref/srs.rs
dmarteau: Changed to Result return type for both `get_axis_orientation` and `get_axis_name`

jdroenner: great thanks :+1:

PR #146, opened 2021-01-19
update to geo-types 0.7

michaelkirk:
---

geo-types <T> must now support Debug. Rather than change all the `num_traits:Float` to `num_traits::Float + Debug`, I've leveraged the new `geo_types::CoordFloat` to the same effect. Consequently, we no longer need to explicitly depend on num_traits.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,11 @@
 # Changes
+
+## UNRELEASED
+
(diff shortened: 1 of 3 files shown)


jdroenner:
i guess we should also merge the open PRs since this will trigger a minor release anyway?

michaelkirk:
That makes sense to me. 👍

michaelkirk:
Glad to see I haven't been made redundant yet. 😅

PR #147, opened 2021-01-21
github-ci + dependabot

jdroenner:
---


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+
+version: 2
+updates:
+
+  - package-ecosystem: "cargo"
```
michaelkirk, line 5: I've been considering doing this for some other crates. I'll be watching to see how well this works for you!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
...
+  ubuntu_20_04:
+    name: "ci ubuntu-20.04"
+    runs-on: "ubuntu-20.04"
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Install GDAL
+        run: |
+          sudo apt-get update
```
michaelkirk, line 26: Alternatively if the setup is slow you can prepare a docker container. We have a couple hosted at https://github.com/georust/docker-images that we use for georust/geo and georust/proj.

It's much faster to run CI this way, especially since we're compiling proj from source. 

The downside is I seem to get into a three hour wrestling match with Docker every 6 months.

jdroenner: yeah i guess the docker container is the next step after getting this to work.

PR #152, opened 2021-01-27
conventional rust naming

michaelkirk:
---

Nitpicky follow up to #145. Sorry it didn't occur to me during review! 

Idiomatically, rust API's omit the `get_` prefix from "getters". 

https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

### DIFF
--- a/src/spatial_ref/srs.rs
+++ b/src/spatial_ref/srs.rs
@@ -315,7 +315,7 @@ impl SpatialRef {
     }
 
     #[cfg(major_ge_3)]
-    pub fn get_name(&self) -> Result<String> {
+    pub fn name(&self) -> Result<String> {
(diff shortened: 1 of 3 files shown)


src/spatial_ref/srs.rs
```diff
@@ -439,12 +439,18 @@ impl SpatialRef {
     }
 
     #[cfg(major_ge_3)]
+    #[deprecated note = "use `axis_mapping_strategy` instead"]
     pub fn get_axis_mapping_strategy(&self) -> gdal_sys::OSRAxisMappingStrategy::Type {
```
michaelkirk, line 443: This method was released with 0.7.0, so I gracefully deprecated it. The other methods changed haven't been released yet so I just changed them inline.

PR #153, opened 2021-01-27
Implement SQL queries

dead10ck:
Hello! I am trying to implement support for doing SQL queries through the `ExecuteSQL` functions. I've got basic functionality working: you can pass a SQL query to `Dataset::execute_sql` and you get back a handle to a new type `ResultSet`, which is a thin wrapper around a handle to a `Layer` and the layer's `Dataset`. The new type was necessary because SQL result sets must be freed explicitly, whereas regular layers owned by a `Dataset` do not.

However, I'm submitting a draft, as I'd like your feedback on the changes, and also because I'm running into a problem, and I was hoping maybe you could help. It seems that there is some kind of problem with choosing the sqlite dialect. Using `ogrinfo` works just fine using sqlite with spatialite functions:

```
✦ ❯ ogrinfo -dialect sqlite -sql "select * from roads where highway = 'pedestrian' and NumPoints(GEOMETRY) = 3" -spat 26.1017 44.4297 26.1025 44.4303 fixtures/roads.geojson
INFO: Open of `fixtures/roads.geojson'
      using driver `GeoJSON' successful.

Layer name: SELECT
Geometry: Line String
Feature Count: 1
Extent: (26.102255, 44.429870) - (26.102745, 44.430318)
Layer SRS WKT:
GEOGCRS["WGS 84",
    DATUM["World Geodetic System 1984",
        ELLIPSOID["WGS 84",6378137,298.257223563,
            LENGTHUNIT["metre",1]]],
    PRIMEM["Greenwich",0,
        ANGLEUNIT["degree",0.0174532925199433]],
    CS[ellipsoidal,2],
        AXIS["geodetic latitude (Lat)",north,
            ORDER[1],
            ANGLEUNIT["degree",0.0174532925199433]],
        AXIS["geodetic longitude (Lon)",east,
            ORDER[2],
            ANGLEUNIT["degree",0.0174532925199433]],
    ID["EPSG",4326]]
Data axis to CRS axis mapping: 2,1
Geometry Column = GEOMETRY
kind: String (0.0)
sort_key: Real (0.0)
is_link: String (0.0)
is_tunnel: String (0.0)
is_bridge: String (0.0)
railway: String (0.0)
highway: String (0.0)
OGRFeature(SELECT):0
  kind (String) = path
  sort_key (Real) = -9
  is_link (String) = no
  is_tunnel (String) = no
  is_bridge (String) = no
  railway (String) = (null)
  highway (String) = pedestrian
  LINESTRING (26.1027453 44.4303179,26.1024025 44.4300045,26.102255 44.4298696)
```

However, doing the same query from inside a unit test, it seems like it is not using the sqlite dialect, even though there are no errors that say as much. The failing unit test is `test_sql_with_dialect`. It currently fails with:

```
thread 'vector::vector_tests::test_sql_with_dialect' panicked at 'called `Result::unwrap()` on an `Err` value: CplError { class: 3, number: 1, msg: "Undefined function \'NumPoints\' used." }', src/vector/vector_tests/mod.rs:133:73
```

I'm not sure if there's something special that needs to be done at compile time to enable sqlite/spatialite support, and my search-fu has not been very effective. Do you have any ideas about what is needed to get sqlite working?

## Work left
- [ ] Add an entry to `CHANGES.md`
- [ ] Add documentation for `Dataset::execute_sql`
- [ ] Get SQLITE dialect working
- [ ] Add more unit tests around the boundaries of the API: empty result sets, erroneous queries, etc.


michaelkirk:
Sounds neat! Unfortunately I'm not very knowledgeable on spatialite+gdal, though I have used sqlite a bit.

My uninformed guess would be that spatial lite works by binding a bunch of GIS methods via https://www.sqlite.org/c3ref/create_function.html

So, based on the error message, maybe there's some missing setup somewhere that does `create_function` for NumPoints (and presumably others).

Maybe `create_function` is a thing to grep for in ogrinfo.

dead10ck:
Interesting, so it turns out that if I create an actual sqlite table with the `SPATIALITE=YES` option, it works. But the docs mention that sqlite should work for any data source because it will create a virtual SQLite table in memory in order to execute the query. I was able to find out how to turn on debug logging; with that, we can see `ogrinfo` finding spatialite and creating a virtual table:

```sh
✦ ❯ CPL_DEBUG=ON ogrinfo -dialect sqlite -sql "select * from roads where highway = 'pedestrian' and NumPoints(GEOMETRY)
= 3" -spat 26.1017 44.4297 26.1025 44.4303 fixtures/roads.geojson
GeoJSON: First pass: 100.00 %
GDAL: GDALOpen(fixtures/roads.geojson, this=0x560dd18550e0) succeeds as GeoJSON.
INFO: Open of `fixtures/roads.geojson'
      using driver `GeoJSON' successful.
SQLITE: SpatiaLite v4 DB found !
OGR: Unloading VirtualOGR module
SQLITE: SpatiaLite v4 DB found !
OGR2SQLITE: sqlite3_declare_vtab(CREATE TABLE "roads"("kind" VARCHAR,"sort_key" FLOAT,"is_link" VARCHAR,"is_tunnel" VARC
HAR,"is_bridge" VARCHAR,"railway" VARCHAR,"highway" VARCHAR,OGR_STYLE VARCHAR HIDDEN,"GEOMETRY" BLOB_LINESTRING_XY_4326,
 OGR_NATIVE_DATA VARCHAR, OGR_NATIVE_MEDIA_TYPE VARCHAR))
SQLITE: SpatiaLite v4 DB found !
GeoJSON: First pass: 100.00 %
GDAL: GDALOpen(fixtures/roads.geojson, this=0x560dd1b8a500) succeeds as GeoJSON.
GDAL: GDALClose(fixtures/roads.geojson, this=0x560dd1b8a500)
GeoJSON: First pass: 100.00 %
GDAL: GDALOpen(fixtures/roads.geojson, this=0x560dd1e25870) succeeds as GeoJSON.
GDAL: GDALClose(fixtures/roads.geojson, this=0x560dd1e25870)

Layer name: SELECT
Geometry: Line String
SQLITE: Running SELECT COUNT(*) FROM (select * from roads where MBRIntersects("GEOMETRY", BuildMBR(26.101699999990, 44.4
29699999990, 26.102500000010, 44.430300000010)) AND (highway = 'pedestrian' and NumPoints(GEOMETRY) = 3))
Feature Count: 1
Extent: (26.102255, 44.429870) - (26.102745, 44.430318)
GEOGCRS["WGS 84",
    DATUM["World Geodetic System 1984",
        ELLIPSOID["WGS 84",6378137,298.257223563,
            LENGTHUNIT["metre",1]]],
    PRIMEM["Greenwich",0,
        ANGLEUNIT["degree",0.0174532925199433]],
    CS[ellipsoidal,2],
        AXIS["geodetic latitude (Lat)",north,
            ORDER[1],
            ANGLEUNIT["degree",0.0174532925199433]],
        AXIS["geodetic longitude (Lon)",east,
            ORDER[2],
            ANGLEUNIT["degree",0.0174532925199433]],
    ID["EPSG",4326]]
Data axis to CRS axis mapping: 2,1
Geometry Column = GEOMETRY
kind: String (0.0)
sort_key: Real (0.0)
is_link: String (0.0)
is_tunnel: String (0.0)
is_bridge: String (0.0)
railway: String (0.0)
highway: String (0.0)
OGRFeature(SELECT):0
  kind (String) = path
  sort_key (Real) = -9
  is_link (String) = no
  is_tunnel (String) = no
  is_bridge (String) = no
  railway (String) = (null)
  highway (String) = pedestrian
  LINESTRING (26.1027453 44.4303179,26.1024025 44.4300045,26.102255 44.4298696)

SQLite: 2 features read on layer 'SELECT'.
OGR: Unloading VirtualOGR module
GDAL: GDALClose(fixtures/roads.geojson, this=0x560dd18550e0)
GDAL: In GDALDestroy - unloading GDAL shared library.
```

But with debug logging on the test, all we see are statements about opening the data set, then the error happens, then it closes the data set.

```sh
✦ ❯ CPL_DEBUG=ON cargo test --tests vector::vector_tests::sql::test_sql_with_dialect -- --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 0.02s
     Running target/debug/deps/gdal-321fa8db77b70252

running 1 test
GeoJSON: First pass: 100.00 %
GDAL: GDALOpen(fixtures/roads.geojson, this=0x7f61dc07f140) succeeds as GeoJSON.
ERROR 1: Undefined function 'NumPoints' used.
thread 'vector::vector_tests::sql::test_sql_with_dialect' panicked at 'called `Result::unwrap()` on an `Err` value: CplE
rror { class: 3, number: 1, msg: "Undefined function \'NumPoints\' used." }', src/vector/vector_tests/sql.rs:72:10
GDAL: GDALClose(fixtures/roads.geojson, this=0x7f61dc07f140)
test vector::vector_tests::sql::test_sql_with_dialect ... FAILED

failures:

failures:
    vector::vector_tests::sql::test_sql_with_dialect

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 112 filtered out

error: test failed, to rerun pass '--lib'
```

Just taking a poke through [the `ogrinfo` source code](https://github.com/OSGeo/gdal/blob/0941b0f7aaa5bff5fd12a6fed03d4e385c4dcae4/gdal/apps/ogrinfo.cpp#L949), I also don't see it do anything special with the dialect, or any special calls to enable SpatiaLite, before it enters the same API call.

This is making me suspect that it possibly has something to do with how the binary is compiled and/or linked? Maybe it's unable to find SpatiaLite for some reason.

dead10ck:
Okay, I haven't made any luck figuring out the issue with Spatialite. I think in order to make some incremental progress, though, it would be good to merge this anyway, as it's still very useful even without being able to use the sqlite dialect on non-sqlite data sources. We could make a separate issue for it. What do you guys think?

dead10ck:
@michaelkirk what do you think?

src/dataset.rs
```diff
@@ -508,6 +511,84 @@ impl Dataset {
...
+            Some(d) => d.as_ptr(),
+        };
+
+        if let Some(spatial_filter) = spatial_filter {
+            filter_geom = unsafe { spatial_filter.c_geometry() };
+        }
+
+        let c_dataset = unsafe { self.c_dataset() };
+
+        unsafe { gdal_sys::CPLErrorReset() };
```
lnicola, line 569: Is this really needed? From the docs I gather that one needs it call it after handling an error, presumably in the GDAL code.

src/dataset.rs
```diff
@@ -508,6 +511,84 @@ impl Dataset {
...
+    ///         .field("highway")
+    ///         .unwrap()
+    ///         .unwrap()
+    ///         .into_string()
+    ///         .unwrap();
+    ///
+    ///     assert_eq!("pedestrian", highway);
+    /// }
+    /// ```
+    pub fn execute_sql<S: AsRef<str>>(
```
lnicola, line 544: Can you split the non-generic code to a different function? Like this in `File::open`:

```rust
    pub fn open<P: AsRef<Path>>(&self, path: P) -> io::Result<File> {
        self._open(path.as_ref())
    }

    fn _open(&self, path: &Path) -> io::Result<File> {
        fs_imp::File::open(path, &self.0).map(|inner| File { inner })
    }
```

src/vector/feature.rs
```diff
@@ -72,12 +73,12 @@ impl<'a> Feature<'a> {
     /// If the field has an unsupported type, returns a [`GdalError::UnhandledFieldType`].
     ///
     /// If the field is null, returns `None`.
-    pub fn field(&self, name: &str) -> Result<Option<FieldValue>> {
-        let c_name = CString::new(name)?;
+    pub fn field<S: AsRef<str>>(&self, name: S) -> Result<Option<FieldValue>> {
```
lnicola, line 76: Same wrt. generic code.

lnicola:
Regarding the missing `NumPoints` function, I can reproduce the error in Python:

```
>>> from osgeo import ogr
>>> ds = ogr.Open("fixtures/roads.geojson")
>>> ds.ExecuteSQL("SELECT * FROM roads WHERE highway = 'pedestrian' and NumPoints(GEOMETRY) = 3")
ERROR 1: Undefined function 'NumPoints' used.
```

I think we should look into it, but not block merging this PR.

src/dataset.rs
```diff
@@ -508,6 +511,84 @@ impl Dataset {
...
+        let mut filter_geom: OGRGeometryH = std::ptr::null_mut();
+
+        let dialect_c_str = match dialect {
+            None => None,
+            Some(d) => Some(CString::new(d.as_ref())?),
+        };
+
+        let dialect_ptr = match dialect_c_str {
+            None => std::ptr::null(),
+            Some(d) => d.as_ptr(),
```
lnicola, line 560:
```suggestion
            Some(ref d) => d.as_ptr(),
```

Damn, this is a subtle one.

dead10ck: Holy crap, nice find! I actually don't understand this. What's going on here? Why does taking the `CString` by reference produce a different pointer than by value?

(Thanks for the review by the way! I'll address the comments when I have some more time.)

lnicola: Move along, nothing to see here...

---

The problem is that the `Some` branch of the `match` consumes `d` and thus `dialect_c_str` (if you duplicate the `match`, you'll get an error that `d` or `dialect_c_str` was moved out of). Then, `dialect_ptr` is a dangling pointer.

dead10ck: Ahhhh, that is subtle! The dangers of unsafe code. Nice find.

lnicola: Yeah, apparently there's a lint for it, but only for simpler cases: https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#temporary-cstring-as-ptr

src/dataset.rs
dead10ck: This is just defensive, in case for some reason by the time this function is called, some previous error had not been cleared. If this was the case, it would return an error even when the query was successful, and the error would be completely unrelated to the query, which would be very confusing.

lnicola: What I meant is that GDAL itself resets the last error when you call into it through a public API. `GDALOpenEx` does that, the GeoJSON reader does it when needed and so on. You could argue that GDAL itself might forget to call it, but it doesn't seem worth it littering the code with these calls.

dead10ck: Just taking a poke through [`gdaldataset.cpp`](https://github.com/OSGeo/gdal/blob/master/gdal/gcore/gdaldataset.cpp), it appears that `CPLErrorReset` is called in `GDALOpenEx`, and also `GDALDataset::CopyLayer`, but not in `GDALExecuteSQL`. It's possible it is called somewhere further down in the stack.

However, you probably have more experience working with the GDAL code base than me, so if you're not worried, I'll defer to your judgment and remove it.

lnicola: It's actually called -- in your SQLite test -- when SQLite opens the dataset, as it goes through `GDALOpenEx` (I checked using GDB).

I don't really have experience working on GDAL itself but it would make sense for GDAL to clear the last error. On the other hand, the Python bindings seem to call `CPLErrorReset`, so maybe you're right and we should follow suit.

dead10ck: Right, it probably calls it when the data set is opened, but does it call it after it enters the `ExecuteSQL` function? It seems the case where it could cause problems is if the user opens a Dataset, calls one or more other functions, and the last one errors, but the user doesn't check, and then calls `ExecuteSQL`.

Now whether this would happen through the use of this specific crate might be a different matter, since it looks like everywhere that errors calls `_last_cpl_err`, which does reset the error state before returning. Although this does assume that `_last_cpl_err` is called at every site where an error could possibly happen.

I'll leave it up to your guys' judgment.

lnicola: > Right, it probably calls it when the data set is opened, but does it call it after it enters the ExecuteSQL function?

Yes, because it opens the dataset again:

```
#0  0x00007ffff7a749a0 in CPLErrorReset () from /usr/lib/libgdal.so.26
#1  0x00007ffff7582f62 in GDALOpenEx () from /usr/lib/libgdal.so.26
#2  0x00007ffff797a427 in OGROpen () from /usr/lib/libgdal.so.26
#3  0x00007ffff79ae9ef in ?? () from /usr/lib/libgdal.so.26
#4  0x00007ffff4b9ebd2 in sqlite3VdbeExec () from /usr/lib/libsqlite3.so.0
#5  0x00007ffff4ba50b8 in sqlite3_step () from /usr/lib/libsqlite3.so.0
#6  0x00007ffff79a866d in OGRSQLiteTableLayer::EstablishFeatureDefn(char const*) () from /usr/lib/libgdal.so.26
#7  0x00007ffff79a9ca7 in OGRSQLiteTableLayer::Initialize(char const*, int, int) () from /usr/lib/libgdal.so.26
#8  0x00007ffff798983e in OGRSQLiteDataSource::OpenTable(char const*, int) () from /usr/lib/libgdal.so.26
#9  0x00007ffff7989966 in OGRSQLiteDataSource::GetLayerByName(char const*) () from /usr/lib/libgdal.so.26
#10 0x00007ffff79899ea in OGRSQLiteDataSource::GetLayerWithGetSpatialWhereByName(char const*) () from /usr/lib/libgdal.so.26
#11 0x00007ffff799d31b in OGRSQLiteSelectLayerCommonBehaviour::GetBaseLayer(unsigned long&) () from /usr/lib/libgdal.so.26
#12 0x00007ffff799d539 in OGRSQLiteSelectLayerCommonBehaviour::BuildSQL() () from /usr/lib/libgdal.so.26
#13 0x00007ffff799e7e9 in OGRSQLiteSelectLayerCommonBehaviour::SetSpatialFilter(int, OGRGeometry*) () from /usr/lib/libgdal.so.26
#14 0x00007ffff79953e3 in OGRSQLiteExecuteSQL(GDALDataset*, char const*, OGRGeometry*, char const*) () from /usr/lib/libgdal.so.26
```

> Now whether this would happen through the use of this specific crate might be a different matter, since it looks like everywhere that errors calls _last_cpl_err, which does reset the error state before returning. Although this does assume that _last_cpl_err is called at every site where an error could possibly happen.

Yeah, the user could still call `gdal-sys` bindings without resetting the last error.

> I'll leave it up to your guys' judgment.

Sorry for the ping-pong with this, but if the Python bindings reset the last error, I think it's safe to do the same for now.

dead10ck: Ah, right, it would create the virtual table, and then open that as a Dataset. So then the case where this is possible would be non-SQLite dialects.

Okay, I've added it back. I think that takes care of all the review comments.

src/vector/feature.rs
lnicola: This is relatively small, but could still be split into a generic wrapper and a non-generic implementation.

src/dataset.rs
```diff
@@ -508,6 +511,95 @@ impl Dataset {
...
+            gdal_sys::GDALDatasetExecuteSQL(c_dataset, query.as_ptr(), filter_geom, dialect_ptr)
+        };
+
+        let cpl_err = unsafe { gdal_sys::CPLGetLastErrorType() };
+
+        if cpl_err != CPLErr::CE_None {
+            return Err(_last_cpl_err(cpl_err));
+        }
+
+        if c_layer.is_null() {
```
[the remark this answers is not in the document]
dead10ck: I'm not sure, I was also confused by the vagueness of the docs for this function. My understanding is that it is only null if the SQL statement is one which has no result set, i.e. ALTER TABLE, CREATE INDEX, DROP INDEX, etc.

I'm not sure if we can assume that if the pointer is non-null that there was no error. I suppose it's possible for the GDAL code to throw an error in the middle of executing the SQL statement, and I don't know if it still returns a result set in this case. It seems prudent to check for an error regardless of what is returned.

lnicola (approved):

src/vector/feature.rs
dead10ck: Sorry, I had actually fixed this, but it got lost between devices in a `git commit --amend`, sorry about that. The fix is back again.

src/dataset.rs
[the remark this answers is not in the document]
lnicola: That sentence seems pretty unambiguous to me.

It might be also possible that GDAL encounters an error it can handle, uses a fallback path and returns a valid result, but doesn't reset the last error. But let's leave it like this, the Python bindings check for errors first.

Who knew that error checking could be so complicated?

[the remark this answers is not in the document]
dead10ck: I mean we are talking about a code base that mixes C, C++, and Rust. Global error state and inconsistent call site checking is pretty par for the course with C lol. That's why I tend to favor defensive code, even at the expense of repetition and readability. In my experience, you can't rely on other code to do the right thing 100% of the time.

dead10ck:
I pushed up a minor change: after a bit more reading, it seems `CString` can be made from any `T: Into<Vec<u8>>`, so I changed the generic type constraints to that instead of `AsRef<str>`. This way, if the caller passes in an owned `String` or any other owned value, it does not need to make a copy.

jdroenner:
Just let me know when the PR is ready :D
I trust lnicola's review and will merge it then.

lnicola:
There's a conflict in `src/dataset.rs`, probably some `use` clauses.

dead10ck:
I rebased and reorganized the uses a bit.

lnicola:
I think this is as good as it will get for now, so it can be merged.

src/dataset.rs
```diff
@@ -535,6 +537,95 @@ impl Dataset {
         }
         Ok(Transaction::new(self))
     }
+
+    /// Execute a SQL query against the Dataset. It is equivalent to calling
+    /// [`GDALDatasetExecuteSQL`](https://gdal.org/api/raster_c_api.html#_CPPv421GDALDatasetExecuteSQL12GDALDatasetHPKc12OGRGeometryHPKc).
+    /// Returns a `ResultSet`, which can be treated just as any other [`Layer`], or `None` if the
+    /// query does not have a `ResultSet`, such as `ALTER TABLE`, etc. Note that this is different
+    /// than a query which has an *empty* `ResultSet`.
```
michaelkirk, line 545: To make sure I understand:

Something like `Select * where false` would return an empty `Some(result_set)` where `result_set.len() == 0` whereas `Alter Table ...` would return `None` right?


At a minimum, I have a grammar nit: "different than a query" should be "different __from__ a query" or "distinct from a query"

Alternatively, maybe this phrasing is clearer:

> Returns a `ResultSet`, which can be treated just as any other [`Layer`]. 
>
> Queries, such as `ALTER TABLE`, etc. have no ResultSet, and return `None`, which is distinct from an empty `ResultSet`.

src/vector/mod.rs
```diff
@@ -22,13 +22,15 @@ mod geo_to_gdal;
...
 mod ops;
+mod sql;
 
 pub use defn::{Defn, Field, FieldIterator};
 pub use feature::{Feature, FieldValue, FieldValueIterator};
 pub use gdal_sys::{OGRFieldType, OGRwkbGeometryType};
 pub use geometry::Geometry;
 pub use layer::{FeatureIterator, FieldDefn, Layer};
 pub use ops::GeometryIntersection;
+pub use sql::ResultSet;
```
michaelkirk, line 33: This is a pretty general name. Should this maybe be `use sql::ResultSet as SqlResultSet` or something to namespace it in the crate namespace?

Just a minor thought, feel free to ignore this feedback.

src/vector/vector_tests/sql.rs
```diff
@@ -0,0 +1,119 @@
...
+        .into_string()
+        .unwrap();
+
+    assert_eq!("pedestrian", highway);
+}
+
+#[test]
+fn test_sql_empty_result() {
+    let ds = Dataset::open(fixture!("roads.geojson")).unwrap();
+    let query = "SELECT kind, is_bridge, highway FROM roads WHERE highway = 'jazz hands 👐'";
```
michaelkirk, line 90: Some excellent test data in this PR. 👏👏👏

src/dataset.rs
```diff
@@ -535,6 +537,95 @@ impl Dataset {
...
+    ///         .into_string()
+    ///         .unwrap();
+    ///
+    ///     assert_eq!("pedestrian", highway);
+    /// }
+    /// ```
+    pub fn execute_sql<S: AsRef<str>>(
+        &self,
+        query: S,
+        spatial_filter: Option<&Geometry>,
```
michaelkirk, line 574: Can you document the spatial_filter and dialect params?

dead10ck:
Thanks so much for reviewing, @michaelkirk! Will address when I get some time in the next couple days.

src/dataset.rs
dead10ck: This is correct. I've made the suggested changes to the wording.

src/vector/mod.rs
```diff
@@ -22,6 +22,7 @@ mod geo_to_gdal;
 mod geometry;
 mod layer;
 mod ops;
+pub mod sql;
```
dead10ck, line 25: Sorry, GitHub is not letting me comment on your original comment now, but I actually had a similar thought. I changed it instead to just make the `sql` module public, that way users can choose to namespace as `sql::ResultSet` if they desire. How does this sound? (Still open to feedback, I just changed it because of the asynchronicity of these reviews.)

dead10ck (commented):
As I was documenting the `dialect` parameter, I realized that since there are only two dialects possible, there really isn't any reason to accept an arbitrary string. At the risk of making this PR take longer, instead I made an enum and let the user choose from the possible choices. What do you guys think?

lnicola:
> At the risk of making this PR take longer, instead I made an enum and let the user choose from the possible choices. What do you guys think?

Yeah, somehow I don't see GDAL adding new ones too soon. An enum would work well.

dead10ck:
@lnicola cool, I'd appreciate if you could look over the changes when you get some spare time 🙂

src/dataset.rs
```diff
@@ -535,6 +537,104 @@ impl Dataset {
...
+        dialect: sql::Dialect,
+    ) -> Result<Option<sql::ResultSet>> {
+        let query = CString::new(query.as_ref())?;
+
+        let dialect_c_str = match dialect {
+            sql::Dialect::DEFAULT => None,
+            sql::Dialect::OGR => Some(CString::new(sql::OGRSQL).unwrap()),
```
lnicola, line 589: Can we use one of those cstring crates to simplify this?

lnicola:
Looks good to me, but I don't have commit rights, so.. 😛

src/dataset.rs
dead10ck: I changed it to use a const byte slice and `CStr::from_bytes_with_nul_unchecked` to save the unnecessary copies.

dead10ck:
As for unnecessary copies, just as a side note, I tried to see if we could change the generic type constraint on `Dataset::execute_sql` and `Feature::field` to `Into<Vec<u8>>` since that's what the constraints are on `CString::new`, and doing so avoids unnecessary copies when the user passes in an owned value, but for some reason, it breaks the ability to coerce `&String` to `&str`, so it would be a breaking change in the case of `Feature::field`, so I opted against it.

Another possibility could be to make variants of these functions that let the caller pass in a `&CStr`, which would give them the option of avoiding copies altogether if they don't mind the inconvenience of constructing it. This might help e.g. with use cases that just have static SQL queries. This wouldn't necessarily need to go in this PR though.

lnicola:
That's not something I would worry about.

src/vector/mod.rs
michaelkirk: sounds good to me

src/dataset.rs
```diff
@@ -535,6 +542,106 @@ impl Dataset {
...
+    /// [`GDALDatasetExecuteSQL`](https://gdal.org/api/raster_c_api.html#_CPPv421GDALDatasetExecuteSQL12GDALDatasetHPKc12OGRGeometryHPKc).
+    /// Returns a [`sql::ResultSet`], which can be treated just as any other [`Layer`].
+    ///
+    /// Queries such as `ALTER TABLE`, `CREATE INDEX`, etc. have no [`sql::ResultSet`], and return
+    /// `None`, which is distinct from an empty [`sql::ResultSet`].
+    ///
+    /// # Arguments
+    ///
+    /// * `query`: The SQL query
+    /// * `spatial_filter`: Limit results of the query to features that are inside the given
```
michaelkirk, line 556: GDAL doesn't make it clear in the docs I saw, but do you know if "inside the given geometry" means "completely within" or is it only required to intersect?

src/vector/sql.rs
```diff
@@ -0,0 +1,46 @@
...
+
+impl<'a> Drop for ResultSet<'a> {
+    fn drop(&mut self) {
+        unsafe { gdal_sys::GDALDatasetReleaseResultSet(self.dataset, self.layer.c_layer()) };
+    }
+}
+
+/// Represents valid SQL dialects to use in SQL queries. See
+/// <https://gdal.org/user/ogr_sql_sqlite_dialect.html>
+pub enum Dialect {
```
michaelkirk, line 32: Better! Thanks.

src/dataset.rs
dead10ck: Good point. I don't have any concrete examples, but I do believe it's an intersection, rather than complete containment. For example, if a bbox has a long road go through it, but is not completely contained in it, it still gets returned. I was able to find [this bit](https://gdal.org/user/vector_data_model.html?highlight=spatial%20filter#layer) in the GDAL docs:

> The OGRLayer includes methods for sequential and random reading and writing. Read access (via the `OGRLayer::GetNextFeature()` method) normally reads all features, one at a time sequentially; however, it can be limited to return features **intersecting** a particular geographic region by installing a spatial filter on the OGRLayer (via the `OGRLayer::SetSpatialFilter()` method).

I'll update the language to say "intersecting the given geometry".

michaelkirk (approved):
This all LGTM, thank you for your patience with the review process!

PR #155, opened 2021-01-30
Support extended flags for open_ex

### LINKED ISSUE #154: Missing Gdal dataset open flags
The `gdal::open_ex` open_flags allow only value from the `GDALAccess::Type` values. 

This prevent using the others `GDAL_OF_XXX` flags combination.  

They are preprocessor `#define`  in gdal.h and thus skipped by bindgen; they are still essential to control some opening behaviors like raster/vector driver selection, or multidimensional raster or gnm drivers.

Maybe a crate like [bitflags](https://docs.rs/bitflags/1.2.1/bitflags/) could help.

dmarteau:
---

This is attempt to fix #154 

This is an api breaking change

This PR allow using `GDAL_OF_XXX` extended flags. 

Implement a `DatasetOptions` to pass to `Dataset::open_ex`

```rust
    use gdal::{ Dataset, DatasetOptions }

    let dataset = Dataset::open_ex(
        "roads.geojson",
        DatasetOptions { 
            open_flags: GdalOpenFlags::GDAL_OF_UPDATE|GdalOpenFlags::GDAL_OF_VECTOR,
            ..DatasetOptions::default()
       }
    )
     .unwrap();
```

    `GDALAccess` values are supported using [`From`] implementation

```rust
    Dataset::open_ex(
        "roads.geojson",
        DatasetOptions {
            open_flags: GDALAccess::GA_Update.into(),
            ..DatasetOptions::default()
        },
    )
    .unwrap();
```

Note that [following this note](https://github.com/georust/gdal/issues/154#issuecomment-770187560)  `GDAL_OF_SHARED` option is removed from the set of allowed option because it subverts the [`Send`] implementation that allow passing the dataset the another thread.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,35 @@
 # Changes
 
 ## Unreleased
+* **Breaking**: Use `DatasetOptions` to pass as `Dataset::open_ex` parameters and
+    add support for extended open flags.
+
+    ```rust
+        use gdal::{ Dataset, DatasetOptions }
+
+        let dataset = Dataset::open_ex(
+            "roads.geojson",
+            DatasetOptions { 
+                open_flags: GdalOpenFlags::GDAL_OF_UPDATE|GdalOpenFlags::GDAL_OF_VECTOR,
+                ..DatasetOptions::default()
+            }
+        )
+        .unwrap();
+    ```
+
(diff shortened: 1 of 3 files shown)


dmarteau:
Remove the `GDAL_OF_SHARED` option and added a note.

PR #158, opened 2021-02-02
Add documentation

rmanoka:
Added documentations for `Dataset` and `RasterBand`, and fixed `cfg()` for some of the `open_ex` flags (it wasn't enabled on gdal 3.0 and above).

Open question:  how to refer to GDAL documentations from our docs?  For many functions, the relevant docs are doxygen links (eg. [GDALOpenEx](https://gdal.org/doxygen/gdal_8h.html#a9cb8585d0b3c16726b08e25bcc94274a)) which could break if it is regenerated upstream (and annoying to update).

---
---

### DIFF
--- a/src/dataset.rs
+++ b/src/dataset.rs
@@ -16,9 +16,23 @@ use std::convert::TryInto;
 
 use bitflags::bitflags;
 
+/// A 2-D affine transform mapping pixel coordiates to world
+/// coordinates. See [GDALGetGeoTransform] for more details.
+///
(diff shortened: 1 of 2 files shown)


jdroenner:
cool!more  documentation is always good. I don't have an answer to the doxygen links. I guess we could link the doxygen startpage and leave the user to find the method?

rmanoka:
Broken links lead to the start page (via a 404) anyway;  hopefully the text of the link tells the user what to search for.  Let's take a call when any of these links break.

michaelkirk:
Excellent - Thank you @rmanoka!

PR #160, opened 2021-02-04
Implement wrapper for OGR_L_TestCapability

dmarteau:
---

* Map `OLCxxx GDAL` preprocessor defines to  `LayerCaps::OLCxxx` constants.
* Add wrapper for `OGR_L_TestGetCapability` as `Layer::test_capability(&self,  capability: LayerCaps::Type) -> Result<bool>`

### DIFF
--- a/src/vector/layer.rs
+++ b/src/vector/layer.rs
@@ -13,6 +13,51 @@ use std::{convert::TryInto, ffi::CString, marker::PhantomData};
 
 use crate::errors::*;
 
+/// Layer capabilities, map GDAL defines
+#[allow(non_upper_case_globals)]
+#[allow(non_snake_case)]
+pub mod LayerCaps {
+
+    pub type Type = &'static str;
+
+    /// Layer capability for random read
+    pub const OLCRandomRead: Type = "RandomRead";
+    /// Layer capability for sequential write
+    pub const OLCSequentialWrite: Type = "SequentialWrite";
+    /// Layer capability for random write
+    pub const OLCRandomWrite: Type = "RandomWrite";
+    /// Layer capability for fast spatial filter
+    pub const OLCFastSpatialFilter: Type = "FastSpatialFilter";
+    /// Layer capability for fast feature count retrieval
+    pub const OLCFastFeatureCount: Type = "FastFeatureCount";
+    /// Layer capability for fast extent retrieval
+    pub const OLCFastGetExtent: Type = "FastGetExtent";
+    /// Layer capability for field creation
+    pub const OLCCreateField: Type = "CreateField";
+    /// Layer capability for field deletion
+    pub const OLCDeleteField: Type = "DeleteField";
+    /// Layer capability for field reordering
+    pub const OLCReorderFields: Type = "ReorderFields";
+    /// Layer capability for field alteration
+    pub const OLCAlterFieldDefn: Type = "AlterFieldDefn";
+    /// Layer capability for transactions
+    pub const OLCTransactions: Type = "Transactions";
+    /// Layer capability for feature deletiond
+    pub const OLCDeleteFeature: Type = "DeleteFeature";
+    /// Layer capability for setting next feature index
+    pub const OLCFastSetNextByIndex: Type = "FastSetNextByIndex";
+    /// Layer capability for strings returned with UTF-8 encoding
+    pub const OLCStringsAsUTF8: Type = "StringsAsUTF8";
+    /// Layer capability for field ignoring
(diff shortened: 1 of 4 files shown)


src/vector/layer.rs
```diff
@@ -101,6 +146,13 @@ impl<'a> Layer<'a> {
         _string(rv)
     }
 
+    pub fn test_capability(&self, capability: LayerCaps::Type) -> Result<bool> {
```
rmanoka, line 149: A couple of minor comments:
- the usual rust convention is to name it `has_capability`
- the argument is of type `&'static str` and hence prevents for eg. testing a string input by the user at runtime.  Would `&'a str` make sense for the method?

rmanoka (changes requested):
Just a couple of minor suggestions.

src/vector/layer.rs
dmarteau: > the usual rust convention is to name it has_capability

Changed to `has_capability`.

dmarteau: > the argument is of type &'static str and hence prevents for eg. testing a string input by the user at runtime. Would &'a str make sense for the method?

I don't know: if it was a pure Rust implementation I would have used an `Enum LayerCaps` and thus enforcing the use of a dedicated type as parameter (and  I'm not sure even why this is a string in gdal ).  

Formally, that does not prevent testing from string input but require a little more work: you need to write a proper parser for the string depending the input come from and/or eventually deal with typo error - which is never obvious to debug.  

I'm  in favor of narrowing the possibilities for safety but I don't mind to change to `&str` if you think it may be useful in other use case.

rmanoka: I think it's okay to leave as `&'static str` unless a need to check for dynamic / user-input arbitarry strings arises.

rmanoka (approved):

src/vector/layer.rs
```diff
@@ -101,6 +146,13 @@ impl<'a> Layer<'a> {
         _string(rv)
     }
 
+    pub fn has_capability(&self, capability: LayerCaps::Type) -> Result<bool> {
+        let rv = unsafe {
+            gdal_sys::OGR_L_TestCapability(self.c_layer, CString::new(capability)?.as_ptr()) == 1
```
michaelkirk, line 151: When would this fail? If you do go the way of only accepting an enum of input strings, I'd argue you could `unwrap` *here* and offer a nicer API to your users: `bool` instead of `Result<bool>`.

I think we're better equipped to carefully scrutinize `unwrap` code centrally in a library like this, and save the attention of downstream users who now will have to ask "will *their* code explode" whenever they use this method.

Feel free to ignore this suggestion, reasonable minds may disagree. =)

dmarteau: I'd love to get rid of the Result<>, but the input type is actually a alias to '&'static str' so it does not prevent to use invalid candidate input that may fail the conversion to `CString`. If everybody think that the intention notice is ok, I will happily go with `unwrap`.

rmanoka: I don't think `unwrap()` in the present implementation is acceptable.  I believe what @michaelkirk suggested is to accept a enum as input rather than `&str` in which case we could even get rid of the `&str -> &Cstr` conversion.  That is a lot of boilerplate (unless there's a nice macro crate), so we could postpone it to a later PR.

michaelkirk: If the input type doesn't forbid passing in something that would explode, I would personally be in favor of keeping the `Result` as is.

rmanoka: So, if we could somehow define the `const`s in `mod LayerCaps` directly as `&'static Cstr`, then we can avoid both the extra allocation to convert `&str` to `&Cstr`, as well as the `Result<>`.  However, it seems to need some complicated macro:  see [SO post](https://stackoverflow.com/questions/58349489/const-static-cstr).

michaelkirk: Oh wow, good find. Looks fun.

dmarteau: Implemented the `enum` alternative solution, so one may compare.

dmarteau: The `enum`  alternative does not save the conversion but limit  the input to acceptable values so conversion to CStr will not return an error.

michaelkirk: Thanks for putting that together @dmarteau! I personally prefer the #162 version.

dmarteau: So, should we close that PR in favor of https://github.com/georust/gdal/pull/162 ?

jdroenner:
i merged the enum version.

PR #161, opened 2021-02-04
Use mutable references as per semantics

### LINKED ISSUE #150: Should mutable borrow of RasterBand require mutable borrow of it's Dataset?
Open Question from #144

### LINKED ISSUE #159: Feature reading not reseted between two FeatureIterator instanciation
Between two calls of  `Layer::features(&self) -> FeatureIterator`, there is no call to `OGR_L_ResetReading`:  the consequence is that features cannot be iterated twice from a Layer and the Layer has to be recreated.

Ex 
```
let mut ds = Dataset::open(fixture!("roads.geojson")).unwrap();
let layer = ds.layer(0).unwrap();

assert_eq!(layer.features().count(), 21);  <- Ok
assert_eq!(layer.features().count(), 21);  <- Fail, gives 0 !
```

As the Layer is not consummed by the iterator and taken from an immutable layer, the internal state should not be changed and it should be possible to get a new iterator  over all features. 

Since iterating features change the Internal state of a GDAL layer and having two iterator at the same time on the same layer is hazardous, may be the Iterator should be extracted from a mutable layer: `Layer::features(&mut self) -> FeatureIterator`

see https://gdal.org/api/vector_c_api.html#_CPPv418OGR_L_ResetReading9OGRLayerH

rmanoka:
---

This builds on #158 and also adds quite a few breaking changes.

closes #159 
closes #150


rmanoka:
In this PR we add a (commented out) test that ensures that the semantics we enforce indeed leads to a compile error if used incorrectly:

```rust
fn test_features_mut_lifetime_enforce() {
    with_layer("roads.geojson", |mut layer| {
        for _ in layer.features() {
            // the next line is a compile error as layer is mut borrowed above
            let _ = layer.defn();
        }
    });
}
```

Should we add [trybuild](https://docs.rs/trybuild/1.0.40/trybuild/) as a dev-dependency so that we can express this as a regular test?  The crate does add a few heavy deps (syn, rustc), but it is only a dev dependency and may be okay.  I think it'll be useful in this crate as we try to ensure borrow semantics from an underlying library that doesn't have those.

jdroenner:
it would be fine for me

michaelkirk:
This makes a lot of sense to me - I think it's good to have the compiler enforce "rust like" API usage.

Taking it a step further, do you think it would be worthwhile for things like:
```
pub fn layer_by_name(&self, name: &str) -> Result<Layer> {
```

To propagate some kind of lifetime:
```
pub fn layer_by_name<'a>(&'a self, name: &str) -> Result<Layer<'a>> {
```

To clarify that `&self` is still "borrowed" as long as Layer exists?

rmanoka:
@michaelkirk 

> To clarify that &self is still "borrowed" as long as Layer exists?

I think it is already implicitly inferred.  See `Dataset::child_layer` which connects it to the `Layer`.

michaelkirk:
Oh you're absolutely right, sorry.

src/vector/vector_tests/mod.rs
```diff
@@ -491,4 +491,30 @@ mod tests {
...
+    #[test]
+    fn test_features_reset() {
+        with_layer("roads.geojson", |mut layer| {
+            assert_eq!(layer.features().count(), layer.features().count(),);
+        });
+    }
+
+    #[test]
+    fn test_features_aliasing_compile_fail() {
+        let t = trybuild::TestCases::new(); // A compilation test that should fail.
```
michaelkirk, line 504: Cool to see `trybuild`! 

I've never seen tests like this outside of the actual compiler dev, but it makes sense!

rmanoka: It's a bit of a heavy dep., but seems useful for this crate as we try to impose rust semantics on ffi libs.  I figure we could always comment it out, and remove the deps if CI takes too long.

src/vector/vector_tests/mod.rs
```diff
@@ -491,4 +491,30 @@ mod tests {
...
+        t.compile_fail("tests/ui/01-features-aliasing-errors.rs");
```
jdroenner, line 505: why is the tests folder called "ui"?

jdroenner (commented):
i think we should merge it with the trybuild. I have only one question left :)

src/vector/vector_tests/mod.rs
rmanoka: Not sure;  I just followed the instructions from the the trybuild crate docs.  Not having the `ui/` folder caused some errors as (I think) cargo tried to run them as normal tests, and failed.  I didn't try a different folder name though.

rmanoka: It seems that any folder inside `tests/` works;  have renamed it to `compile-tests`

jdroenner:
uh i broke it by merging the other PR -_-

rmanoka:
@jdroenner Not sure what the failure is, but I'll take a closer look next week and try to resolve it.

PR #162, opened 2021-02-04
Implement wrapper for OGR_L_TestCapability - enum implementation

dmarteau:
---

Alternative implementation of https://github.com/georust/gdal/pull/160 using `enum`.

### DIFF
--- a/src/vector/layer.rs
+++ b/src/vector/layer.rs
@@ -13,6 +13,73 @@ use std::{convert::TryInto, ffi::CString, marker::PhantomData};
 
 use crate::errors::*;
 
+/// Layer capabilities
+pub enum LayerCaps {
+    /// Layer capability for random read
+    OLCRandomRead,
+    /// Layer capability for sequential write
+    OLCSequentialWrite,
+    /// Layer capability for random write
+    OLCRandomWrite,
+    /// Layer capability for fast spatial filter
+    OLCFastSpatialFilter,
+    /// Layer capability for fast feature count retrieval
+    OLCFastFeatureCount,
+    /// Layer capability for fast extent retrieval
+    OLCFastGetExtent,
+    /// Layer capability for field creation
+    OLCCreateField,
+    /// Layer capability for field deletion
+    OLCDeleteField,
+    /// Layer capability for field reordering
+    OLCReorderFields,
+    /// Layer capability for field alteration
+    OLCAlterFieldDefn,
+    /// Layer capability for transactions
+    OLCTransactions,
+    /// Layer capability for feature deletiond
+    OLCDeleteFeature,
+    /// Layer capability for setting next feature index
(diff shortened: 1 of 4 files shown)


src/vector/layer.rs
```diff
@@ -13,6 +13,73 @@ use std::{convert::TryInto, ffi::CString, marker::PhantomData};
 
 use crate::errors::*;
 
+/// Layer capabilities
+pub enum LayerCaps {
+    /// Layer capability for random read
+    OLCRandomRead,
```
michaelkirk, line 19:
```suggestion
    RandomRead,
```

Just a style nit, but since this is already namespaced under `Layer::LayerCaps`, do we get anything with the uniformly repeated `OLC` prefix for each case?

src/vector/layer.rs
```diff
@@ -13,6 +13,73 @@ use std::{convert::TryInto, ffi::CString, marker::PhantomData};
...
+            Self::OLCTransactions => "Transactions",
+            Self::OLCDeleteFeature => "DeleteFeature",
+            Self::OLCFastSetNextByIndex => "FastSetNextByIndex",
+            Self::OLCStringsAsUTF8 => "StringsAsUTF8",
+            Self::OLCIgnoreFields => "IgnoreFields",
+            Self::OLCCreateGeomField => "CreateGeomField",
+            Self::OLCCurveGeometries => "CurveGeometries",
+            Self::OLCMeasuredGeometries => "MeasuredGeometries",
+        })
+        .unwrap()
```
michaelkirk, line 79: I think judicious use of `unwrap` like this in carefully audited code is a net win for library users. 👍

src/vector/layer.rs
dmarteau: Oh, the idea was to simply to stick to the Gdal original naming.

src/vector/layer.rs
lnicola: There are some crates for `CString` literals, we could use one of them.

src/vector/layer.rs
lnicola: :+1: for dropping the prefix, although we don't really have prior art for this.

src/vector/layer.rs
dmarteau: @lnicola If you are refering to [this discussion](https://stackoverflow.com/questions/58349489/const-static-cstr), I don't see a big win using some complicated macro.

src/vector/layer.rs
dmarteau: I'm really not sure with that: this would be inconsistent with other refererences to gdal constants like `OGRwkbGeometryType` or other type imported from `gdal-sys`. Imho, constant naming should stick to Gdal name or be renamed for all constant definition, but not a mix of both naming convention.

PR #164, opened 2021-02-17
Add Dataset::build_overviews

lnicola:
---

### DIFF
--- a/src/dataset.rs
+++ b/src/dataset.rs
@@ -146,7 +146,9 @@ impl Dataset {
     }
 
     /// Open a dataset with extended options. See
-    /// [GDALOpenEx].
+    /// [`GDALOpenEx`].
+    ///
+    /// [`GDALOpenEx`]: https://gdal.org/doxygen/gdal_8h.html#a9cb8585d0b3c16726b08e25bcc94274a
     pub fn open_ex(path: &Path, options: DatasetOptions) -> Result<Dataset> {
         _register_drivers();
         let filename = path.to_string_lossy();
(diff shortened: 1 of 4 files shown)


src/dataset.rs
```diff
@@ -315,6 +317,39 @@ impl Dataset {
...
+    /// Builds overviews for the current `Dataset`. See [`GDALBuildOverviews`].
+    ///
+    /// # Arguments
+    /// * `resampling` - resampling method, as accepted by GDAL, e.g. `"CUBIC"`
+    /// * `overviews` - list of overview decimation factors, e.g. `&[2, 4, 8, 16, 32]`
+    /// * `bands` - list of bands to build the overviews for, or empty for all bands
+    ///
+    /// [`GDALBuildOverviews`]: https://gdal.org/doxygen/gdal_8h.html#a767f4456a6249594ee18ea53f68b7e80
+    pub fn build_overviews(
+        &self,
```
michaelkirk, line 329: In line with https://github.com/georust/gdal/pull/161, I wonder if @rmanoka or @jdroenner have thoughts about whether this should be &mut.

frewsxcv, line 329: Should this take `&mut self` if it's modifying `self.c_dataset`?

frewsxcv (approved):
Besides my comment, lgtm!

src/dataset.rs
frewsxcv: Lol @michaelkirk just commented about the same thing :)

lnicola:
Updated.

Unrelated, tests fail intermittently on my computer:

  process didn't exit successfully: `/home/grayshade/Projects/georust/gdal/target/debug/deps/gdal-4103840a8afd223b` (signal: 11, SIGSEGV: invalid memory reference)

src/dataset.rs
jdroenner: i guess &mut is the better choice.

PR #166, opened 2021-02-23
Avoid calling GDALAllRegister twice

lnicola:
---

Fixes https://github.com/georust/gdal/pull/164#issuecomment-783607743


jdroenner:
uh i guess this is a leftover from when raster and vector datasets where separated. Good catch!

PR #167, opened 2021-02-23
ASan

frewsxcv:
---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
         with:
           toolchain: nightly
           default: true
-          components: "clippy, rustfmt"
+          components: "clippy, rustfmt, rust-src"
       - name: Check with Rustfmt
         uses: actions-rs/cargo@v1
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,8 @@ jobs:
     name: "ci ubuntu-20.04"
     runs-on: "ubuntu-20.04"
     env:
-      RUSTFLAGS: "-Zsanitizer=address RUSTDOCFLAGS=-Zsanitizer=address"
+      RUSTFLAGS: "-Zsanitizer=address"
+      RUSTDOCFLAGS: "-Zsanitizer=address"
     
     strategy:
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,8 +15,8 @@ jobs:
     name: "ci ubuntu-20.04"
     runs-on: "ubuntu-20.04"
     env:
-      RUSTFLAGS: "-Zsanitizer=address"
-      RUSTDOCFLAGS: "-Zsanitizer=address"
+      RUSTFLAGS: "-Zsanitizer=address -Zbuild-std "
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -33,23 +39,28 @@ jobs:
...
         uses: actions-rs/cargo@v1
         with:
           command: fmt
           args: --all -- --check
-      - name: Check with Clippy
-        uses: actions-rs/clippy-check@v1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-targets ${{ matrix.features }} -- -D warnings
+#       - name: Check with Clippy
```
michaelkirk, line 52: Why did we lose the clippy check?

.github/workflows/ci.yml
```diff
@@ -33,23 +39,28 @@ jobs:
...
+#         uses: actions-rs/clippy-check@v1
+#         with:
+#           token: ${{ secrets.GITHUB_TOKEN }}
+#           args: --all-targets ${{ matrix.features }} -- -D warnings
       - name: Build
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: ${{ matrix.features }} --verbose
+          args: ${{ matrix.features }} --verbose --target x86_64-unknown-linux-gnu -Zbuild-std
```
michaelkirk, line 61: why do we need to specify these build args?

frewsxcv: * `--target ...`: see https://github.com/rust-lang/rust/issues/48199 for some context. the last comment sorta explains it, but i don't really get it. effectively, building with sanitizers will fail without `--target` if a crate uses a build script or procedural macros
* `-Zbuild-std`: this rebuilds rust's standard library. it's not necessary for address sanitizers, so we can remove it here. but you need it for the memory sanitizer ([source](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html)) so it can accurately instrument all sources of allocations. but as long as we're using a prebuilt gdal in our tests here, allocations from gdal won't be instrumented so we can't do that anyways. in [some past runs in this pull request](https://github.com/georust/gdal/runs/1961853656), you can see how the sanitizer thinks there's a bunch of memory that's been leaked, presumably because gdal is not instrumented with sanitizers

frewsxcv (commented):
> Why did we lose the clippy check?

i brought back the clippy check in the latest commit, if you wanna see the errors that get returned. maybe there's new rules in nightly rust? not sure

michaelkirk:
>> Why did we lose the clippy check?
>>
> i brought back the clippy check in the latest commit, if you wanna see the errors that get returned. maybe there's new rules in nightly rust? not sure

Ah, so we can't run ASan without nightly?

frewsxcv:
> > > Why did we lose the clippy check?
> > 
> > 
> > i brought back the clippy check in the latest commit, if you wanna see the errors that get returned. maybe there's new rules in nightly rust? not sure
> 
> Ah, so we can't run ASan without nightly?

yep, that's the biggest downside. all `-Z` compiler options require nightly. the effects of that could be minimized by locking to a specific nightly, but still not ideal

metasim:
Reenvisioned and deployed here: https://github.com/georust/gdal/pull/398

PR #172, opened 2021-03-10
Fix memory leak in Geometry::from_wkt

ttencate:
---


lnicola:
CI fails because of a formatting difference.

src/vector/geometry.rs
```diff
@@ -75,9 +75,11 @@ impl Geometry {
     /// [WKT](https://en.wikipedia.org/wiki/Well-known_text) string.
     pub fn from_wkt(wkt: &str) -> Result<Geometry> {
         let c_wkt = CString::new(wkt)?;
-        let mut c_wkt_ptr = c_wkt.into_raw();
+        let c_wkt_ptr = c_wkt.into_raw();
+        let mut c_wkt_end_ptr = c_wkt_ptr; // The pointer is updated to point just beyond the consumed bytes.
         let mut c_geom = null_mut();
-        let rv = unsafe { gdal_sys::OGR_G_CreateFromWkt(&mut c_wkt_ptr, null_mut(), &mut c_geom) };
+        let rv = unsafe { gdal_sys::OGR_G_CreateFromWkt(&mut c_wkt_end_ptr, null_mut(), &mut c_geom) };
```
lnicola, line 81: Doesn't

```rust
let mut c_wkt_end_ptr = c_wkt.as_ptr()
```

work?

ttencate: That's much better, thanks!

lnicola:
Interesting, this is the only `into_raw()` call in the entire crate.

PR #177, opened 2021-03-29
GCP Transformer and create_and_reproject warp.

michaeljfazio:
- [X ] I agree to follow the project's [code of conduct](https://github.com/georust/gdal/blob/master/CODE_OF_CONDUCT.md).
- [X ] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

This change adds a wrapper for the GDAL Transformer abstraction. An initial implementation of the [GCP based polynomial transformer](https://gdal.org/api/gdal_alg.html) has also been added. Future versions should look to extend the Transformer abstraction to include other implementations such as:

* [GDALCreateGenImgProjTransformer](https://gdal.org/api/gdal_alg.html#_CPPv431GDALCreateGenImgProjTransformer12GDALDatasetHPKc12GDALDatasetHPKcidi)
* [GDALCreateReprojectionTransformerEx](https://gdal.org/api/gdal_alg.html#_CPPv435GDALCreateReprojectionTransformerEx20OGRSpatialReferenceH20OGRSpatialReferenceHPPCKc)
* [GDALCreateApproxTransformer](https://gdal.org/api/gdal_alg.html#_CPPv427GDALCreateApproxTransformer19GDALTransformerFuncPvd)

Also included in the change set is an implementation of [GDALCreateAndReprojectImage](https://gdal.org/doxygen/gdalwarper_8h.html#a71855a3db3fc77d37384c274333fee3d) function.


michaeljfazio:
I've rebased onto master for this commit. Master appears to currently been broken which is failing the build.

michaeljfazio:
Fixed all clippy issues. Should be good to go now.

src/raster/warp.rs
```diff
@@ -25,3 +29,36 @@ pub fn reproject(src: &Dataset, dst: &Dataset) -> Result<()> {
...
+    let psz_dst_filename = dst_path
+        .to_str()
+        .expect("Destination path must be supplied.");
+    let psz_dst_wkt = dst_srs.to_wkt().expect("Failed to obtain WKT for SRS.");
+
+    let rv = unsafe {
+        gdal_sys::GDALCreateAndReprojectImage(
+            src.c_dataset(),
+            null(),
+            CString::new(psz_dst_filename)?.as_ptr(),
```
lnicola, line 48: This is a use-after-free: the `CString` is destroyed before `GDALCreateAndReprojectImage` runs.

src/alg/transform.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// Extracts the Ground Control Points from the specified dataset.
+    ///
+    /// # Arguments
+    ///
+    /// * `dataset` - The `Dataset` to extract Ground Control Points from.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// use std::path::Path;
```
metasim, line 74: Does `rustdoc` still compile this without `rust` in fence suffix. i.e. does this need to be `rust, no_run`?

src/alg/transform.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// use gdal::Dataset;
+    /// use gdal::alg::transform::Transformer;
+    ///
+    /// let path = Path::new("/path/to/dataset");
```
metasim, line 78: Can we use a file in `fixtures` to demo this? If no, can we construct a very small example file with GCPs and save it there?

src/alg/transform.rs
```diff
@@ -0,0 +1,184 @@
...
+    Second = 2,
+    Third = 3,
+}
+/// Transformer used to map between pixel/line/height to longitude/latitude/height.
+pub struct Transformer {
+    c_transformer_ref: RefCell<*mut c_void>,
+}
+
+impl Transformer {
+    /// Construct a ```Transformer``` from a valid C GDAL pointer to a transformer instance.
```
metasim, line 58:
```suggestion
    /// Construct a [`Transformer`] from a valid C GDAL pointer to a transformer instance.
```

src/alg/transform.rs
```diff
@@ -0,0 +1,184 @@
...
+        Transformer {
+            c_transformer_ref: RefCell::new(c_transformer),
+        }
+    }
+
+    /// Extracts the Ground Control Points from the specified dataset.
+    ///
+    /// # Arguments
+    ///
+    /// * `dataset` - The `Dataset` to extract Ground Control Points from.
```
metasim, line 69:
```suggestion
    /// * `dataset` - The [`Dataset`] to extract Ground Control Points from.
```

src/raster/warp.rs
```diff
@@ -25,3 +29,36 @@ pub fn reproject(src: &Dataset, dst: &Dataset) -> Result<()> {
...
+
+pub fn create_and_reproject(
+    src: &Dataset,
+    dst_path: &Path,
+    dst_srs: &SpatialRef,
+    dst_driver: &Driver,
+) -> Result<()> {
+    let psz_dst_filename = dst_path
+        .to_str()
+        .expect("Destination path must be supplied.");
```
metasim, line 41: Won't this `panic`? Can you return an `Err` instead?

src/raster/warp.rs
```diff
@@ -25,3 +29,36 @@ pub fn reproject(src: &Dataset, dst: &Dataset) -> Result<()> {
...
+    let psz_dst_wkt = dst_srs.to_wkt().expect("Failed to obtain WKT for SRS.");
```
metasim, line 42: ditto?

src/alg/transform.rs
```diff
@@ -0,0 +1,184 @@
...
+                let psz_info = match &p.info {
+                    Some(arg) => {
+                        let tmp = CString::new(arg.clone()).unwrap();
+                        tmp.into_raw()
+                    }
+                    None => std::ptr::null_mut(),
+                };
+
+                gdal_sys::GDAL_GCP {
+                    pszId: CString::new(p.id.clone()).unwrap().into_raw(),
```
metasim, line 116: Is this use after free?
Instead of `unwrap`, should an `Err` be returned?

metasim (changes requested):

PR #186, opened 2021-05-05
Replace create_layer arguments list by LayerOptions struct

### LINKED ISSUE #185: Missing support for layer creation options
I need to pass `OVERWRITE=YES` in case the target layer already exists, so passing `null_mut()` as the last argument to `OGR_DS_CreateLayer` is counterproductive for me.

I propose to merge `create_layer` and `create_layer_blank` into a single function, `create_layer`, which takes a new `struct LayerOptions` that implements `Default`. This is consistent with how `Dataset::open_ex` works with `DatasetOptions`.

I'll create a PR soon, but if anyone already has some input, please give a shout.

ttencate:
This adds support for driver-specific `name=value` pairs while also
simplifying the API a bit.

Removes `create_layer_blank` because it seems to serve only niche use
cases and can easily be replaced by `create_layer(Default::default())`.

Fixes #185.

---

### DIFF
--- a/src/dataset.rs
+++ b/src/dataset.rs
@@ -121,6 +121,33 @@ pub struct DatasetOptions<'a> {
     pub sibling_files: Option<&'a [&'a str]>,
 }
 
+/// Parameters for [`Dataset::create_layer`].
+#[derive(Clone, Debug)]
+pub struct LayerOptions<'a> {
+    /// The name of the newly created layer. May be an empty string.
+    pub name: &'a str,
+    /// The SRS of the newly created layer, or `None` for no SRS.
+    pub srs: Option<&'a SpatialRef>,
+    /// The type of geometry for the new layer.
+    pub ty: OGRwkbGeometryType::Type,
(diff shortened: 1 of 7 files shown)


src/dataset.rs
```diff
@@ -935,6 +1013,21 @@ mod tests {
...
+        let (_temp_path, mut ds) = open_gpkg_for_update(fixture!("poly.gpkg"));
+        let mut options = LayerOptions {
+            name: "new",
+            ty: wkbPoint,
+            ..Default::default()
+        };
+        ds.create_layer(options.clone()).unwrap();
+        assert!(ds.create_layer(options.clone()).is_err());
+        options.options = Some(&["OVERWRITE=YES"]);
+        assert!(ds.create_layer(options.clone()).is_ok());
```
lnicola, line 1028: Nit: I think you could drop the `clone()` here, just in case `clippy` starts complaining at some point.

lnicola (approved):

src/dataset.rs
ttencate: Good find. Done.

PR #187, opened 2021-05-24
Fix small doc errors in rasterband

Pascal-So:
---

`read_band_as` and `write` had incorrect argument lists in the documentation.


lnicola:
Oof.

PR #192, opened 2021-06-10
Add `Feature::c_feature`

lnicola:
---

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add `Feature::c_feature` that returns the OGR feature handle.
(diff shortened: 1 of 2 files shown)


lnicola:
r? anyone?

jdroenner:
we could think about the pointer exposure in general. maybe such a method should be unsafe?

lnicola:
I wanted to do a follow-up PR for that. I don't think they need to be unsafe. In rust you, the place where you dereference a pointer is unsafe, not where you create it.

So if the feature is valid, there's no reason for this method to be unsafe. And if it's not valid, it was created by unsafe code.

lnicola:
Added `unsafe`, so it can be merged before deciding on all of the others.

Issue #190, opened 2021-06-10
Create options not supported


geohardtke:
Hi all,

I didn't find a way to set creation options for new raster datasets. In a local copy of the repo I implemented it like this:

```rust
pub struct RasterCreationOption<'a>{
    key : &'a str,
    value : &'a str,
}

// in the impl block for Dataset:
//
    pub fn create_with_band_type_with_options<T: GdalType>(
        &self,
        filename: &str,
        size_x: isize,
        size_y: isize,
        bands: isize,
        options: Vec<RasterCreationOption>,
    ) -> Result<Dataset> {

        // process options:
        let mut options_c = null_mut();
        for option in options{
            let psz_name  = CString::new(option.key)?;
            let psz_value = CString::new(option.value)?;
            unsafe{
                options_c = CSLSetNameValue(options_c, psz_name.as_ptr(), psz_value.as_ptr());
            }
        }

        let c_filename = CString::new(filename)?;
        let c_dataset = unsafe {
            gdal_sys::GDALCreate(
                self.c_driver,
                c_filename.as_ptr(),
                size_x as c_int,
                size_y as c_int,
                bands as c_int,
                T::gdal_type(),
                options_c as *mut *mut i8,
            )
        };
        if c_dataset.is_null() {
            return Err(_last_null_pointer_err("GDALCreate"));
        };
        Ok(unsafe { Dataset::from_c_dataset(c_dataset) })
    }
``` 

and I use it like:

```rust 
    let driver = Driver::get("GTiff").unwrap();
    let options = vec![RasterCreationOption{key:"COMPRESS", value:"LZW"} ,
                       RasterCreationOption{key:"TILED", value: "YES"}] ;
    let mut dataset = driver
          .create_with_band_type_with_options::<u8>("testing.tif",
             2048,
             2048,
             1,
             options)
          .unwrap();
}
```
And it seems to do what I want. I'm not sure if the implantation is safe/correct since I'm still new to rust. If you are interested I can create a PR for this (or improved with some guidance)?

lnicola:
Yeah, that looks reasonable for a start, so a PR would be great.

geohardtke:
Test in #193 are passing now. Didn't know that the dataset had to be `dropped` before the metadata could be read.

geohardtke:
closed. See #193.

PR #193, opened 2021-06-11
Add raster creation options

geohardtke:
---

Please review the additions I made (related to #190 ).

Also, I think there may be a bug in `dataset.metadata_item();` which makes the test ` test_create_with_band_type_with_options() ` fail, since I can see the COMPRESSION field in the image structure metadata when using `gdalinfo`. Or am I missing something?

gdalinfo output:
```bash
Files: /tmp/test.tif
Size is 256, 256
Image Structure Metadata:
  COMPRESSION=LZW
  INTERLEAVE=BAND
Corner Coordinates:
Upper Left  (    0.0,    0.0)
Lower Left  (    0.0,  256.0)
Upper Right (  256.0,    0.0)
Lower Right (  256.0,  256.0)
Center      (  128.0,  128.0)
Band 1 Block=128x64 Type=Byte, ColorInterp=Gray
```

test error:
```
thread 'raster::tests::test_create_with_band_type_with_options' panicked at 'assertion failed: `(left == right)`
  left: `None`,
 right: `Some("LZW")`', src/raster/tests.rs:273:5
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,18 +2,14 @@
 
 ## Unreleased
 
-- Add support to define creation options for raster datasets. To support this a new struct to store a creation option key and value pair  (`RasterCreationOption`) and and function (`driver.create_with_band_type_with_options()`) were created.
+- Add support for raster dataset creation options. A new struct (`RasterCreationOption`) and function (`driver.create_with_band_type_with_options()`) are now available for this.
 
 ```rust
     let driver = Driver::get("GTiff").unwrap();
-    let options = vec![RasterCreationOption{key:"COMPRESS", value:"LZW"} ,
-                       RasterCreationOption{key:"TILED", value: "YES"}] ;
+    let options = [RasterCreationOption{key:"COMPRESS", value:"LZW"} ,
+                   RasterCreationOption{key:"TILED", value: "YES"}] ;
     let mut dataset = driver
-          .create_with_band_type_with_options::<u8>("testing.tif",
-             2048,
-             2048,
-             1,
-             options)
+          .create_with_band_type_with_options::<u8>("testing.tif",2048,2048,1,&options)
           .unwrap();
 ```
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 - Add support for raster dataset creation options. A new struct (`RasterCreationOption`) and function (`driver.create_with_band_type_with_options()`) are now available for this.
 
+  - <https://github.com/georust/gdal/pull/193>
+
 ```rust
 let driver = Driver::get("GTiff").unwrap();
 let options = &[
@@ -19,6 +21,7 @@ let options = &[
 let mut dataset = driver
     .create_with_band_type_with_options::<u8>("testing.tif", 2048, 2048, 1, options)
     .unwrap();
+```
 
 - **Breaking**: Add support to select a resampling algorithm when reading a raster
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,21 @@
 
 ## Unreleased
 
+- Add support to define creation options for raster datasets. To support this a new struct to store a creation option key and value pair  (`RasterCreationOption`) and and function (`driver.create_with_band_type_with_options()`) were created.
+
+```rust
+    let driver = Driver::get("GTiff").unwrap();
+    let options = vec![RasterCreationOption{key:"COMPRESS", value:"LZW"} ,
+                       RasterCreationOption{key:"TILED", value: "YES"}] ;
+    let mut dataset = driver
+          .create_with_band_type_with_options::<u8>("testing.tif",
+             2048,
+             2048,
+             1,
+             options)
+          .unwrap();
+```
+
 - **Breaking**: Add support to select a resampling algorithm when reading a raster
 
   - <https://github.com/georust/gdal/pull/141>
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -90,22 +90,8 @@ impl Driver {
         size_y: isize,
         bands: isize,
     ) -> Result<Dataset> {
-        let c_filename = CString::new(filename)?;
-        let c_dataset = unsafe {
-            gdal_sys::GDALCreate(
-                self.c_driver,
-                c_filename.as_ptr(),
-                size_x as c_int,
-                size_y as c_int,
-                bands as c_int,
-                T::gdal_type(),
-                null_mut(),
-            )
-        };
-        if c_dataset.is_null() {
-            return Err(_last_null_pointer_err("GDALCreate"));
-        };
-        Ok(unsafe { Dataset::from_c_dataset(c_dataset) })
+        let options = [];
+        self.create_with_band_type_with_options::<T>(filename, size_x, size_y, bands, &options)
     }
 
     pub fn create_with_band_type_with_options<T: GdalType>(
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -34,10 +34,10 @@ extern "C" {
         pszName: *const libc::c_char,
         pszValue: *const libc::c_char) -> *mut *mut libc::c_char;
 }
-
+#[derive(Debug)]
 pub struct RasterCreationOption<'a>{
-    key : &'a str,
-    value : &'a str,
+    pub key : &'a str,
+    pub value : &'a str,
 }
 
 impl Driver {
@@ -76,6 +76,7 @@ impl Driver {
         let rv = unsafe { gdal_sys::GDALGetDriverLongName(self.c_driver) };
         _string(rv)
     }
+
 
     pub fn create(
         &self,
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -4,7 +4,7 @@ use crate::metadata::Metadata;
 use crate::raster::GdalType;
 use crate::utils::{_last_null_pointer_err, _string};
 use gdal_sys::{self, GDALDriverH, GDALMajorObjectH};
-use libc::c_int;
+use libc::{c_int, c_char};
 use std::ffi::CString;
 use std::ptr::null_mut;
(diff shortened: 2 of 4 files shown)


geohardtke:
It was not a bug in `dataset.metadata_item()`. The dataset needs to be explicitly dropped before being able to retrieve the metadata. Tests are passing now.

src/driver.rs
```diff
@@ -98,6 +108,42 @@ impl Driver {
         Ok(unsafe { Dataset::from_c_dataset(c_dataset) })
     }
 
+    pub fn create_with_band_type_with_options<T: GdalType>(
+        &self,
+        filename: &str,
+        size_x: isize,
+        size_y: isize,
+        bands: isize,
+        options: Vec<RasterCreationOption>,
```
lnicola, line 117: I think this can be `options: &[RasterCreationOption]`.

src/driver.rs
```diff
@@ -98,6 +108,42 @@ impl Driver {
...
+    ) -> Result<Dataset> {
+        // process options:
```
lnicola, line 119: That comment is probably not worth it.

src/driver.rs
```diff
@@ -98,6 +108,42 @@ impl Driver {
         Ok(unsafe { Dataset::from_c_dataset(c_dataset) })
```
lnicola, line 108: The old function could call the new one instead of having two similar implementations.

CHANGES.md
```diff
@@ -2,6 +2,21 @@
...
+    let options = vec![RasterCreationOption{key:"COMPRESS", value:"LZW"} ,
+                       RasterCreationOption{key:"TILED", value: "YES"}] ;
+    let mut dataset = driver
+          .create_with_band_type_with_options::<u8>("testing.tif",
+             2048,
+             2048,
+             1,
+             options)
+          .unwrap();
+```
```
lnicola, line 18:
```suggestion
```rust
let driver = Driver::get("GTiff").unwrap();
let options = &[
    RasterCreationOption {
        key: "COMPRESS",
        value: "LZW",
    },
    RasterCreationOption {
        key: "TILED",
        value: "YES",
    },
];
let mut dataset = driver
    .create_with_band_type_with_options::<u8>("testing.tif", 2048, 2048, 1, options)
    .unwrap();
```
```

File: CHANGES.md [hunk 5]
```diff
@@ -2,6 +2,21 @@
 
 ## Unreleased
 
+- Add support to define creation options for raster datasets. To support this a new struct to store a creation option key and value pair  (`RasterCreationOption`) and and function (`driver.create_with_band_type_with_options()`) were created.
```
```suggestion
- Add support for raster dataset creation options. A new struct (`RasterCreationOption`) and function (`driver.create_with_band_type_with_options()`) are now available for this.
```

src/raster/tests.rs
```diff
@@ -239,6 +240,53 @@ fn test_create_with_band_type() {
...
+    let tmp_filename = "/tmp/test.tif";
+    let dataset = driver
+        .create_with_band_type_with_options::<u8>(tmp_filename, 256, 256, 1, options)
+        .unwrap();
+
+    let rasterband = dataset.rasterband(1).unwrap();
+    let block_size = rasterband.block_size();
+    assert_eq!(block_size, (128, 64));
+
+    drop(dataset);
```
lnicola, line 278: This is fine, but you can also put it in an inner block:

```rust
{
    let dataset = driver
        .create_with_band_type_with_options::<u8>(tmp_filename, 256, 256, 1, options)
        .unwrap();

    let rasterband = dataset.rasterband(1).unwrap();
    let block_size = rasterband.block_size();
    assert_eq!(block_size, (128, 64));
}
```

src/raster/tests.rs
```diff
@@ -239,6 +240,53 @@ fn test_create_with_band_type() {
     assert_eq!(rb.band_type(), GDALDataType::GDT_Float32)
 }
 
+#[test]
+fn test_create_with_band_type_with_options() {
+    let driver = Driver::get("GTiff").unwrap();
+    let options = vec![
```
lnicola, line 246:
```suggestion
    let options = [
```

src/raster/tests.rs
```diff
@@ -239,6 +240,53 @@ fn test_create_with_band_type() {
...
+        },
+        RasterCreationOption {
+            key: "COMPRESS",
+            value: "LZW",
+        },
+        RasterCreationOption {
+            key: "INTERLEAVE",
+            value: "BAND",
+        },
+    ];
```
lnicola, line 267:
```suggestion
    ; 5];
```

src/raster/tests.rs
```diff
@@ -239,6 +240,53 @@ fn test_create_with_band_type() {
...
+
+    let tmp_filename = "/tmp/test.tif";
+    let dataset = driver
+        .create_with_band_type_with_options::<u8>(tmp_filename, 256, 256, 1, options)
```
lnicola, line 271:
```suggestion
        .create_with_band_type_with_options::<u8>(tmp_filename, 256, 256, 1, &options)
```

src/raster/tests.rs
```diff
@@ -239,6 +240,53 @@ fn test_create_with_band_type() {
...
+    let rasterband = dataset.rasterband(1).unwrap();
+    let block_size = rasterband.block_size();
+    assert_eq!(block_size, (128, 64));
+
+    drop(dataset);
+    let dataset = Dataset::open(Path::new(tmp_filename)).unwrap();
+    let key = "INTERLEAVE";
+    let domain = "IMAGE_STRUCTURE";
+    let meta = dataset.metadata_item(key, domain);
+    assert_eq!(meta, Some(String::from("BAND")));
```
lnicola, line 283: Fine, too, but you can also write it as `assert_eq!(meta.as_deref(), Some("BAND"));`

src/driver.rs
geohardtke: Yes, I agree. Looks much better!

src/driver.rs
geohardtke: agreed.

src/raster/tests.rs
geohardtke: Again, much better this way.

src/driver.rs
geohardtke: I agree.
I came up with this, which works, but I'm not sure passing empty key value pairs is a good idea. Is there a better way?

```rust
    pub fn create_with_band_type<T: GdalType>(
        &self,
        filename: &str,
        size_x: isize,
        size_y: isize,
        bands: isize,
    ) -> Result<Dataset> {
        let options = [RasterCreationOption { key: "", value: "" }];
        self.create_with_band_type_with_options::<T>(filename, size_x, size_y, bands, &options)
    }
```

geohardtke:
Thanks for reviewing my changes and your suggestions to improve them, I guess I still have a lot to learn!

Sorry I made a bit of a mess incorporating your suggestions! I didn't see the "commit suggestion" button at first (we use gitlab at work and the workflow is a bit different), I'll be more careful next time.

src/driver.rs
lnicola: Oh, so it didn't work with `[]`?

I agree that the API is clunky, but that's a larger change and I can't promise anything about it.

lnicola:
Yeah, don't worry about the commit attribution (I'd squash these anyway if it was my PR), and sorry for taking so long to review. Tests seem to fail now because of some extra bit of whitespace or something like that.

Unfortunately, I can't merge it (well, I can, but we use bors instead of merging from the GitHub UI), so someone else must take a look.

lnicola:
Just testing:

bors d+

geohardtke:
yes, it seems to be an extra space in the test. I'll fix it now.

lnicola:
PS: you can add your (presumably) other email address to your GitHub account so that you don't show as two different people.

src/driver.rs
geohardtke: oh, yes. It does, sorry! :)

lnicola:
Can you add link to the PR in the changelog?

lnicola:
So if you want to look into a builder API in the meanwhile, it would be awesome :-).

geohardtke:
Sure, I can open another issue and we discuss there.

src/driver.rs
```diff
@@ -89,7 +120,7 @@ impl Driver {
                 size_y as c_int,
                 bands as c_int,
                 T::gdal_type(),
-                null_mut(),
+                options_c as *mut *mut i8,
```
jdroenner, line 123: shouldn't this be a *mut * mut char?

jdroenner (commented):
i like the feature. will test it this weekend :)

src/driver.rs
lnicola: You mean `c_char`?

geohardtke: yes, `options_c as *mut *mut i8` is correct. I'm not sure how it ended up being `null_mut()` again, sorry.

lnicola: It's not `null_mut()`, it's `options_c as *mut *mut i8`, jdroenner is asking if it should be `options_c as *mut *mut c_char`.

geohardtke: oh, sorry! my bad.  Blame Friday night.

geohardtke: gdal_sys::GDALCreate expects *mut *mut i8.

lnicola: Yeah, it's defined as `char**` in the GDAL headers. We might be able to improve the bindings, but I think we shouldn't block merging this.

geohardtke: Any thoughts @jdroenner?  Does the bindings need to be updated before merging this? I think I don't have enough knowledge to do that.

jdroenner: no its fine. i just assumed it to be a char in the bindings.

geohardtke: Cool! Thanks!

PR #200, opened 2021-07-28
OGR Layer: Set Attribute Filter & Set Spatial Filter as Rectangle

ChristianBeilschmidt:
---

I've added two new functions to the OGR layer wrapper that call…
 * OGR_L_SetAttributeFilter
 * OGR_L_SetSpatialFilterRect

I've also added tests so that we can see that they work.

### DIFF
--- a/src/vector/layer.rs
+++ b/src/vector/layer.rs
@@ -334,15 +334,11 @@ impl<'a> Layer<'a> {
     /// Set a new attribute query that restricts features when using the feature iterator.
     ///
     /// Parameters:
-    /// - `psz_query` `Some(query)` in restricted SQL WHERE format, or `None` to clear the current query.
+    /// - `psz_query` in restricted SQL WHERE format
     ///
-    pub fn set_attribute_filter(&self, psz_query: Option<&str>) -> Result<()> {
-        let rv = if let Some(query) = psz_query {
-            let c_str = CString::new(query)?;
-            unsafe { gdal_sys::OGR_L_SetAttributeFilter(self.c_layer, c_str.as_ptr()) }
-        } else {
-            unsafe { gdal_sys::OGR_L_SetAttributeFilter(self.c_layer, std::ptr::null()) }
-        };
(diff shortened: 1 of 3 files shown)


src/vector/layer.rs
```diff
@@ -330,6 +335,32 @@ impl<'a> Layer<'a> {
             gdal_sys::OGR_L_ResetReading(self.c_layer);
         }
     }
+
+    /// Set a new attribute query that restricts features when using the feature iterator.
+    ///
+    /// Parameters:
+    /// - `psz_query` in restricted SQL WHERE format
+    ///
+    pub fn set_attribute_filter(&self, psz_query: &str) -> Result<()> {
```
lnicola, line 344: I _think_ the other methods use Rust parameter naming conventions, so maybe we can call this `query` instead.

lnicola (approved):

src/vector/layer.rs
ChristianBeilschmidt: 05c99b5

lnicola:
r? @jdroenner 

(I have merge, but not bors rights :smile:)

jdroenner:
@lnicola i never use bors. how do i grant bors rights?

jdroenner:
XD

lnicola:
Oh, that clears it up :smile:.

jdroenner:
i think its an issue caused by github teams

lnicola:
I think bors is configured separately from GitHub -- you can grant rights to someone without adding them to any GitHub team.

jdroenner:
yes but there is no reviewer and i think an owner of the repo hast to grand initial rights. However, i'm only in the gdal team not an owner of the repo.

lnicola:
@notriddle hey, I was wondering if you happen to know when bors was set up on https://app.bors.tech/repositories/34, given its low id. Around 2017, maybe?

I think we're in a bit of a pickle regarding the ownership of the integration.

PR #202, opened 2021-07-28
add bors

jdroenner:
---

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -1,0 +1,3 @@
+tatus = [
+    "ci result",
+]
--- a/bors.toml
+++ b/bors.toml
@@ -1,3 +1,3 @@
-tatus = [
+status = [
     "ci result",
 ]


bors.toml
```diff
@@ -0,0 +1,3 @@
+status = [
+    "ci result",
```
michaelkirk, line 2: This string needs to correspond to the job name(s) from GH actions that must succeed in order for bors to consider CI to have been successful. 

georust/geo uses "ci result", which corresponds to https://github.com/georust/geo/commit/c169918826af4113208bba10eed676ce4f13011e#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R11

I don't expect this to work with your current setup unless you also alter your .workflow file similarly.

My understanding is that there's some other ownership snafu keeping your from using bors at the moment anyway, but just a heads up about the *next* hurdle. =)

urschrei: Oh, dammit, this is my fault. I forgot that the CI result is customisable. Sorry!

Issue #204, opened 2021-08-03
add valgrind on tests in ci and fix all the leaks


jdroenner:
run valgrind with:

CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="valgrind --num-callers=100 --track-origins=yes --leak-check=full --time-stamp=yes" cargo test

current report is attached
[rustgdal_valgrind.txt](https://github.com/georust/gdal/files/6924869/rustgdal_valgrind.txt)

ChristianBeilschmidt:
In my opinion, it is good to run Valgrind since we deal with lots of unsafe code here. This would also prevent you from doing these kinds of clean-ups every now and then.

ChristianBeilschmidt:
I saw that ASAN was already discussed here: https://github.com/georust/gdal/issues/165

I guess it is nicer than valgrind for our use case, but comes with the nightly issue. From my experience, we would have to fix a nightly because it will definitely crash for some nightlies. In our Geo Engine project, we have a CI task that updates the nightly version from time to time :shrug:.

Why is `trybuild` necessary, since it isn't part of https://github.com/georust/gdal/pull/167/files as far as I see?

metasim:
Not sure if it's as comprehensive as valgrind, but this has been added, which includes ASAN:

https://github.com/georust/gdal/pull/398?

PR #205, opened 2021-08-04
Fix memory leaks reported by Valgrind & re-generate pre-build bindings

jdroenner:
---

This PR fixes memory leaks reported by Valgrind [#204].

Run Valgrind like this:
`CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="valgrind --num-callers=100 --track-origins=yes --leak-check=full --time-stamp=yes" cargo test --all-features`

However, this still reports a lot of "leaks":
`CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="valgrind --num-callers=100 --track-origins=yes --leak-check=full --show-leak-kinds=all  --time-stamp=yes" cargo test --all-features`
They appear to be caused by the driver registration call and may not be a real problem.


jdroenner:
[Fix GDAL memory leaks](https://trello.com/c/kLoubFhu/323-fix-gdal-memory-leaks)

src/driver.rs
```diff
@@ -123,6 +123,8 @@ impl Driver {
                 options_c as *mut *mut i8,
             )
         };
+        unsafe { gdal_sys::CSLDestroy(options_c) };
```
ChristianBeilschmidt, line 126: On line 110, there are loops over `options_c`. Is this pointer just re-used in 123? Otherwise, we would only destroy the last return of `CSLSetNameValue` if we would have more than one option.
And if we have no option, we would destroy `null_mut`?

src/metadata.rs
```diff
@@ -27,6 +27,7 @@ pub trait Metadata: MajorObject {
                 domains.push(_string(p));
             }
         }
+        unsafe { gdal_sys::CSLDestroy(c_res) };
```
ChristianBeilschmidt, line 30: Shouldn't this be inside the not-null block or is it okay to call this function if `c_res` is null?

src/spatial_ref/srs.rs
```diff
@@ -203,14 +203,16 @@ impl SpatialRef {
...
+        let res = if rv != OGRErr::OGRERR_NONE {
             Err(GdalError::OgrError {
                 err: rv,
                 method_name: "OSRExportToWkt",
             })
         } else {
             Ok(_string(c_wkt))
-        }
+        };
+        unsafe { gdal_sys::VSIFree(c_wkt.cast::<std::ffi::c_void>()) };
```
ChristianBeilschmidt, line 214: For all `c_wkt` in this file. What is the pointer if we get an error? Do we need to destroy it or is this a null pointer?

ChristianBeilschmidt (changes requested):
It's awesome to have leaks fixed.

I've asked some questions on some critical areas, though. We should double-check that the behavior is correct.

Moreover, we should sit together and advance our CI to capture this stuff.

src/driver.rs
jdroenner: Frees the passed string list (null terminated array of strings). It is safe to pass NULL.
https://gdal.org/doxygen/ogr__api_8h.html#a4000d426bf26ad7cc7d4012634c93f09

src/metadata.rs
jdroenner: Frees the passed string list (null terminated array of strings). It is safe to pass NULL.
https://gdal.org/doxygen/ogr__api_8h.html#a4000d426bf26ad7cc7d4012634c93f09

src/spatial_ref/srs.rs
jdroenner: with the changes we always call free on the pointer

src/driver.rs
jdroenner: it is also used in gdal this way: https://github.com/OSGeo/gdal/blob/d0a12d1a74e08557605f57cf2d94dac82b9fdda9/gdal/apps/gdalwarp_lib.cpp#L664

ChristianBeilschmidt: Seems okay to me, I somewhat can't resolve this comment.

src/metadata.rs
ChristianBeilschmidt: The correct link is https://gdal.org/doxygen/cpl__string_8h.html#a5a39b9c5896a273cc6f06c1d5be93238.
Okay, seems fine with me. Again, I cannot resolve this comment.

src/spatial_ref/srs.rs
ChristianBeilschmidt: Yes, but I can't find the info on VSIFree that it is okay to insert a null pointer.
If this is the case, we can resolve this.

jdroenner: it is ok to call free on nullotr  in c: https://linux.die.net/man/3/free

and i think it is ok to call VSIfree on nullptr:
https://github.com/OSGeo/gdal/blob/fec15b146f8a750c23c5e765cac12ed5fc9c2b85/gdal/port/cpl_vsisimple.cpp#L827

ChristianBeilschmidt: Ok for me, still I can't resolve this conversation.

ChristianBeilschmidt (approved):

lnicola:
Never mind, we have to merge this manually.

michaelkirk:
bors r- 

Manually stopping this run.

Because the bors.toml in the branch has the wrong value for `status`, I think bors will wait indefinitely (maybe it'd eventually time out).

PR #208, opened 2021-08-05
Fix bors integration

michaelkirk:
---

I think the elements in `status` need to match all the GH Action job names we require to succeed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, trying ]
   pull_request:
-    branches: [ master ]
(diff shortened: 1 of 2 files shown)


michaelkirk:
bors try

lnicola:
Is it stuck?

michaelkirk:
bors try-

michaelkirk:
bors try

michaelkirk:
> Build failed:

woohoo! I think that's actually what we want to see at this point, because it means bors is working. The clippy lint causing the build failure is fixed in #205.

So you can merge #205 first, or manually merge this while the build is still broken, or whatever else you like.

michaelkirk:
bors retry

😒

michaelkirk:
bors retry

😒 🤖

PR #211, opened 2021-08-06
Update semver

jdroenner:
- [ x] I agree to follow the project's [code of conduct](https://github.com/georust/gdal/blob/master/CODE_OF_CONDUCT.md).
- [ x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---


jdroenner:
will release a new version when this is merged :)

PR #212, opened 2021-09-01
Add set description

msalib:
---
This is a really simple change that allows users to set descriptions on major GDAL objects; that's mostly useful for RasterBands. If you prefer I could just make this a method on `RasterBand` only instead of more generally making it part of the `Metadata` trait.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -196,6 +196,9 @@ let mut dataset = driver
 
   - <https://github.com/georust/gdal/pull/203>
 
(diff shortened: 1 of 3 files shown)


lnicola:
Thanks! And don't sigh.

bors r+

lnicola:
bors r-

Well, while we're at sighing, can you add a link to the PR in the changelog?

bors d+

msalib:
Sure thing!

bors r+

PR #214, opened 2021-09-03
Update types to fix build on ppc64le

mbrobbel:
---

`libc::c_char` is a `u8` on `ppc64le`. Using the type definition from `libc` fixes it (maybe for other archs too).


lnicola:
Thanks! Can you also add the changelog entry?

PR #215, opened 2021-09-07
Creating error callbacks

ChristianBeilschmidt:
---

I've attached two options for setting up the global GDAL error callback. One is thread-safe and comes with a new dependecy and the other one is not.

The reason for storing the callback in a static variable is that we cannot pass the ownership to GDAL. We can access a generic pointer via the `pUserData` field and cast it to our function type. Therefore, in order not to leak our `FnMut`, we have to store it somehow and drop it later on.

When we decide on which version to use, I can remove the other one and convert the draft to a PR.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -126,9 +126,11 @@ static ERROR_CALLBACK: Lazy<Mutex<Option<PinnedErrorCallback>>> = Lazy::new(Defa
 // Stores the callback in the static variable [`ERROR_CALLBACK`].
 // Internally, it passes a pointer to the callback to GDAL as `pUserData`.
 //
+/// The function must be `Send` and `Sync` since it is potentially called from multiple threads.
+///
 pub fn set_error_handler<F>(callback: F)
 where
-    F: FnMut(CplErrType, i32, &str) + 'static + Send,
+    F: FnMut(CplErrType, i32, &str) + 'static + Send + Sync,
 {
     unsafe extern "C" fn error_handler(
         error_type: CPLErr::Type,
@@ -152,7 +154,7 @@ where
 
     let mut callback_lock = match ERROR_CALLBACK.lock() {
         Ok(guard) => guard,
-        // poor man's lock poisoning handling, i.e., ignoring it
+        // poisoning could only occur on `CPLSetErrorHandler(Ex)` panicing, thus the value must be valid nevertheless
         Err(poison_error) => poison_error.into_inner(),
     };
 
@@ -169,7 +171,7 @@ where
 pub fn remove_error_handler() {
     let mut callback_lock = match ERROR_CALLBACK.lock() {
         Ok(guard) => guard,
-        // poor man's lock poisoning handling, i.e., ignoring it
+        // poisoning could only occur on `CPLSetErrorHandler(Ex)` panicing, thus the value must be valid nevertheless
         Err(poison_error) => poison_error.into_inner(),
     };
 
@@ -235,7 +237,7 @@ mod tests {
             set_error_handler(move |_a, _b, _c| {});
         });
 
-        // A thread that makes a lot of mistakes
+        // A thread that provokes potential race conditions
         let join_handle = thread::spawn(move || {
             for _ in 0..100 {
                 unsafe {
--- a/src/config.rs
+++ b/src/config.rs
@@ -28,7 +28,7 @@ use libc::{c_char, c_void};
 
 use crate::errors::Result;
 use crate::utils::_string;
-use lazy_static::lazy_static;
+use once_cell::sync::Lazy;
 use std::ffi::CString;
 use std::sync::Mutex;
 
@@ -127,10 +127,9 @@ type CallbackType = dyn FnMut(CplErr, i32, &str) + 'static;
 static mut ERROR_CALLBACK: Option<Box<CallbackType>> = None;
 
 type CallbackTypeThreadSafe = dyn FnMut(CplErr, i32, &str) + 'static + Send;
-lazy_static! {
-    static ref ERROR_CALLBACK_THREAD_SAFE: Mutex<Option<Box<CallbackTypeThreadSafe>>> =
-        Mutex::new(None);
-}
+
+static ERROR_CALLBACK_THREAD_SAFE: Lazy<Mutex<Option<Box<CallbackTypeThreadSafe>>>> =
+    Lazy::new(Default::default);
 
 /// Set a custom error handler for GDAL.
 /// Could be overwritten by setting a thread-local error handler.
--- a/src/config.rs
+++ b/src/config.rs
@@ -114,19 +114,19 @@ pub fn clear_thread_local_config_option(key: &str) -> Result<()> {
     Ok(())
 }
 
-type CallbackType = dyn FnMut(CplErrType, i32, &str) + 'static + Send;
-type PinnedCallback = Pin<Box<Box<CallbackType>>>;
+type ErrorCallbackType = dyn FnMut(CplErrType, i32, &str) + 'static + Send;
+type PinnedErrorCallback = Pin<Box<Box<ErrorCallbackType>>>;
 
 /// Static variable that holds the current error callback function
-static ERROR_CALLBACK: Lazy<Mutex<Option<PinnedCallback>>> = Lazy::new(Default::default);
+static ERROR_CALLBACK: Lazy<Mutex<Option<PinnedErrorCallback>>> = Lazy::new(Default::default);
 
 /// Set a custom error handler for GDAL.
 /// Could be overwritten by setting a thread-local error handler.
 ///
-/// Note:
-/// Stores the callback in the static variable [`ERROR_CALLBACK_THREAD_SAFE`].
-/// Internally, it passes a pointer to the callback to GDAL as `pUserData`.
-///
+// Note:
+// Stores the callback in the static variable [`ERROR_CALLBACK`].
+// Internally, it passes a pointer to the callback to GDAL as `pUserData`.
+//
 pub fn set_error_handler<F>(callback: F)
 where
     F: FnMut(CplErrType, i32, &str) + 'static + Send,
@@ -141,15 +141,15 @@ where
 
         // reconstruct callback from user data pointer
         let callback_raw = CPLGetErrorHandlerUserData();
-        let callback: &mut Box<CallbackType> = &mut *(callback_raw as *mut Box<_>);
+        let callback: &mut Box<ErrorCallbackType> = &mut *(callback_raw as *mut Box<_>);
 
         callback(error_type, error_num as i32, &error_msg);
     }
 
     // pin memory location of callback for sending its pointer to GDAL
-    let mut callback: PinnedCallback = Box::pin(Box::new(callback));
+    let mut callback: PinnedErrorCallback = Box::pin(Box::new(callback));
 
-    let callback_ref: &mut Box<CallbackType> = callback.as_mut().get_mut();
+    let callback_ref: &mut Box<ErrorCallbackType> = callback.as_mut().get_mut();
 
     let mut callback_lock = match ERROR_CALLBACK.lock() {
         Ok(guard) => guard,
--- a/src/config.rs
+++ b/src/config.rs
@@ -23,9 +23,14 @@
 //! Refer to [GDAL `ConfigOptions`](https://trac.osgeo.org/gdal/wiki/ConfigOptions) for
 //! a full list of options.
 
+use gdal_sys::{CPLErr, CPLErrorNum, CPLGetErrorHandlerUserData};
+use libc::{c_char, c_void};
+
 use crate::errors::Result;
 use crate::utils::_string;
+use lazy_static::lazy_static;
 use std::ffi::CString;
+use std::sync::Mutex;
 
 /// Set a GDAL library configuration option
 ///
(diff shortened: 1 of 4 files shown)


lnicola:
Can you use `once_cell` instead? It might get folded into `std` at some point.

ChristianBeilschmidt:
> Can you use [`once_cell`](https://docs.rs/once_cell/1.8.0/once_cell/#lazy-initialized-global-data) instead? It might get folded into `std` at some point.

Done that in da4c2ac

ChristianBeilschmidt:
Okay, I've done two things:

1. 0111fbd
I removed the non-thread-safe variant.
Moreover, I noticed that the previous version had a potential race condition when we would switch the callback in the mutex and then install the new callback. Thus, we have to first install the new callback to not have the old callback point to a dangling pointer. However, we have to first call it and then move the callback into the mutex without changing the reference that GDAL uses in p_user_data. This is why I wrapped the Box<CallbackFn> into a Pin<Box<_>>. If you have any improvement on that, I will try it.
2. 6387a29
I moved the error type enum to error.rs

src/config.rs
```diff
@@ -108,10 +114,111 @@ pub fn clear_thread_local_config_option(key: &str) -> Result<()> {
...
+
+    // pin memory location of callback for sending its pointer to GDAL
+    let mut callback: PinnedCallback = Box::pin(Box::new(callback));
+
+    let callback_ref: &mut Box<CallbackType> = callback.as_mut().get_mut();
+    unsafe {
+        gdal_sys::CPLSetErrorHandlerEx(Some(error_handler), callback_ref as *mut _ as *mut c_void);
+    };
+
+    let mut callback_lock = match ERROR_CALLBACK.lock() {
```
rmanoka, line 157: I think there may be a race condition here.  Suppose two threads A, B race to set a error handler.  Then, it is possible A sets the handler, then B sets, thus replacing it (from GDAL's perspective).  However, during the subsequent Mutex lock, the order could get reversed, and the mutex ends up storing the wrong handler; worse yet, the correct handler has now been dropped, and could cause a crash when accessed.

rmanoka: I was able to modify the test case you had, to show how it could trigger:

```rust
#[test]
fn error_handler() {
    use std::thread;
    // Two racing threads trying to set error handlers
    // First one
    thread::spawn(move || {
        loop {
            set_error_handler(move |a, b, c| {
            });
        }
    });

    // Second one
    thread::spawn(move || {
        loop {
            set_error_handler(move |a, b, c| {
            });
        }
    });


    // A thread that makes a lot of mistakes
    thread::spawn(move || {
        loop {
            unsafe {
                let msg = CString::new("foo".as_bytes()).unwrap();
                gdal_sys::CPLError(CPLErr::CE_Failure, 42, msg.as_ptr());
            };

            unsafe {
                let msg = CString::new("bar".as_bytes()).unwrap();
                gdal_sys::CPLError(std::mem::transmute(CplErrType::Warning), 1, msg.as_ptr());
            };
        }
    });

    // Wait for the segfault to happen.  Happens within a few seconds typically.
    loop {}
}
```

ChristianBeilschmidt: provided a fix and included the test (kind of modified due to the infinite loops)

bce691d

src/config.rs
```diff
@@ -108,10 +114,111 @@ pub fn clear_thread_local_config_option(key: &str) -> Result<()> {
...
+type PinnedCallback = Pin<Box<Box<CallbackType>>>;
+
+/// Static variable that holds the current error callback function
+static ERROR_CALLBACK: Lazy<Mutex<Option<PinnedCallback>>> = Lazy::new(Default::default);
+
+/// Set a custom error handler for GDAL.
+/// Could be overwritten by setting a thread-local error handler.
+///
+/// Note:
+/// Stores the callback in the static variable [`ERROR_CALLBACK_THREAD_SAFE`].
```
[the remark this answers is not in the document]
ChristianBeilschmidt: 40e7fc8

src/config.rs
```diff
@@ -108,10 +114,111 @@ pub fn clear_thread_local_config_option(key: &str) -> Result<()> {
     Ok(())
 }
 
+type CallbackType = dyn FnMut(CplErrType, i32, &str) + 'static + Send;
+type PinnedCallback = Pin<Box<Box<CallbackType>>>;
```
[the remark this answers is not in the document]
ChristianBeilschmidt: You have to use the double-box in order to get a stable pointer and move something that boxes it around.
I could remove the `Pin` as it looks. The idea was to prevent it from moving around.

rmanoka:
@ChristianBeilschmidt Thanks for the implementation.  Since concurrency is a bit hard to debug / reason about, I wanted to discuss a few points on the overall impl.

1. Callback Storage. Currently you're storing the callback both in the static variable, and in GDAL's `ErrorHandlerUserData`. We should have a single source of truth; I think just using the static variable is simpler.

1. Function Type.  I was a bit worried about the `FnMut` possibly being called in parallel in `error_handler`.  However, it seems GDAL [uses a mutex](https://github.com/OSGeo/gdal/blob/543d6ffd39b66a727da49cb398591959a8f644cf/gdal/port/cpl_error.cpp#L355) to ensure the error handler is not called / assigned in parallel.  Could you add some comments or GDAL doc links  explaining this.  The function needs `Sync` in addition to `Send + 'static` as it is being accessed from different threads.

1. Poison Handling. I think the only panic while holding the mutex in your implementation is if `CPLSetErrorHandler` panics.  Thus, the function pointer is valid even if poisoned.  Pls. explain this in the code for future reference.

1. Locking.  The mutex code seems good to me.  I still think just a `Box<ErrorCallbackType>` suffices, as the locking mechanism ensures the pointer doesn't change while being accessed.  

   If you are feeling brave, there is a way to fully avoid using any locks at all.  The idea is to use an `AtomicPtr` as the static variable, initialized to null (`Box` pointers are never null).

   Before setting the handler, use `AtomicPtr::swap`:

   ``` rust
   // Swap should be done before CPLSetHandler
   let old_ptr = ERROR_CALLBACK.swap(Box::into_raw(callback), Ordering::SeqCst);
   CPLSetHandler(...)
   // Drop logic must happen after CPLSetHandler
   if old_addr > 0 {
       // Convert back to Box to destruct properly
       unsafe { Box::from_raw(old_addr as *mut ErrorCallbackType) };
   }
   ```

   Removal is simple:

   ``` rust
   // Set None before swapping
   CPLSetHandler(None);
   // DO NOT swap out and drop the old addr
   ```

   This is somewhat like the [Treiber stack](https://en.wikipedia.org/wiki/Treiber_stack), but much simpler as we just store one value and piggy-back on GDAL's mutex.  See this neat [rust exposition](https://aturon.github.io/blog/2015/08/27/epoch/)  that explains why we can't free stuff easily in lock-free algorithms.  In our case, we may leak at most one `Box` which seems reasonable.

   Calling is simple.   If our `fn error_handler` is called, then we know a handler we set is currently active.  So we can use `AtomicPtr::load` to read the ptr, and call the function.  We could even `assert` the pointer is not null.  We can also be sure the pointed memory is not freed as the `CPLSetHandler` call won't finish before the error handler return.

ChristianBeilschmidt:
> @ChristianBeilschmidt Thanks for the implementation. Since concurrency is a bit hard to debug / reason about, I wanted to discuss a few points on the overall impl.
> 
>     1. Callback Storage. Currently you're storing the callback both in the static variable, and in GDAL's `ErrorHandlerUserData`. We should have a single source of truth; I think just using the static variable is simpler.

I guess we can't. GDAL doesn't take the ownership of that pointer, it treats it as a reference. Thus, we would leak if we transfer the ownership to GDAL nevertheless and it gets discarded by a call to `CPLSetErrorHandler` (or whatever) that is not covered by our two functions.

> 
>     2. Function Type.  I was a bit worried about the `FnMut` possibly being called in parallel in `error_handler`.  However, it seems GDAL [uses a mutex](https://github.com/OSGeo/gdal/blob/543d6ffd39b66a727da49cb398591959a8f644cf/gdal/port/cpl_error.cpp#L355) to ensure the error handler is not called / assigned in parallel.  Could you add some comments or GDAL doc links  explaining this.  The function needs `Sync` in addition to `Send + 'static` as it is being accessed from different threads.

I changed it here: bd4beaa 

> 
>     3. Poison Handling. I think the only panic while holding the mutex in your implementation is if `CPLSetErrorHandler` panics.  Thus, the function pointer is valid even if poisoned.  Pls. explain this in the code for future reference.

I've addressed it here: bd4beaa 

> 
>     4. Locking.  The mutex code seems good to me.  I still think just a `Box<ErrorCallbackType>` suffices, as the locking mechanism ensures the pointer doesn't change while being accessed.

 if you only use one Box and you move it into the Mutex, the address changes. The inner Box makes the address stable.
And we need two stable addresses while swapping our new Box into the mutex while simultaneously ensuring that all error callbacks that happen in-between (from actual errors, not our function calls) have a valid pointer.

>        If you are feeling brave, there is a way to fully avoid using any locks at all.  The idea is to use an `AtomicPtr` as the static variable, initialized to null (`Box` pointers are never null).
>        Before setting the handler, use `AtomicPtr::swap`:
>        ```rust
>        // Swap should be done before CPLSetHandler
>        let old_ptr = ERROR_CALLBACK.swap(Box::into_raw(callback), Ordering::SeqCst);
>        CPLSetHandler(...)
>        // Drop logic must happen after CPLSetHandler
>        if old_addr > 0 {
>            // Convert back to Box to destruct properly
>            unsafe { Box::from_raw(old_addr as *mut ErrorCallbackType) };
>        }
>        ```
>        
>            
>              
>            
>        
>              
>            
>        
>            
>          
>        Removal is similar, swap in `null_mut`
>        ```rust
>        let old_ptr = ERROR_CALLBACK.swap(null_mut(), Ordering::SeqCst);
>        CPLSetHandler(None);
>        // Drop old_ptr
>        ```
>        
>        
>            
>              
>            
>        
>              
>            
>        
>            
>          
>        This is somewhat like the [Treiber stack](https://en.wikipedia.org/wiki/Treiber_stack), but much simpler as we just store one value and piggy-back on GDAL's mutex.  See also this neat [rust exposition](https://aturon.github.io/blog/2015/08/27/epoch/).  Unlike the treiber stack, we can in fact safely drop the memory as we piggy-back on GDAL's mutex mechanism.
>        Calling is simple.   If our `fn error_handler` is called, then we know a handler we set is currently active.  So we can use `AtomicPtr::load` to read the ptr, and call the function.  If the pointer is not null, we can be sure the pointed memory is not freed as the `CPLSetHandler` call won't finish before the error handler return.  Only edge case is if the handler is called while we're removing the handler, in which case the ptr may be null.  We just need to exit without calling in this case.
> 
> 
> Edit: fix lock-free logic to not require leaking the box ptr.

I could try it and see if we don't introduce new race conditions. However, the function is most likely not called in huge concurrency. So I don't see the benefit in locking, here.

jdroenner:
can we merge it like it is or do we need more discussion / changes?

rmanoka:
@ChristianBeilschmidt 

> > 1. Callback Storage. Currently you're storing the callback both in the static variable, and in GDAL's `ErrorHandlerUserData`. We should have a single source of truth; I think just using the static variable is simpler.

> I guess we can't. GDAL doesn't take the ownership of that pointer, it treats it as a reference. Thus, we would leak if we transfer the ownership to GDAL nevertheless and it gets discarded by a call to `CPLSetErrorHandler` (or whatever) that is not covered by our two functions.

I'm suggesting not using user-data of GDAL at all, and only using our locked structure.  So the only pointer being passed to GDAL is the `error_handler` function which is a static function.   It's mostly for code clarity, so optional.


> > 4. Locking.  The mutex code seems good to me.  I still think just a `Box<ErrorCallbackType>` suffices, as the locking mechanism ensures the pointer doesn't change while being accessed.

> if you only use one Box and you move it into the Mutex, the address changes. The inner Box makes the address stable.
> And we need two stable addresses while swapping our new Box into the mutex while simultaneously ensuring that all error callbacks that happen in-between (from actual errors, not our function calls) have a valid pointer.

If you move a `Box` into a `Mutex`, the data pointed to by the `Box` doesn't move.  For instance:

```rust
let mut boxed = Box::new(0usize);
let ptr: *mut _ = boxed.as_mut() as *mut _;

let mutex = Mutex::new(boxed);
assert_eq!(
    mutex.lock().unwrap().as_mut() as *mut _,
    ptr
);
```

> >    If you are feeling brave, there is a way to fully avoid using any locks at all.  The idea is to use an `AtomicPtr` as the static variable, initialized to null (`Box` pointers are never null).
> I could try it and see if we don't introduce new race conditions. However, the function is most likely not called in huge concurrency. So I don't see the benefit in not locking, here.

Ok.

@jdroenner A few minor suggestions, mostly for clarity and remove extra allocations.

ChristianBeilschmidt:
> @ChristianBeilschmidt
> 
> > > 1. Callback Storage. Currently you're storing the callback both in the static variable, and in GDAL's `ErrorHandlerUserData`. We should have a single source of truth; I think just using the static variable is simpler.
> 
> > I guess we can't. GDAL doesn't take the ownership of that pointer, it treats it as a reference. Thus, we would leak if we transfer the ownership to GDAL nevertheless and it gets discarded by a call to `CPLSetErrorHandler` (or whatever) that is not covered by our two functions.
> 
> I'm suggesting not using user-data of GDAL at all, and only using our locked structure. So the only pointer being passed to GDAL is the `error_handler` function which is a static function. It's mostly for code clarity, so optional.

I hesitate to switch to `AtomicPtr` since we then guarantee to *not* `drop` the `Box<Fn>` at the end of our program. But I haven't used it before. If you have an idea for that, I could try it.

> 
> > > 1. Locking.  The mutex code seems good to me.  I still think just a `Box<ErrorCallbackType>` suffices, as the locking mechanism ensures the pointer doesn't change while being accessed.
> 
> > if you only use one Box and you move it into the Mutex, the address changes. The inner Box makes the address stable.
> > And we need two stable addresses while swapping our new Box into the mutex while simultaneously ensuring that all error callbacks that happen in-between (from actual errors, not our function calls) have a valid pointer.
> 
> If you move a `Box` into a `Mutex`, the data pointed to by the `Box` doesn't move. For instance:
> 
> ```rust
> let mut boxed = Box::new(0usize);
> let ptr: *mut _ = boxed.as_mut() as *mut _;
> 
> let mutex = Mutex::new(boxed);
> assert_eq!(
>     mutex.lock().unwrap().as_mut() as *mut _,
>     ptr
> );
> ```
> 
> > > If you are feeling brave, there is a way to fully avoid using any locks at all.  The idea is to use an `AtomicPtr` as the static variable, initialized to null (`Box` pointers are never null).
> > > I could try it and see if we don't introduce new race conditions. However, the function is most likely not called in huge concurrency. So I don't see the benefit in not locking, here.
> 
> Ok.
> 

In your example, you have a pointer to a `usize`. However, we have to store a pointer to a `Box<dyn Trait>`, since we can't have a pointer to a `dyn Trait`. Thus, we have to wrap this into another `Box<_>` for which _then_ your examined behavior holds.

rmanoka:
> I hesitate to switch to `AtomicPtr` since we then guarantee to _not_ `drop` the `Box<Fn>` at the end of our program. But I haven't used it before. If you have an idea for that, I could try it.

Rust actually never drops any static variable, so even now things won't get dropped when the program exits unless `remove_handler` is called explicitly.  The lock-free stuff is probably unnecessary complexity, so let's stick with the current impl.

> In your example, you have a pointer to a `usize`. However, we have to store a pointer to a `Box<dyn Trait>`, since we can't have a pointer to a `dyn Trait`. Thus, we have to wrap this into another `Box<_>` for which _then_ your examined behavior holds.

See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=534a07f5e38378c5cd2d6a18552fa2a5) for example of boxing a `FnMut` into a `Mutex` and calling it via a raw ptr.  Basically the same use-case as here.

Also, pls. add a line in CHANGES.md about this feature.

ChristianBeilschmidt:
> > I hesitate to switch to `AtomicPtr` since we then guarantee to _not_ `drop` the `Box<Fn>` at the end of our program. But I haven't used it before. If you have an idea for that, I could try it.
> 
> Rust actually never drops any static variable, so even now things won't get dropped when the program exits unless `remove_handler` is called explicitly. The lock-free stuff is probably unnecessary complexity, so let's stick with the current impl.
> 
> > In your example, you have a pointer to a `usize`. However, we have to store a pointer to a `Box<dyn Trait>`, since we can't have a pointer to a `dyn Trait`. Thus, we have to wrap this into another `Box<_>` for which _then_ your examined behavior holds.
> 
> See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=534a07f5e38378c5cd2d6a18552fa2a5) for example of boxing a `FnMut` into a `Mutex` and calling it via a raw ptr. Basically the same use-case as here.
> 

The playground, unfortunately, disregards one aspect as far as I see:
Currently, we store a pointer to a `Box<Fn>`. This is a 64-bit pointer type.
However, a pointer to `dyn Fn` is a 128-bit pointer type. Rust uses a fat pointer to store the length information that it can derive for sized structs.
Thus, I can cast the `*mut c_void` from `pUserData`, which is 64-bit, to `*mut Box<Fn>` but I cannot cast it to `*mut dyn Fn`.
This is why we need the doubled box.

> Also, pls. add a line in CHANGES.md about this feature.

a4df453

rmanoka:
> The playground, unfortunately, disregards one aspect as far as I see:
> Currently, we store a pointer to a `Box<Fn>`. This is a 64-bit pointer type.
> However, a pointer to `dyn Fn` is a 128-bit pointer type. Rust uses a fat pointer to store the length information that it can derive for sized structs.
> Thus, I can cast the `*mut c_void` from `pUserData`, which is 64-bit, to `*mut Box<Fn>` but I cannot cast it to `*mut dyn Fn`.
> This is why we need the doubled box.

Hmm, that's interesting.  Could you add the explanations in the code for future reference?  I suppose we can avoid the extra allocation if we don't use user-data at all, and just use the lock within the error handler.

rmanoka:
I'll go ahead and approve the pr as I don't see any major issues.  Thanks @ChristianBeilschmidt

rmanoka (approved):

ChristianBeilschmidt:
> > The playground, unfortunately, disregards one aspect as far as I see:
> > Currently, we store a pointer to a `Box<Fn>`. This is a 64-bit pointer type.
> > However, a pointer to `dyn Fn` is a 128-bit pointer type. Rust uses a fat pointer to store the length information that it can derive for sized structs.
> > Thus, I can cast the `*mut c_void` from `pUserData`, which is 64-bit, to `*mut Box<Fn>` but I cannot cast it to `*mut dyn Fn`.
> > This is why we need the doubled box.
> 
> Hmm, that's interesting. Could you add the explanations in the code for future reference? I suppose we can avoid the extra allocation if we don't use user-data at all, and just use the lock within the error handler.

44db095

PR #216, opened 2021-09-07
Support 2D in CoordTransform

rmanoka:
---

Currently `CoordTransform::transform` accepts three `&mut [f64]` slices (for x, y, and z coordinates).  This prevents us from using it for 2D transform (whereas in the C api, we would have passed null ptr for z).  This PR supports that by checking if the z slice is empty, and passes null_ptr instead.

Also, currently we only check if the x and y slices have the same length.  This causes sigsegv when passing incorrect z slice.  We've also added a assertion for that.

Modified one of the existing tests that didn't use the z coordinate to instead pass an empty slice.

### DIFF
--- a/src/spatial_ref/srs.rs
+++ b/src/spatial_ref/srs.rs
@@ -2,7 +2,7 @@ use crate::utils::{_last_cpl_err, _last_null_pointer_err, _string};
 use gdal_sys::{self, CPLErr, OGRCoordinateTransformationH, OGRErr, OGRSpatialReferenceH};
 use libc::{c_char, c_int};
 use std::ffi::{CStr, CString};
-use std::ptr;
+use std::ptr::{self, null_mut};
 use std::str::FromStr;
 
 use crate::errors::*;
@@ -42,7 +42,12 @@ impl CoordTransform {
                 nb_coords as c_int,
                 x.as_mut_ptr(),
                 y.as_mut_ptr(),
-                z.as_mut_ptr(),
+                if z.is_empty() {
+                    null_mut()
+                } else {
+                    assert_eq!(nb_coords, z.len());
+                    z.as_mut_ptr()
+                },
             ) == 1
         };
 
--- a/src/spatial_ref/srs.rs
+++ b/src/spatial_ref/srs.rs
@@ -44,7 +44,7 @@ impl CoordTransform {
         assert_eq!(
             nb_coords,
             y.len(),
-            "transform coordinate slices have different length: {} != {}",
+            "transform coordinate slices have different lengths: {} != {}",
             nb_coords,
             y.len()
         );
@@ -60,7 +60,7 @@ impl CoordTransform {
                     assert_eq!(
                         nb_coords,
                         z.len(),
-                        "transform coordinate slices have different length: {} != {}",
+                        "transform coordinate slices have different lengths: {} != {}",
                         nb_coords,
                         z.len()
                     );
--- a/src/spatial_ref/srs.rs
+++ b/src/spatial_ref/srs.rs
@@ -33,9 +33,21 @@ impl CoordTransform {
         })
     }
 
+    /// Transform coordinates in place.
+    ///
+    /// # Arguments
+    /// * x - slice of x coordinates
+    /// * y - slice of y coordinates (must match x in length)
+    /// * z - slice of z coordinates, or an empty slice to ignore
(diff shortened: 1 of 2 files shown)


rmanoka:
One test was failing because of [this line](https://github.com/georust/gdal/blob/6ad95e7f977f5ef69d835d3507c93b884e4ae8ae/src/spatial_ref/tests.rs#L154).  The z slice is longer, but we weren't checking it earlier.  The test was about coordinate range, which would still happen, if not for the assertion that is now triggerred.  Fixed it to pass the same length for all slices.

ChristianBeilschmidt:
Since the method already return a `Result`, why do we assert on the number of inputs and panic?

michaelkirk:
Runtime error vs. panic is often a subjective matter, but panic's are preferable in some circumstances.

I'm not speaking for @rmanoka - but FWIW, I'd also have it panic here, rather than return a runtime error. 

Runtime error's are useful for things that can be handled by the calling code, but this error seems to indicate a bug in the calling code.

from https://doc.rust-lang.org/book/ch09-03-to-panic-or-not-to-panic.html (emphasis mine)
> __If someone calls your code and passes in values that don’t make sense, the best choice might be to call panic! and alert the person using your library to the bug in their code so they can fix it during development.__ Similarly, panic! is often appropriate if you’re calling external code that is out of your control and it returns an invalid state that you have no way of fixing.
>
> However, when failure is expected, it’s more appropriate to return a Result than to make a panic! call. Examples include a parser being given malformed data or an HTTP request returning a status that indicates you have hit a rate limit. In these cases, returning a Result indicates that failure is an expected possibility that the calling code must decide how to handle.

rmanoka:
Ya, I wonder about it too.  There are a couple of other such calls, where we do `assert!` even though we return `Result`.  Example: `RasterBand::read_*`.  I also didn't find a natural candidate in `enum GdalError` so decided to leave it as is for now.

rmanoka:
Ya, I agree with @michaelkirk .  Passing wrong slice lengths is clearly a wrong usage, so the assertion sounds better.   Added some helpful docs declaring the semantics, and also improved the assertion message.

rmanoka:
@jdroenner @michaelkirk @lnicola  anyone have time for a quick review of this?  It's a fairly minor, non-breaking change, but makes it easy to use with 2-D data (which is a fairly common use case)

src/spatial_ref/srs.rs
```diff
@@ -33,16 +33,39 @@ impl CoordTransform {
...
+    /// * x - slice of x coordinates
+    /// * y - slice of y coordinates (must match x in length)
+    /// * z - slice of z coordinates, or an empty slice to ignore
     pub fn transform_coords(&self, x: &mut [f64], y: &mut [f64], z: &mut [f64]) -> Result<()> {
         let nb_coords = x.len();
-        assert_eq!(nb_coords, y.len());
+        assert_eq!(
+            nb_coords,
+            y.len(),
+            "transform coordinate slices have different length: {} != {}",
```
lnicola, line 47: Nit:

```suggestion
            "transform coordinate slices have different lengths: {} != {}",
```

src/spatial_ref/srs.rs
```diff
@@ -33,16 +33,39 @@ impl CoordTransform {
...
                 x.as_mut_ptr(),
                 y.as_mut_ptr(),
-                z.as_mut_ptr(),
+                if z.is_empty() {
+                    null_mut()
+                } else {
+                    assert_eq!(
+                        nb_coords,
+                        z.len(),
+                        "transform coordinate slices have different length: {} != {}",
```
lnicola, line 63:
```suggestion
            "transform coordinate slices have different lengths: {} != {}",
```

lnicola:
Might be worth adding a tests with a 3D transform too.

bors d+

lnicola:
By the way, these are basically nits, so you can just `r+` if you prefer.

lnicola:
Maybe another test with an empty z slice? The failing one doesn't check that it works.

rmanoka:
Thanks @lnicola ;  have added a trivial test for z coordinate (ensure it remains same as input).  I wasn't sure what vertical datums get shipped with all versions of gdal, so didn't want to break any builds by adding a complicated test.  Will merge this once CI passes.

rmanoka:
> Maybe another test with an empty z slice? The failing one doesn't check that it works.

It does because it panics or segfaults if the z slice isn't handled properly.

PR #226, opened 2021-10-18
Rename and delete datasets

ChristianBeilschmidt:
---

The `Driver` was missing the methods `rename` and `delete` (and more). I've added them together with a test.

I was quite unsure to use `&str` or `&Path` as input since we use both in our codebase, in particular, `&Path` for `Dataset::open`. This is why I stuck with `&Path` here. For creating the `CString` out of the `&Path`, I've added a method in util to have a common way to do this.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+- Remove the `datetime` feature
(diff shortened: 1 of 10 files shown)


lnicola:
Looks good, I'll merge it tomorrow unless someone else has remarks.

PR #231, opened 2021-11-13
Prebuilt 34

geohardtke:
Hi there,
I'd like to contribute the pre built bindings for the new 3.4 version of GDAL.


lnicola:
I'd like to merge this tomorrow unless anyone objects.

jdroenner:
i guess we have to adapt these tests first

lnicola:
Is anyone opposed to dropping the `trybuild` thing? It very fragile, especially when trying to support multiple compiler versions, because the error messages sometimes change.

jdroenner:
there is already a discussion here: https://github.com/georust/gdal/pull/234
i'm not against dropping it if we can't make it work

jdroenner:
now we have both options #237

jdroenner:
we should first try to adapt the test. if it is not maintainable,, e.g. if it fails again ext week, then we can drop it.

PR #233, opened 2021-11-18
M darray

### LINKED ISSUE #232: Multidimensional Raster Data Model not yet supported.
GDAL introduced the `Multidimensional Raster Data Model` upstream in version 3.1, and we don't have support for it in this crate. 

I explored the API recently and wrote a few wrappers (See linked [PR](https://github.com/georust/gdal/pull/233) ). It is still very rough, but it could be a useful starting point for a  discussion in case there is some interest in supporting this.

So, please review the code and let me know what you think.

geohardtke:
---

Work in progress to add support for Multidimensional arrays.

resolves #232


geohardtke:
I see ci jobs are failing. The problem is ubuntu 20.04 ships with `gdal 3.0.4`, which doesn't include the Multidimensional array API. Can the ci be configured to grab a more modern version of gdal?

jdroenner:
you need to tag the your code to only be included with gdal >= 3.1 you can do that with something like this:
#[cfg(all(major_ge_3, minor_ge_1))]

we should add another ubunti version with a more modern gdal to the ci. i will look into that.

geohardtke:
> we should add another ubunti version with a more modern gdal to the ci. i will look into that.

Maybe you could set up a pipeline against each of the supported version (versions with pre-built bindings) of gdal. There are several versions available (>2.4) in dockerhub  https://hub.docker.com/r/osgeo/gdal

I'm happy to help with this as well.

jdroenner:
i'm always open to suggestions. Feel free to add ideas for the ci extension as new issue or pr :)

geohardtke:
> i'm always open to suggestions. Feel free to add ideas for the ci extension as new issue or pr :)

See https://github.com/georust/gdal/pull/234 

Some test are failing though any idea how to fix them?

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+        unsafe { GDALMDArrayGetTotalElementsCount(self.c_mdarray) }
+    }
+
+    pub fn get_dimensions(&self) -> Result<Vec<Dimension>> {
+        // break on ndims not is_null
+        let n_dimension = self.n_dimension();
+        unsafe {
+            let mut pn_count: usize = 0;
+            let pn_count_ptr: *mut usize = &mut pn_count;
+            let c_dimensions = GDALMDArrayGetDimensions(self.c_mdarray, pn_count_ptr);
```
ChristianBeilschmidt, line 187: > The returned array must be freed with GDALReleaseDimensions(). If only the array itself needs to be freed, CPLFree() should be called (and GDALDimensionRelease() on individual array members).

Do we account for that somewhere? I guess we need to call it when dropping a dimension and here, we need to free the array?

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+    pub fn n_dimension(&self) -> usize {
+        unsafe { GDALMDArrayGetDimensionCount(self.c_mdarray) }
+    }
+
+    pub fn n_elements(&self) -> u64 {
+        unsafe { GDALMDArrayGetTotalElementsCount(self.c_mdarray) }
+    }
+
+    pub fn get_dimensions(&self) -> Result<Vec<Dimension>> {
+        // break on ndims not is_null
```
ChristianBeilschmidt, line 182: Maybe you can rephrase this: This breaks if the number of dimensions is not null?

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+        let n_dimension = self.n_dimension();
+        unsafe {
+            let mut pn_count: usize = 0;
+            let pn_count_ptr: *mut usize = &mut pn_count;
+            let c_dimensions = GDALMDArrayGetDimensions(self.c_mdarray, pn_count_ptr);
+            // if c_group.is_null() {
+            //     return Err(_last_null_pointer_err("GDALGetRasterBand"));
+            // }
```
ChristianBeilschmidt, line 190: If there is no possibility of a null pointer, you can remove it. Otherwise, we should integrate it.

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+            let mut dimensions: Vec<Dimension> = Vec::new();
+            let mut i = 0;
+            while i < n_dimension {
+                let ptr = c_dimensions.add(i);
+                let next = ptr.read();
+                let value = Dimension::from_c_dimension(self, next);
+                i += 1;
```
ChristianBeilschmidt, line 197: looks like a `for i in 0..n_dimension` loop

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+            let pn_count_ptr: *mut usize = &mut pn_count;
+            let c_dimensions = GDALMDArrayGetDimensions(self.c_mdarray, pn_count_ptr);
+            // if c_group.is_null() {
+            //     return Err(_last_null_pointer_err("GDALGetRasterBand"));
+            // }
+            let mut dimensions: Vec<Dimension> = Vec::new();
+            let mut i = 0;
+            while i < n_dimension {
+                let ptr = c_dimensions.add(i);
+                let next = ptr.read();
```
ChristianBeilschmidt, line 195: Do you have to check for null-pointers here?

Maybe you can create a helper function that is similar to `_string_array`, just for Dimensions

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+                let value = Dimension::from_c_dimension(self, next);
+                i += 1;
+                dimensions.push(value);
+            }
+            Ok(dimensions)
+        }
+    }
+
+    pub fn read_into_slice<T: Copy + GdalType>(
```
ChristianBeilschmidt, line 204: Maybe you can add some docs or copy some docs from the original documentation to understand the params.

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+        &self,
+        buffer: &mut [T],
+        array_start_index: Vec<u64>,
+        count: Vec<usize>,
+    ) -> Result<()> {
+        // let array_start_index = [array_start_index.0, array_start_index.1];
+        // let count =  [count.0, count.1];
```
ChristianBeilschmidt, line 211:
```suggestion
```

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+        let array_step: *const i64 = std::ptr::null();
+        let buffer_stride: *const i64 = std::ptr::null();
```
ChristianBeilschmidt, line 213: Maybe add a comment that this means using defaults, e.g. [1, …] for the steps.

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+        let rv = unsafe {
+            gdal_sys::GDALMDArrayRead(
+                self.c_mdarray,
+                array_start_index.as_ptr(),
+                count.as_ptr(),
+                array_step,
+                buffer_stride,
+                data_type,
+                buffer.as_mut_ptr() as *mut c_void, // pDstBuffer: *mut libc::c_void,
+                p_dst_buffer_alloc_start,           // pDstBufferAllocStart: *const libc::c_void,
```
ChristianBeilschmidt, line 227:
```suggestion
                buffer.as_mut_ptr() as *mut c_void,
                p_dst_buffer_alloc_start,
```

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+                n_dst_buffer_alloc_size,
+            )
+        };
+
+        if rv != 1 {
+            return Err(_last_cpl_err(rv.try_into().unwrap())); // this is probably incorrect!
+        }
```
ChristianBeilschmidt, line 234: What are your thoughts on this one? Maybe we can help getting this correct.

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+
+        // Safety: the read_into_slice line below writes
+        // exactly pixel elements into the slice, before we
+        // read from this slice. This paradigm is suggested
+        // in the rust std docs
+        // (https://doc.rust-lang.org/std/vec/struct.Vec.html#examples-18)
+        unsafe {
+            data.set_len(pixels);
+        };
+        self.read_into_slice(&mut data, array_start_index, count)?;
```
ChristianBeilschmidt, line 260: In the example, they read into the array-ptr and then set the length. Does that make any difference?

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+    ) -> Result<ArrayD<T>> {
+        let data = self.read_as::<T>(array_start_index, count)?;
+        // Matrix shape is (rows, cols) and raster shape is (cols in x-axis, rows in y-axis)
+
+        let dim: IxDyn = IxDyn(&array_size);
+        Ok(ArrayD::from_shape_vec(dim, data)?)
+    }
+}
+
+/// Represent a mdarray in a dataset
```
ChristianBeilschmidt, line 289:
```suggestion
/// Represent a `Group` in a `Dataset`
```

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+    pub fn array_names(&self, options: CslStringList) -> Vec<String> {
+        let options = options.as_ptr();
+        let c_array_names = unsafe { GDALGroupGetMDArrayNames(self.c_group, options) };
+        _string_array(c_array_names)
+    }
+
+    pub fn md_array(&self, name: String, options: CslStringList) -> MDArray {
+        let name = CString::new(name).unwrap();
+        let c_mdarray =
+            unsafe { GDALGroupOpenMDArray(self.c_group, name.as_ptr(), options.as_ptr()) };
```
ChristianBeilschmidt, line 321: > Returns    the array names, to be freed with CSLDestroy()

Do we do this somewhere?

I guess @jdroenner used valgrind at some point. Maybe we can do this here. I also struggle to get these C-calls and frees right, so maybe this helps spotting errors.

src/raster/mdarray.rs
```diff
@@ -0,0 +1,331 @@
...
+
+        MDArray::from_c_mdarray(self, c_mdarray)
+    }
+    // pub unsafe fn array(&self, array_name: String, options: CslStringList) -> MDArray{
+    //     let name = CString::new(array_name).unwrap();
+    //     let options = options.as_ptr();
+    //     let array_h =  GDALGroupOpenMDArray(self.c_group, name.as_ptr(), options);
+    //     MDArray::from_c_mdarray(&self, array_h)
+    // }
```
ChristianBeilschmidt, line 330:
```suggestion
```

ChristianBeilschmidt (changes requested):
Thanks for the work!

Since I've been trying to use the multi-dimensional API from GDAL recently, I have a little experience to at least give you a first PR for this.

I've spotted some things and asked some questions. We have to be especially careful with memory handling where GDAL expects us to free things.

jdroenner:
@geohardtke  are you still working on this? we merged the 3.4 integration.

geohardtke:
Hi @jdroenner  and @ChristianBeilschmidt. Thanks for having a look. I'll try to smooth the rough edges a bit and update the branch.

jdroenner:
closed this PR and merged 273

PR #234, opened 2021-11-22
upgrade ci

geohardtke:
CI should now build and test against gdal 3.0, 3.1, 3.2, 3.3 and 3.4

Unfortunately some tests are now failing. Any idea?


geohardtke:
Some more work is needed here, but I don't have time right now. Will come back here later

geohardtke:
in gdal 3.2 and gdal 3.3; test_sql_with_dialect is broken. 

```
test vector::vector_tests::tests::test_features_aliasing_compile_fail ... ok

failures:

---- vector::vector_tests::sql::test_sql_with_dialect stdout ----
thread 'vector::vector_tests::sql::test_sql_with_dialect' panicked at 'called `Result::unwrap()` on an `Err` value: CplError { class: 3, number: 1, msg: "In ExecuteSQL(): sqlite3_prepare_v2(SELECT * FROM roads WHERE highway = 'pedestrian' and NumPoints(GEOMETRY) = 3):\n  no such function: NumPoints" }', src/vector/vector_tests/sql.rs:82:10


failures:
    vector::vector_tests::sql::test_sql_with_dialect
 ```

then in gdal 3.4 a second test breaks; auto_identify : 

```
---- spatial_ref::tests::auto_identify stdout ----
thread 'spatial_ref::tests::auto_identify' panicked at 'called `Result::unwrap()` on an `Err` value: OgrError { err: 7, method_name: "OSRAutoIdentifyEPSG" }', src/spatial_ref/tests.rs:207:38
```

Which shows why it is a good idea to test against as many versions as posible.

**Comment by geohardtke** [comment]
Ok. so the problem with one of the function was probably a driver not available on the minimal ubuntu base, swapping to the full image solved the problem.

**Comment by jdroenner** [comment]
great work! 
i guess the failure in 3.4 is caused by Proj not accepting the SRS identification we have in that test.

**Comment by geohardtke** [comment]
> Proj not accepting the SRS

Any idea how to fix it? I'm not very familiar with that side of the API.

**Comment by michaelkirk** [comment]
I was able to get the test to pass with these changes:
https://github.com/geohardtke/gdal/compare/upgrade_ci...michaelkirk:mkirk/upgrade-ci?expand=1

I've never used the WKT SRS format before. Is it possible that it changed with the more recent versions?

**Comment by michaelkirk** [comment]
bors try

**Comment by geohardtke** [comment]
> I was able to get the test to pass with these changes: https://github.com/geohardtke/gdal/compare/upgrade_ci...michaelkirk:mkirk/upgrade-ci?expand=1
> 


Excellent, thanks! We do have all green lights now.

Maybe it would be nice to add a note about this in the release notes when gdal 3.4 bindings are added? Unfortunately I can't  find anything in the release notes upstream to explain the change in behaviour.  I'm not even sure if it is a change in gdal or proj?

**Comment by michaelkirk** [comment]
I don't know either. One data point I have is that I was actually seeing this same error locally with gdal 3.3 and proj 8.2 (on apple aarch64). 

So that might point to it being related to the underlying proj instance... maybe 🤷

**Comment by geohardtke** [comment]
yes, I'm inclined to think that is a change in proj. I just checked and the docker image for gdal 3.4 uses proj 8.2 while the 3.3 version uses 8.0.1. 
So the actual behaviour will depend on the proj version available in the system and not the gdal version. So probably we don't need to add anything to the release notes.

Thanks again!

**Comment by jdroenner** [comment]
thanks! this a huge improvement. I also think think it depends on the proj version.

**Comment by jdroenner** [comment]
bors try

**Comment by jdroenner** [comment]
the ci is not happy with some of the tests :(

**Comment by michaelkirk** [comment]
I was able to get 69b64a4f849e895d91957affe68b7dcc0e145d3c to pass.

There's no visible diff:
https://github.com/georust/gdal/compare/69b64a4f849e895d91957affe68b7dcc0e145d3c...fae64cc39b5760aa6f95112bb186e9d6c2a7896b

Seemingly something in the infrastructure changed around this test of a compiler error (notice the underlined section in the error): https://github.com/georust/gdal/runs/4397358541?check_suite_focus=true#step:10:268
```
let _ = layer.defn();
268
  |                 ^^^^^ immutable borrow occurs here
```
vs:
```
9 |         let _ = layer.defn();
282
  |                 ^^^^^^^^^^^^ immutable borrow occurs here
```

I'm not sure why we're testing a compiler error in the gdal crate, but if we want to keep it maybe it's as simple as updating the expected output.

jdroenner:
maybe we should just remove test if it depends on the compiler version

michaelkirk:
Looks like the test was added in https://github.com/georust/gdal/pull/161

I don't remember that PR at all, but I left some comments in it at the time, so I cannot claim innocence. 😅 

After re-reviewing - I think the idea of the test is:
Semantically these things are mutable references and should be treated as such. So let's mark them as mut and have the rust compiler help us.  But they are not actually mutated in rust, so it would be possible to mark them as immutable. So we have this test as a reminder in case some well meaning persons were to remove the "unnecessary mut".

Is that about right @rmanoka?


I could go either way. It's a quick update to make, and now that I remember why it exists, it seems like a useful test.

jdroenner:
bors try

jdroenner:
we merged #231 . Since this PR also includes the prebuild files i guess there will be a merge conflict. @geohardtke can rebase this PR to master?

lnicola:
bors retry

lnicola:
bors try-

lnicola:
bors try

lnicola:
Looks like `try` worked, but didn't report the status for some reason (misconfigured bors on that branch?). I don't think the bindings are a problem, because they're identical and don't cause a merge conflict. So we can merge this right now.

jdroenner:
great then lets try it

rmanoka:
Yeah, the compile tests were added to check rust semantics are enforced well.   I'm okay with removing them if maintaining it is messy.  We should have decent documentation explaining the semantics anyway.

lnicola:
The reasons I think it would be better to remove that test are:

 - the compiler error messages sometimes change
 - it makes it harder to support multiple compiler versions
 - a subtler issue is that with `cargo check --all-features`, the compile test runs `cargo check` with the default features, causing a rebuild; this takes two minutes on CI

--

Regarding this PR, I'm not sure why bors times out. This branch seems to include the fixes from August.

jdroenner:
bors retry

jdroenner:
bors retry

lnicola:
We might have to merge this manually.

jdroenner:
yes. just thought it should try again :laughing:

jdroenner:
i will rebase and merge it later

jdroenner:
ok now we have a new issue. Thanks clippy -_-

jdroenner:
i think we would have to make the FFI  call in the same  method. we can not use a &mut ref ans pass it to the other method.

jdroenner:
@rmanoka this way it works. Don't know what i like better.

jdroenner:
bors try

rmanoka:
@jdroenner I like your approach; it's a bit of copied code, but safer.  We used `set_len` earlier because this could be part of a hot-loop in many raster reading paradigms.  That said, safety and easier maintenance is more important, so I don't mind either the copied code or the `T::zero()` approach.  There may be a way using `MaybeUninit` as clippy suggests, but it's beyond my understanding.

rmanoka:
Another option, if you would like to not have copied code:  move the common `GDALRasterIOEx` code into a private fn that takes *mut instead of &mut.  Then both the method returning `Buffer` and the `read_into_slice` can call that.

jdroenner:
lets use it like this then

jdroenner:
bors merge

jdroenner:
i guess the CI is just to much for bors

PR #236, opened 2022-01-03
Drop the compile-fail test

lnicola:
There is a single `trybuild` test but it's very fragile because the
error messages tend to change. Since there's no multi-baseline
support in `trybuild`, this makes supporting multiple compiler
versions impossible.

---

CC #231


lnicola:
bors try

PR #237, opened 2022-01-03
adapt try build expected error to new output

jdroenner:
---
it works with current stable beta and nightly on linux


lnicola:
:disappointed:

bors r+

lnicola:
See https://github.com/georust/gdal/runs/4691714467?check_suite_focus=true

This also adds 1 minute and a half or so to the run time of the `--all-features` test because it builds the test without `--all-features`, causing an extra recompile.

PR #239, opened 2022-01-10
Add wrappers for GDALBuildVRT, GDALApplyGeoTransform, GDALInvGeoTransform

amartin96:
---

I did not know where to put `apply_geo_transform` and `inv_geo_transform`, so I left them in `lib.rs`. Please let me know where they should go.

I also did not know what manner of tests should be included. Please let me know what I should contribute in that regard.


src/lib.rs
```diff
@@ -42,6 +43,37 @@ pub use dataset::{
...
+        );
+    }
+    (geo_x, geo_y)
+}
+
+/// Invert Geotransform.
+/// Wraps [GDALInvGeoTransform].
+///
+/// [GDALInvGeoTransform]: https://gdal.org/api/raster_c_api.html#_CPPv419GDALInvGeoTransformPdPd
+pub fn inv_geo_transform(geo_transform: &GeoTransform) -> GeoTransform {
```
lnicola, line 69:
```suggestion
pub fn invert_geo_transform(geo_transform: &GeoTransform) -> Result<GeoTransform> {
```

This function is fallible, we should expose that in the interface.

src/raster_programs.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    // Convert args to raw strings
+    let mut c_args = Vec::<*mut c_char>::new();
+    for arg in args {
+        c_args.push(CString::new(arg)?.into_raw());
+    }
+    c_args.push(null_mut()); // Null-terminate it
+
+    let result = unsafe {
+        let options = gdal_sys::GDALBuildVRTOptionsNew(c_args.as_mut_ptr(), null_mut());
```
lnicola, line 34: I'd prefer wrapper around this that calls `GDALBuildVRTOptionsFree` on drop.

src/raster_programs.rs
```diff
@@ -0,0 +1,72 @@
...
+        None => null_mut(),
+    };
+
+    // Convert args to raw strings
+    let mut c_args = Vec::<*mut c_char>::new();
+    for arg in args {
+        c_args.push(CString::new(arg)?.into_raw());
+    }
+    c_args.push(null_mut()); // Null-terminate it
+
```
lnicola, line 32: Something like `let c_options_ptrs = std::iter::chain(args.iter().map(|o| o.as_ptr()), std::iter::once(null_mut())).collect::<Vec<_>>(); ` would be nicer here because you don't have to free them afterwards.

src/raster_programs.rs
```diff
@@ -0,0 +1,72 @@
...
+/// [GDALBuildVRT]: https://gdal.org/api/gdal_utils.html#gdal__utils_8h_1a057aaea8b0ed0476809a781ffa377ea4
+/// [program docs]: https://gdal.org/programs/gdalbuildvrt.html
+pub fn build_vrt<D: Borrow<Dataset>>(
+    dest: Option<String>,
+    datasets: &[D],
+    args: Vec<String>,
+) -> Result<Dataset> {
+    // Convert dest to raw string
+    let dest = match dest {
+        Some(s) => CString::new(s)?.into_raw(),
```
lnicola, line 22: Same as below, `dest.map(|p| p.as_mut_ptr())`.

src/raster_programs.rs
```diff
@@ -0,0 +1,72 @@
...
+};
+
+/// Build a VRT from a list of datasets.
+/// Wraps [GDALBuildVRT].
+/// See the [program docs] for more details.
+///
+/// [GDALBuildVRT]: https://gdal.org/api/gdal_utils.html#gdal__utils_8h_1a057aaea8b0ed0476809a781ffa377ea4
+/// [program docs]: https://gdal.org/programs/gdalbuildvrt.html
+pub fn build_vrt<D: Borrow<Dataset>>(
+    dest: Option<String>,
```
lnicola, line 16: This should probably be `Option<Path>` or similar?

src/raster_programs.rs
```diff
@@ -0,0 +1,72 @@
...
+    datasets: &[D],
+    args: Vec<String>,
+) -> Result<Dataset> {
```
lnicola, line 19: It might be nicer to split this function into generic and non-generic parts, like in https://doc.rust-lang.org/src/std/path.rs.html#1237-1239.

src/raster_programs.rs
```diff
@@ -0,0 +1,72 @@
...
+            let _ = CString::from_raw(dest);
+        }
+        for c_arg in c_args {
+            if !c_arg.is_null() {
+                let _ = CString::from_raw(c_arg);
+            }
+        }
+
+        if dataset_out.is_null() {
+            return Err(GdalError::NullPointer {
```
lnicola, line 62: There's a `_last_null_pointer_err` helper that gets the correct error message.

amartin96:
Thanks for the feedback. I've summarized my updates below:

- `invert_geo_transform` now returns a result.
- Added `struct BuildVRTOptions` that frees the owned pointer on drop.
- `build_vrt`
   - Uses `_last_null_pointer_err` to populate the error result.
   - Employs that thin generic wrapper pattern. (is there a name for this?)
   - Takes `dest: Option<&Path>` instead of `String`, though it gets converted using `Path::to_string_lossy`

I didn't take the suggestions re: raw strings verbatim, as I didn't see how they added terminating null bytes. I did look at `GDALBuildVRTOptionsNew` and determine that it doesn't actually mutate the strings, so I left them owned by the `CString`s, which hopefully addresses part of the concern.

src/raster_programs.rs
```diff
@@ -0,0 +1,107 @@
...
+            .iter()
+            .map(|x| x.borrow())
+            .collect::<Vec<&Dataset>>(),
+        args,
+    )
+}
+
+fn _build_vrt(dest: Option<&Path>, datasets: &[&Dataset], args: Vec<String>) -> Result<Dataset> {
+    // Convert dest to CString
+    let dest = match dest {
```
lnicola, line 75: It's not great, but let's use `utils::_path_to_c_string` here.

src/raster_programs.rs
```diff
@@ -0,0 +1,107 @@
...
+    c_options: *mut GDALBuildVRTOptions,
+}
+
+impl BuildVRTOptions {
+    /// See [GDALBuildVRTOptionsNew].
+    ///
+    /// [GDALBuildVRTOptionsNew]: https://gdal.org/api/gdal_utils.html#_CPPv422GDALBuildVRTOptionsNewPPcP28GDALBuildVRTOptionsForBinary
+    fn new(args: Vec<String>) -> Result<Self> {
+        // Convert args to CStrings to add terminating null bytes
+        let mut cstr_args = Vec::<CString>::new();
```
lnicola, line 24: Untested, but:

```rust
let cstr_args = args.into_iter()
    .map(|arg| CString::new(arg))
    .collect::<Result<Vec<_>>>()?;
```

This looks nicer and (as a minor point) pre-allocates the destination `Vec`.

amartin96:
Incorporated latest feedback.

src/raster_programs.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/// Build a VRT from a list of datasets.
+/// Wraps [GDALBuildVRT].
+/// See the [program docs] for more details.
+///
+/// [GDALBuildVRT]: https://gdal.org/api/gdal_utils.html#gdal__utils_8h_1a057aaea8b0ed0476809a781ffa377ea4
+/// [program docs]: https://gdal.org/programs/gdalbuildvrt.html
+pub fn build_vrt<D: Borrow<Dataset>>(
+    dest: Option<&Path>,
```
lnicola, line 65: Sorry for missing this. I guess `dest: Option<P> ... where P: AsRef<Path>` would be nicer, so you can call it with strings.

lnicola (approved):

amartin96:
@lnicola Taking `dest: Option<P> ... where P: AsRef<Path>` has made it so a type annotation is needed when passing `None`:

```
gdal::raster_programs::build_vrt(None, datasets, args)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `P` declared on the function `build_vrt`
```

A default type cannot be specified because that is not allowed for functions:

```
error: defaults for type parameters are only allowed in `struct`, `enum`, `type`, or `trait` definitions
pub fn build_vrt<D: Borrow<Dataset>, P: AsRef<Path> = &'static str>(
                                     ^
```

Is there a cleaner way to do this?

lnicola:
Ah, good point. You can still specify the type, but it's an ugly API. Let's revert this change .

src/lib.rs
```diff
@@ -42,6 +43,43 @@ pub use dataset::{
 pub use driver::Driver;
 pub use metadata::Metadata;
 
+/// Apply GeoTransform to x/y coordinate.
```
rmanoka, line 46: The `GeoTransform` relevant methods might be better placed inside the `dataset` or the `rasters` module.  It is anyway only relevant to rasters IIUC.

src/lib.rs
```diff
@@ -42,6 +43,43 @@ pub use dataset::{
...
+/// Wraps [GDALApplyGeoTransform].
+///
+/// [GDALApplyGeoTransform]: https://gdal.org/api/raster_c_api.html#_CPPv421GDALApplyGeoTransformPdddPdPd
+pub fn apply_geo_transform(geo_transform: &GeoTransform, pixel: f64, line: f64) -> (f64, f64) {
```
rmanoka, line 50: Would it be more ergonomic to throw these inside a trait that is impl by `GeoTransform` so we get the nicer:

```rust
let (wx, wy) = geo_t.apply(pix, line);
let inv_t = geo_t.invert()?;
```

Optional suggestion.

src/raster_programs.rs
```diff
@@ -0,0 +1,107 @@
+use crate::{
```
rmanoka, line 1: Would be nice to move these into a sub-module `raster_programs::vrt` or similar.   Not a major concern, but this mod. might get cluttered if we bring in more of the programs like Warp, etc.

src/raster_programs.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+/// Build a VRT from a list of datasets.
+/// Wraps [GDALBuildVRT].
+/// See the [program docs] for more details.
+///
+/// [GDALBuildVRT]: https://gdal.org/api/gdal_utils.html#gdal__utils_8h_1a057aaea8b0ed0476809a781ffa377ea4
+/// [program docs]: https://gdal.org/programs/gdalbuildvrt.html
+pub fn build_vrt<D: Borrow<Dataset>>(
```
rmanoka, line 64: Could we expose the `BuildVRTOptions` struct, and let this method take an `Option<BuildVrtOptions>` instead of a `Vec<String>` ?  Seems more similar to the C api, and also a bit more natural.

src/raster_programs.rs
```diff
@@ -0,0 +1,107 @@
...
+/// [GDALBuildVRTOptions]: https://gdal.org/api/gdal_utils.html#_CPPv419GDALBuildVRTOptions
+struct BuildVRTOptions {
+    c_options: *mut GDALBuildVRTOptions,
+}
+
+impl BuildVRTOptions {
+    /// See [GDALBuildVRTOptionsNew].
+    ///
+    /// [GDALBuildVRTOptionsNew]: https://gdal.org/api/gdal_utils.html#_CPPv422GDALBuildVRTOptionsNewPPcP28GDALBuildVRTOptionsForBinary
+    fn new(args: Vec<String>) -> Result<Self> {
```
rmanoka, line 26: Does `args` have to be a `Vec<String>`?  Wouldn't any `IntoIterator<Item = ...>` work?  This would avoid allocations if it's a fixed sized array.

Going one more step, does the inner type have to be `String` or anything that impls `TryInto<CString>` (like `Vec<U8>` or `Vec<NonZeroU8>`, etc.)?

rmanoka (changes requested):
@amartin96 Nice PR!  The overall code looks great; my suggestions are mostly cosmetic.

src/raster_programs.rs
amartin96: I attempted to implement this as follows:
```rust
pub fn new<S: TryInto<CString>, I: IntoIterator<Item = S>>(args: I) -> Result<Self> {
    let cstr_args = args.into_iter().map(|x| x.try_into()).collect::<std::result::Result<Vec<CString>, _>>()?;
```
However, I do not know how to convert `TryInto<CString>::Error` to `GdalError::FfiNulError`.

Can you show me how its done? Is the rest of it as you intended?

rmanoka: Hmm, good pt.; we can't give a good error for arbitrary conversion failures, and the user could anyway `map` as req. before passing it.  How about:

```
fn new<S: Into<Vec<u8>>, I: IntoIterator<Item = S>>(args: I) -> Result<Self> {
    let c_args = args
        .into_iter()
        .map(|x| CString::new(x))
        .collect::<std::result::Result<Vec<CString>, _>>()?;
    ....
}
```
This is inline with the rest of the lib, where we expect `&str` to pass `CString` to the FFI.  The above seems to accept both `&str`, and `String`, so that should be good.

rmanoka (approved):

rmanoka:
Changes look good to me; thanks again @amartin96 .  Pinging @lnicola @jdroenner :  could you pls. allow running CI and merge if all looks good?

lnicola:
I took a hard page fault on this one :smile:.

CHANGES.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+- Add `raster_programs::build_vrt`
```
```suggestion
- Add `programs::raster::build_vrt`
- Add `GeoTransformTrait` extension trait with `apply` and `invert`
```

rmanoka:
@amartin96 One final nitpick:  it would be good to rename `GeoTransformTrait` as `GeoTransformEx`.  That's the typical rust naming convention for traits to add functionalities.

src/dataset.rs
```diff
@@ -31,6 +31,48 @@ use bitflags::bitflags;
...
+        }
+        (geo_x, geo_y)
+    }
+
+    /// Invert Geotransform.
+    /// Wraps [GDALInvGeoTransform].
+    ///
+    /// [GDALInvGeoTransform]: https://gdal.org/api/raster_c_api.html#_CPPv419GDALInvGeoTransformPdPd
+    fn invert(&self) -> Result<GeoTransform> {
+        let mut gt_out: GeoTransform = Default::default();
```
lnicola, line 64: We could use [`MaybeUninit`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html) here, but it's fine to leave it like this.

Minor nit: I prefer keeping as much code as possible outside `unsafe` blocks, as in `let rv = unsafe { ... }; if rv == 0 { ... }`.

CHANGES.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+- Add `programs::raster::build_vrt`
+- Add `GeoTransformTrait` extension trait with `apply` and `invert`
+- Add `apply_geo_transform`
```
lnicola, line 7: These two lines need to be removed

lnicola:
bors d+

PR #244, opened 2022-01-19
update deps

jdroenner:
---


lnicola:
It's broken, isn't it?

jdroenner:
bors? I guess the new CI is too much for bors.

lnicola:
I guess https://github.com/georust/gdal/blob/master/bors.toml needs to be updated with the job names from the workflow.

jdroenner:
uh good catch

jdroenner:
bors try

lnicola:
That won't work, you need to update `bors.toml` on `master` before using it.

jdroenner:
thats strange

PR #245, opened 2022-01-19
Bump edition to 2021 & update changelog to release 0.12

jdroenner:
- [ x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+## 0.12
+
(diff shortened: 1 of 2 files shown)


lnicola:
I think we also need to bump the version.

jdroenner:
in the cargo toml? i  think the  cargo release tool will do  that

lnicola:
Oh, okay.

PR #246, opened 2022-01-26
Implement ColorTable struct

Barugon:
Implement a `ColorTable` struct and add a `RasterBand::color_table` method.

- [x] I added an entry to `CHANGES.md`.
---


src/raster/rasterband.rs
```diff
@@ -600,3 +612,51 @@ impl ColorInterpretation {
         _string(rv)
     }
 }
+
+pub struct ColorTable<'a> {
```
ChristianBeilschmidt, line 616: Please add some docs on this struct.

src/raster/rasterband.rs
```diff
@@ -600,3 +612,51 @@ impl ColorInterpretation {
...
+            c4: 0,
+        };
+        if unsafe { gdal_sys::GDALGetColorEntryAsRGB(self.c_color_table, index, &mut color_entry) }
+            == 0
+        {
+            return None;
+        }
+        Some(color_entry)
+    }
+}
```
ChristianBeilschmidt, line 662: Could you create a test that uses each of these methods?
Otherwise, we don't know that they work.

As far as I see there is no need to free any value wrt. leaking.

src/raster/rasterband.rs
```diff
@@ -600,3 +612,51 @@ impl ColorInterpretation {
...
+
+pub struct ColorTable<'a> {
+    c_color_table: GDALColorTableH,
+    raster_band: &'a RasterBand<'a>,
+}
+
+impl<'a> ColorTable<'a> {
+    pub fn raster_band(&self) -> &RasterBand {
+        self.raster_band
+    }
```
ChristianBeilschmidt, line 624: Why do we need to return the raster band?

ChristianBeilschmidt (changes requested):
Thank you for the PR. I have some comments.

src/raster/rasterband.rs
ChristianBeilschmidt: Can you generate some small test data, for instance with the GDAL CLI?

src/raster/rasterband.rs
ChristianBeilschmidt: Okay, I vote for removing it.

src/raster/rasterband.rs
ChristianBeilschmidt: It's okay now. Unfortunately, I cannot resolve this issue.

src/raster/rasterband.rs
Barugon: It's removed.

src/raster/rasterband.rs
Barugon: Is there someplace specific that I should put the test file (a small geo-tiff, 22k) or is the raster folder okay?

Barugon: Done.

src/raster/tests.rs
```diff
@@ -693,3 +693,38 @@ fn test_rasterize() {
...
+        band.color_interpretation(),
+        ColorInterpretation::PaletteIndex
+    );
+
+    // Color table is RGB.
+    let color_table = band.color_table().unwrap();
+    assert_eq!(
+        color_table.palette_interpretation(),
+        gdal_sys::GDALPaletteInterp::GPI_RGB
+    );
```
ChristianBeilschmidt, line 715: I'm not sure about outputting a `gdal_sys` type here. That would mean you have to depend on the sys crate to use this API. Maybe we should wrap the `gdal_sys::GDALPaletteInterp::GPI_RGB` into a custom enum like `ColorInterpretation` does.

src/raster/tests.rs
```diff
@@ -693,3 +693,38 @@ fn test_rasterize() {
...
+
+    // Check that entry and entry_as_rgb are the same.
+    for index in 0..entry_count {
+        let entry = color_table.entry(index).unwrap();
+        let rgb_entry = color_table.entry_as_rgb(index).unwrap();
+        assert_eq!(entry.c1, rgb_entry.c1);
+        assert_eq!(entry.c2, rgb_entry.c2);
+        assert_eq!(entry.c3, rgb_entry.c3);
+        assert_eq!(entry.c4, rgb_entry.c4);
+    }
```
ChristianBeilschmidt, line 729: `GDALColorEntry` could work this way, although it is also from `gdal_sys`. But it would be maybe nicer to have a more readable struct. `c1` could be gray, _red, cyan or hue_  so it is not easy to name it. For the RGB part, we could split the structs to have one with r, g, b, a fields. What do you think?

ChristianBeilschmidt (changes requested):

src/raster/tests.rs
Barugon: I'll add an enum.

src/raster/tests.rs
Barugon: Maybe a `ColorEntry` enum that has the correct interpretation and we could forgo `palette_interpretation`?

Barugon: Have a look at the latest changes and let me know what you think.

src/raster/rasterband.rs
```diff
@@ -600,3 +609,143 @@ impl ColorInterpretation {
...
+    fn from_c_int(palette_interpretation: GDALPaletteInterp::Type) -> Option<Self> {
+        match palette_interpretation {
+            GDALPaletteInterp::GPI_Gray => Some(Self::Gray),
+            GDALPaletteInterp::GPI_RGB => Some(Self::Rgba),
+            GDALPaletteInterp::GPI_CMYK => Some(Self::Cmyk),
+            GDALPaletteInterp::GPI_HLS => Some(Self::Hls),
+            _ => None,
+        }
+    }
+}
```
ChristianBeilschmidt, line 631:
```suggestion
impl PaletteInterpretation {
    fn from_c_int(palette_interpretation: GDALPaletteInterp::Type) -> Self {
        match palette_interpretation {
            GDALPaletteInterp::GPI_Gray => Self::Gray,
            GDALPaletteInterp::GPI_RGB => Self::Rgba,
            GDALPaletteInterp::GPI_CMYK => Self::Cmyk,
            GDALPaletteInterp::GPI_HLS => Self::Hls,
            _ => unreachable!("GDAL has implemented a new type of `GDALPaletteInterp`"),
        }
    }
}
```

If you unwrap it anyway, you could just panic here.

ChristianBeilschmidt (changes requested):
From my point of view, this looks now very good.
The really last comment from my side, then it is mergeable for me.

src/raster/tests.rs
ChristianBeilschmidt: :+1:

src/raster/tests.rs
ChristianBeilschmidt: :+1:

ChristianBeilschmidt:
> 1 workflow awaiting approval 

@jdroenner could you approve the CI for this PR?

src/raster/rasterband.rs
```diff
@@ -600,3 +609,143 @@ impl ColorInterpretation {
...
+        }
+    }
+
+    /// How the values of this color table are interpreted.
+    pub fn palette_interpretation(&self) -> PaletteInterpretation {
+        self.palette_interpretation
+    }
+
+    /// Get the number of color entries in this color table.
+    pub fn entry_count(&self) -> i32 {
```
Barugon, line 696: @ChristianBeilschmidt 
Do you think this should return a `usize`? Same question for the `index` parameter of `entry` and `entry_as_rgb`.

ChristianBeilschmidt: Very good point.
For some reason, `layer_count` and `raster_count` on `Dataset` both return 'isize`.
Does maybe @jdroenner know why?

Does GDAL at some point output a negative number? We can't put the whole size range into a function that requires int on the GDAL side, but actually an unsigned type makes more sense.

Barugon: `GDALGetColorEntryCount` returns an `int` but I don't see how it could be negative. The documentation simply says "Get number of color entries in table". Maybe the safest thing is to leave it as `i32`?

ChristianBeilschmidt: For me, it should be an unsigned int in a Rust API. One could argue that if we have `i32` with only positive numbers, we should do something like `u16`. But the _Rusty_ way is `usize`.

But maybe there is some convention for this API that it uses ints for this. Maybe someone else has an opinion for this.

Barugon: I think that I'll go ahead and make it `usize`. I can change it back it it's decided otherwise.

Barugon:
Any chance of getting this merged?

jdroenner:
yes i'm just very busy at the moment. will try to look into it this week.

ChristianBeilschmidt (approved):

Barugon:
@lnicola 
I think this squash went better.

lnicola:
Thanks, I'll take a better look tomorrow.

src/raster/rasterband.rs
```diff
@@ -612,3 +621,144 @@ impl ColorInterpretation {
...
+            GDALPaletteInterp::GPI_CMYK => Self::Cmyk,
+            GDALPaletteInterp::GPI_HLS => Self::Hls,
+            _ => unreachable!("GDAL has implemented a new type of `GDALPaletteInterp`"),
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct GrayEntry {
+    pub g: i16,
```
lnicola, line 647: Similar to the `i32` / `usize` discussion: do you know how often these happen to be negative? I mean, if I have a 16-bit RGB image, I'll still want positive numbers here.

Anyway, this isn't blocking, I was just wondering how GDAL expects these to be used.

src/raster/rasterband.rs
```diff
@@ -612,3 +621,144 @@ impl ColorInterpretation {
...
+
+/// Color table for raster bands that use the PaletteIndex color interpretation.
+///
+/// This object carries the lifetime of the raster band that
+/// contains it. This is necessary to prevent the raster band
+/// from being dropped before the color table.
+pub struct ColorTable<'a> {
+    palette_interpretation: PaletteInterpretation,
+    c_color_table: GDALColorTableH,
+    _raster_band: &'a RasterBand<'a>,
```
lnicola, line 689: Can this be a `PhantomData<&'a RasterBand<'a>>` or something?

src/raster/rasterband.rs
Barugon: I don't think they would ever be negative but I thought it might be best to leave them the same as what gdal_sys provides. I suppose elevation data could be negative but 
I can't imagine that ever being palletized.

src/raster/rasterband.rs
Barugon: I was unaware of `PhantomData`. Thanks for pointing it out.

lnicola:
bors r=ChristianBeilschmidt,lnicola

PR #250, opened 2022-01-27
Add geometry_type_to_name function

ttencate:
---

### DIFF
--- a/src/vector/geometry.rs
+++ b/src/vector/geometry.rs
@@ -437,11 +437,18 @@ impl PartialEq for Geometry {
 
 impl Eq for Geometry {}
 
+pub fn geometry_type_to_name(ty: OGRwkbGeometryType::Type) -> String {
+    let rv = unsafe { gdal_sys::OGRGeometryTypeToName(ty) };
+    // If the type is invalid, OGRGeometryTypeToName returns a valid string anyway.
+    assert!(!rv.is_null());
+    _string(rv)
+}
+
 #[cfg(test)]
 mod tests {
     use crate::spatial_ref::SpatialRef;
 
-    use super::Geometry;
+    use super::{geometry_type_to_name, Geometry};
 
     #[test]
     #[allow(clippy::float_cmp)]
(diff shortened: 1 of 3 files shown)


src/vector/geometry.rs
```diff
@@ -437,11 +437,16 @@ impl PartialEq for Geometry {
 
 impl Eq for Geometry {}
 
+pub fn geometry_type_to_name(ty: OGRwkbGeometryType::Type) -> String {
+    let rv = unsafe { gdal_sys::OGRGeometryTypeToName(ty) };
```
lnicola, line 441: Might be worth adding a null check here, even if it's just an assert.

lnicola: Actually, `OGRwkbGeometryType::Type` is an integer, so it's easy to call this with an invalid value. I'd be inclined to return `Option<String>`.

ttencate: Result<String> then? It's more consistent with other APIs in the crate.

ttencate: https://gdal.org/api/vector_c_api.html#_CPPv421OGRGeometryTypeToName18OGRwkbGeometryType says "NULL on failure". Let's do an assertion then.

Got to run now, will do this next week (or feel free to do it yourself if you have time).

ttencate: Done now, and locked down with an extra test.

lnicola:
Oof, if you're still around, can you rebase to fix the changelog conflict?

ttencate:
Btw how would you feel about having this awkward `OGRwkbGeometryType` hammered into a proper Rust `enum` called `GeometryType`? Just a matter of someone not having done it yet, or are there objections?

ttencate:
I will run Clippy and rustfmt before pushing next time. I will run Clippy and rustfmt before pushing next time. I will run Clippy and rustfmt before pushing next time.

lnicola:
> Btw how would you feel about having this awkward OGRwkbGeometryType hammered into a proper Rust enum called GeometryType? Just a matter of someone not having done it yet, or are there objections?

Might be a good idea, but I'm not sure how extensible it's supposed to be (take a look at the [values defined](https://gdal.org/doxygen/ogr__core_8h.html#a800236a0d460ef66e687b7b65610f12a) in GDAL).

PR #257, opened 2022-02-10
Implement read_dir function and test

Barugon:
Implement a `gdal::vsi::read_dir` function and test.

---


src/vsi.rs
```diff
@@ -7,6 +7,39 @@ use gdal_sys::{VSIFCloseL, VSIFileFromMemBuffer, VSIFree, VSIGetMemFileBuffer, V
 use crate::errors::{GdalError, Result};
 use crate::utils::{_last_null_pointer_err, _path_to_c_string};
 
+/// Read the file names from a virtual file system with optional recursion.
+pub fn read_dir<P: AsRef<Path>>(path: P, recursive: bool) -> Result<Vec<String>> {
```
ChristianBeilschmidt, line 11: Just as an open question. Do we want to return a list of `String`s or of `PathBuf`s?

src/vsi.rs
```diff
@@ -7,6 +7,39 @@ use gdal_sys::{VSIFCloseL, VSIFileFromMemBuffer, VSIFree, VSIGetMemFileBuffer, V
...
+        if index != 0 {
+            loop {
+                let ptr = *(index as *mut *mut i8);
+                if ptr as usize == 0 {
+                    break;
+                }
+
+                if let Ok(file) = std::ffi::CStr::from_ptr(ptr).as_ref().to_str() {
+                    files.push(String::from(file));
+                }
```
ChristianBeilschmidt, line 33: What is the rationale for ignoring the error vs. returning the error?

src/raster/tests.rs
```diff
@@ -693,3 +693,41 @@ fn test_rasterize() {
...
+        fixture!("test_vsi_read_dir.zip").to_str().unwrap(),
+    ]
+    .concat();
+
+    // Read without recursion.
+    let expected = HashSet::from(["folder", "File 1.txt", "File 2.txt", "File 3.txt"]);
+    let files = vsi::read_dir(path.as_str(), false).unwrap();
+    for file in files {
+        assert!(expected.contains(file.as_str()));
+    }
```
ChristianBeilschmidt, line 714: You get back a `Vec<String>` so you can just `assert_eq!` once  I guess.

src/raster/tests.rs
```diff
@@ -693,3 +693,41 @@ fn test_rasterize() {
...
+        "folder/",
+        "folder/File 4.txt",
+        "File 1.txt",
+        "File 2.txt",
+        "File 3.txt",
+    ]);
+    let files = vsi::read_dir(path.as_str(), true).unwrap();
+    for file in files {
+        assert!(expected.contains(file.as_str()));
+    }
```
ChristianBeilschmidt, line 727: Same here.

ChristianBeilschmidt (changes requested):
I've got some comments.

src/vsi.rs
```diff
@@ -7,6 +7,39 @@ use gdal_sys::{VSIFCloseL, VSIFileFromMemBuffer, VSIFree, VSIGetMemFileBuffer, V
...
+
+                index += std::mem::size_of::<*mut i8>();
+            }
```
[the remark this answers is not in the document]
Barugon: I didn't know about this. I'll have a look.

src/vsi.rs
Barugon: I originally had it as PathBuf but changed it because it's easy to go from String to Path but not so easy going from PathBuf to &str. I can change it if you want.

src/raster/tests.rs
Barugon: I don't think there's any guarantee about the order.

src/vsi.rs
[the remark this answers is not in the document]
Barugon: Ok, I left it as `Vec<String>` and simplified the function to just use  `_string_array`.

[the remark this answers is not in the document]
Barugon: I added checks for null pointers from `VSIReadDirRecursive` and `VSIReadDir`.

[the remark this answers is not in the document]
ChristianBeilschmidt: Please check if you can't use `crate::utils::_last_null_pointer_err` for that.

Then, from my side, this PR is good to go, and maybe @jdroenner can allow the CI.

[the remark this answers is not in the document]
Barugon: Thanks for the review @ChristianBeilschmidt

ChristianBeilschmidt:
:+1: 

Can someone, e.g. @jdroenner, activate the CI?

Barugon:
I just realized that the test should be in `vsi.rs`. Fixing....

Barugon:
Could you please activate CI again for this branch?

lnicola:
We can't activate it directly, it's disabled for PRs in the workflow, and some convincing might be required to change that.

But at least we can

bors try

lnicola:
CC #266

src/vsi.rs
```diff
@@ -5,7 +5,29 @@ use std::path::{Path, PathBuf};
...
 use crate::errors::{GdalError, Result};
-use crate::utils::{_last_null_pointer_err, _path_to_c_string};
+use crate::utils::{_last_null_pointer_err, _path_to_c_string, _string_array};
+
+/// Read the file names from a virtual file system with optional recursion.
+pub fn read_dir<P: AsRef<Path>>(path: P, recursive: bool) -> Result<Vec<String>> {
+    let path = _path_to_c_string(path.as_ref())?;
+    let data = unsafe {
+        if recursive {
+            let data = gdal_sys::VSIReadDirRecursive(path.as_ptr());
```
lnicola, line 15: Nit:

```suggestion
            let data = unsafe { gdal_sys::VSIReadDirRecursive(path.as_ptr()) }; 
```

and the same below.

Smaller `unsafe` blocks are nicer because they make it more obvious what exactly is unsafe.

src/vsi.rs
```diff
@@ -5,7 +5,29 @@ use std::path::{Path, PathBuf};
 use gdal_sys::{VSIFCloseL, VSIFileFromMemBuffer, VSIFree, VSIGetMemFileBuffer, VSIUnlink};
 
 use crate::errors::{GdalError, Result};
-use crate::utils::{_last_null_pointer_err, _path_to_c_string};
+use crate::utils::{_last_null_pointer_err, _path_to_c_string, _string_array};
+
+/// Read the file names from a virtual file system with optional recursion.
+pub fn read_dir<P: AsRef<Path>>(path: P, recursive: bool) -> Result<Vec<String>> {
```
lnicola, line 11: I generally prefer to split these in two (one calling `path.as_ref()`, another one taking `&Path`), similarly to https://doc.rust-lang.org/src/std/fs.rs.html#947-949, but you could argue this function is small enough.

The reason is that generic functions are compiled in the caller crate, leading to longer build times and larger binaries.

lnicola:
Hmm, CI works now.

Barugon:
> Hmm, CI works now.

We'll need your change for it to not fail though.

lnicola:
Can you rebase?

Barugon:
I was having trouble rebasing, so I merged. Should be good though.

lnicola:
Oof, sorry, I missed your comment. Can you resolve the conflict, then maybe squash? It's fine if you don't, though.

Barugon:
Hopefully I did that correctly.

lnicola:
> Hopefully I did that correctly.

Almost, except the commit message :sweat_smile:.

Barugon:
:(

src/vsi.rs
lnicola: I think `PathBuf` is indeed more idiomatic here. Do we take paths as `&str` in other places?

Barugon:
> Sorry, I missed this, @ChristianBeilschmidt had a good question up there.

Which comment? AFAIK, everything was resolved.

lnicola:
The return type, which has `String` instead of `PathBuf`.

src/vsi.rs
```diff
@@ -28,7 +28,14 @@ fn _read_dir(path: &Path, recursive: bool) -> Result<Vec<String>> {
         data
     };
 
-    Ok(_string_array(data))
+    let strings = _string_array(data);
```
lnicola, line 31: We can do it in a later PR, but you could add `_pathbuf` and `_pathbuf_array` functions to avoid the reallocation here.

Barugon: It's getting late here. Let me do that tomorrow.

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
I'm sorry if I blocked this. I guess my comments were resolved.
@lnicola has a comment open and can resolve this then.

src/utils.rs
```diff
@@ -11,7 +11,23 @@ pub fn _string(raw_ptr: *const c_char) -> String {
...
+
+pub fn _pathbuf_array(raw_ptr: *mut *mut c_char) -> Vec<PathBuf> {
+    _convert_raw_ptr_array(raw_ptr, _pathbuf)
+}
+
+fn _convert_raw_ptr_array<F, R>(raw_ptr: *mut *mut c_char, convert: F) -> Vec<R>
+where
+    F: Fn(*const c_char) -> R,
+{
+    let mut ret_val: Vec<R> = vec![];
```
Barugon, line 30: Is there a reason that this couldn't just be `let mut ret_val = Vec::new();`?

lnicola: Probably not

PR #260, opened 2022-03-04
Add wrapper for OGR_L_SetFeature

geohardtke:
---

Hi, I tried to find a way to update the value of a field in a vector and couldn't find a way to achieve that. Maybe I missed something? If so, sorry. Otherwise here is a pull request that implements a wrapper for OGR_L_SetFeature. 

It can be used as follows:

```rust
        let ds_options = DatasetOptions {
            ..DatasetOptions::default()
        };
        let ds = Dataset::open_ex("three_layer_ds.s3db"), ds_options).unwrap();
        let mut layer = ds.layer(0).unwrap();
        let fids: Vec<u64> = layer.features().map(|f| f.fid().unwrap()).collect();
        let feature = layer.feature(fids[0]).unwrap();
        feature.set_field_integer("id", 1).ok();
        layer.set_feature(feature).ok();
 ```


geohardtke:
Needs https://github.com/georust/gdal/pull/262 merged in order to pass `clippy` and `fmt`

geohardtke:
Also messed up this one. Will start from scratch.

Issue #261, opened 2022-03-07
current version of the linter doesn't allow `map(..).flatten()` to be called on an `Option`


geohardtke:
Hi there,

As the tile says, the current version of the linter doesn't allow `map(..).flatten()` to be called on an `Option`. This currently breaks the CI and any new PR will fail. See for example (https://github.com/georust/gdal/runs/5443188765?check_suite_focus=true)

The solution suggested by the compiler seems to work all the tests still pass. This [PR](https://github.com/georust/gdal/pull/262) fixes the issue.

geohardtke:
unfortunately two tests (unrelated?) are now failing. 
```
    vector::vector_tests::tests::test_field_in_layer
    vector::vector_tests::tests::test_owned_layers 
```

will see if I can fix them

geohardtke:
not accurate description of the problem and messy solution in the PR. Will open new issue and PR. Sorry for the noise

PR #262, opened 2022-03-07
Fix map flatten on option

geohardtke:
---


geohardtke:
See https://github.com/georust/gdal/issues/261

CHANGES.md
```diff
@@ -1,7 +1,7 @@
 # Changes
 
 ## Unreleased
-
+- Fixed: Calling `map(..).flatten()` not allowed on an `Option` from in `rust >= 1.59`
```
lnicola, line 4: It is allowed. The error you got is from `clippy`, which is a code linter for style problems and was recently added to the CI.

It's not worth adding this to the changelog, since it's only an internal change.

lnicola:
Two test files shows up as modified, was that intentional? Also, please squash.

geohardtke:
Sorry I was in a rush and made a mess. Spent the last half an hour to make it all look nice and easy to follow, but didn't manage to get it right. I'll just close it and start a new nice and clean PR.

PR #264, opened 2022-03-08
Added wrapper for OGR_L_SetFeature

geohardtke:
---

Hi, I tried to find a way to update the value of a field in a vector and couldn't find a way to achieve that. Maybe I missed something? If so, sorry. Otherwise here is a pull request that implements a wrapper for OGR_L_SetFeature. 

It can be used as follows:

```rust
        let ds_options = DatasetOptions {
            ..DatasetOptions::default()
        };
        let ds = Dataset::open_ex("three_layer_ds.s3db"), ds_options).unwrap();
        let mut layer = ds.layer(0).unwrap();
        let fids: Vec<u64> = layer.features().map(|f| f.fid().unwrap()).collect();
        let feature = layer.feature(fids[0]).unwrap();
        feature.set_field_integer("id", 1).ok();
        layer.set_feature(feature).ok();
 ```


rmanoka:
This is a good API addition; thanks and apologies for the delay @geohardtke .   Could you pls. fix the conflicts?

rmanoka (approved):
lgtm; will merge it in a couple days unless there are more suggestions.

PR #272, opened 2022-05-12
ENH: Add CoordTransform::transform_bounds()

brendan-ward:
---

This wraps `OCTTransformBounds` to provide `CoordTransform::transform_bounds()` for GDAL >= 3.4

Output values were verified manually using [rasterio](https://github.com/rasterio/rasterio).


brendan-ward:
I'd like to add an example to `examples/spatial_reference.rs`, but it is unclear how to do so for GDAL version-specific functions; I did not see other version-specific functionality in any of the examples.  Is there a pattern for doing so, or it better to leave out such examples?

src/spatial_ref/srs.rs
```diff
@@ -33,6 +33,61 @@ impl CoordTransform {
...
+                &mut out_ymax as *mut c_double,
+                densify_pts as c_int,
+            ) == 1
+        };
+
+        if ret_val {
+            Ok([out_xmin, out_ymin, out_xmax, out_ymax])
+        } else {
+            let err = _last_cpl_err(CPLErr::CE_Failure);
+            let msg = if let GdalError::CplError { msg, .. } = err {
```
lnicola, line 74: A `match` might look nicer here (you can probably drop the `trim`), but it probably doesn't matter too much.

lnicola:
Test failure seems unrelated:

```
---- config::tests::test_error_handler stdout ----
thread 'config::tests::test_error_handler' panicked at 'assertion failed: `(left == right)`
  left: `[(Failure, 42, "foo"), (Warning, 1, "bar"), (Warning, 1, "GPKG: bad application_id=0x00000000 on '/tmp/.tmp25HDOI.gpkg'")]`,
 right: `[(Failure, 42, "foo"), (Warning, 1, "bar")]`', src/config.rs:230:9
```

src/spatial_ref/srs.rs
brendan-ward: This was copy-pasted from `transform_coords` for consistency; ideally both would get updated to be the same if `match` is more appropriate.  I can update just this one here and the other in a follow up PR, or both in a follow up PR.

I'd probably also want to invert the logic a little bit so that only the error case is in the conditional and have a bare `Ok`:

```
if !ret_val {...handle error and return... }

Ok([out_xmin, out_ymin, out_xmax, out_ymax])
```

But didn't do so in this PR out of consistency.

lnicola: >  I can update just this one here and the other in a follow up PR, or both in a follow up PR.

Right, it's fine either way.

src/spatial_ref/srs.rs
```diff
@@ -33,6 +33,61 @@ impl CoordTransform {
...
+    ///            interpreted in the axis order of the source SpatialRef,
+    ///            typically [xmin, ymin, xmax, ymax]
+    /// * densify_pts - number of points per edge (recommended: 21)
+    ///
+    /// # Returns
+    /// Some([f64; 4]) with bounds in axis order of target SpatialRef
+    /// None if there is an error.
+    #[cfg(all(major_ge_3, minor_ge_4))]
+    pub fn transform_bounds(&self, bounds: &[f64; 4], densify_pts: i32) -> Result<([f64; 4])> {
+        let mut out_xmin: f64 = 0.;
```
lnicola, line 50: You'll have to move the `use` here to avoid the warning. Or you might be able to skip the casts, like in `transform_coords`, I'm not sure.

lnicola:
Still a formatting error (try `cargo fmt`).

lnicola:
LGTM, I'll merge this tomorrow unless anyone else has objections.

PR #274, opened 2022-05-14
Ability to get owned reference to inner geometry.

### LINKED ISSUE #269: `Geometry::get_point_vec()` returns empty `Vec`
I'm unable to access get the points of any GDAL `Geometry` instance I create. The following will fail at the last test:

```rust
#[test]
fn test_get_points() -> std::result::Result<(), GdalError> {
    let sample = "/vsicurl/https://gist.githubusercontent.com/metasim/15fb05bbe76c1d7e5583033f2b6757f7/raw/fac8b385b733f50f0b7a1466db03e7918084b4a1/test.geojson";
    let ds = Dataset::open(sample)?;
    let layer = ds.layer(0)?;
    let feature = layer.feature(0).unwrap();
    let geom = feature.geometry();

    assert_eq!(geom.geometry_count(), 1);
    assert!(geom.area() > 0.);
    assert_eq!(geom.geometry_type(), gdal_sys::OGRwkbGeometryType::wkbPolygon);
    assert!(geom.json()?.contains("Polygon"));
    let points = geom.get_point_vec();
    assert!(!points.is_empty());

    Ok(())
}
```

metasim:
---

Closes #269


src/vector/geometry.rs
```diff
@@ -310,6 +312,13 @@ impl Geometry {
         Geometry::with_c_geometry(c_geom, false)
     }
 
+    /// Get a reference to the geometry at given `index`
+    pub fn get_geometry(&self, index: usize) -> Result<GeometryRef> {
+        let geom = unsafe { self.get_unowned_geometry(index) };
+        let gref = GeometryRef { geom, _lifetime: PhantomData::default() };
+        Ok(gref)
```
urschrei, line 319: Just wondering why you're returning a Result here when there doesn't seem to be any circumstance under which it could be an Err

metasim:
@urschrei Any further thoughts about this PR?

metasim:
@rmanoka FYI, Build is now completing successfully.

PR #283, opened 2022-08-09
'rasterize' should accept references to geometries

ianliu:
As of now, `gdal::raster::rasterize` is accepting `&[Geometry]`, but this is a little bit cumbersome, since `gdal::vector::Feature`'s `geometry` method returns a reference `&Geometry`.

To work around, one has to copy the geometry like so:

```rust
...
let geom = feature.geometry();
let wkb = geom.wkb().unwrap();
let copy = Geometry::from_wkb(&wkb);
rasterize(..., &[copy], ...);
```

By changing the `rasterize` function signature to accept a slice of references to geometries, this is not needed anymore.

---

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- **Breaking**: Change signature of `gdal::raster::rasterize` method to accept a slice of _references_ to Geometries.
+
+  - https://github.com/georust/gdal/pull/283
+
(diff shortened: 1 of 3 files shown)


ianliu:
Ok, now I see that `Geometry` implements `Clone`, so it is not _that_ cumbersome. Nonetheless, `rasterize` shouldn't need a copy to perform its operations, as it only needs a read-only reference to the geometry.

msalib:
Converting a reference to a single element slice can be done without copying though using [`from_ref`](https://doc.rust-lang.org/stable/std/slice/fn.from_ref.html). I think that slices of references are awkward to handle, so maybe instead of changing the interface, you might consider updating the docs to point to `from_ref` for the common single-geometry case instead?

ianliu:
Right! Didn't know about that function, I will change the docs then.

PR #284, opened 2022-08-15
prevent SIGSEGV for non-strings on read string method

ChristianBeilschmidt:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/gdal/blob/master/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

I noticed that there is a segmentation fault when calling read_as_string for the MDArray if it is not of type string.
This is why I check for its datatype first.

Moreover, there are new Clippy lints that I had to consider.

PR #285, opened 2022-08-15
add raster mask methods

jdroenner:
- [ x ] I agree to follow the project's [code of conduct](https://github.com/georust/gdal/blob/master/CODE_OF_CONDUCT.md).
- [ x ] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

### DIFF
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -50,6 +50,37 @@ fn map_resample_alg(alg: &ResampleAlg) -> u32 {
     }
 }
 
+/// Wrapper type for gdal mask flags.
+/// From the GDAL docs:
+/// - `GMF_ALL_VALID`(0x01): There are no invalid pixels, all mask values will be 255. When used this will normally be the only flag set.
+/// - `GMF_PER_DATASET`(0x02): The mask band is shared between all bands on the dataset.
+/// - `GMF_ALPHA`(0x04): The mask band is actually an alpha band and may have values other than 0 and 255.
+/// - `GMF_NODATA`(0x08): Indicates the mask is actually being generated from nodata values. (mutually exclusive of `GMF_ALPHA`)
+pub struct GdalMaskFlags(i32);
+
+impl GdalMaskFlags {
+    const GMF_ALL_VALID: i32 = 0x01;
+    const GMF_PER_DATASET: i32 = 0x02;
+    const GMF_ALPHA: i32 = 0x04;
+    const GMF_NODATA: i32 = 0x08;
+
+    pub fn is_all_valid(&self) -> bool {
+        self.0 & Self::GMF_ALL_VALID != 0
+    }
+
+    pub fn is_per_dataset(&self) -> bool {
+        self.0 & Self::GMF_PER_DATASET != 0
+    }
+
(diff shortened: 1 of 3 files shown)


ChristianBeilschmidt:
Maybe we can create a separate Clippy PR based on https://github.com/georust/gdal/pull/284

jdroenner:
yes we need to solve that issue anyway

ChristianBeilschmidt:
https://github.com/georust/gdal/pull/286

src/raster/rasterband.rs
```diff
@@ -50,6 +50,37 @@ fn map_resample_alg(alg: &ResampleAlg) -> u32 {
...
+/// Wrapper type for gdal mask flags.
+/// From the GDAL docs:
+/// - `GMF_ALL_VALID`(0x01): There are no invalid pixels, all mask values will be 255. When used this will normally be the only flag set.
+/// - `GMF_PER_DATASET`(0x02): The mask band is shared between all bands on the dataset.
+/// - `GMF_ALPHA`(0x04): The mask band is actually an alpha band and may have values other than 0 and 255.
+/// - `GMF_NODATA`(0x08): Indicates the mask is actually being generated from nodata values. (mutually exclusive of `GMF_ALPHA`)
+pub struct GdalMaskFlags(i32);
+
+impl GdalMaskFlags {
+    const GMF_ALL_VALID: i32 = 0x01;
```
lnicola, line 62: I'm a bit surprised we don't have these in `gdal-sys`.

src/raster/rasterband.rs
```diff
@@ -498,6 +529,43 @@ impl<'a> RasterBand<'a> {
...
+    /// `shared_between_all_bands` indicates if all bands of the dataset use the same mask.
+    pub fn create_mask_band(&mut self, shared_between_all_bands: bool) -> Result<()> {
+        let flags = if shared_between_all_bands {
+            GdalMaskFlags::GMF_PER_DATASET // It is the only valid flag here.
+        } else {
+            0x00
+        };
+
+        unsafe {
+            let rv = gdal_sys::GDALCreateMaskBand(self.c_rasterband, flags);
```
lnicola, line 550:
```suggestion
            let rv = unsafe { gdal_sys::GDALCreateMaskBand(self.c_rasterband, flags) };
```

to make the `unsafe` block as small as possible.

lnicola (approved):

src/raster/rasterband.rs
jdroenner: i think it is a define not a const variable in the cpp code. I guess bindgen does not pick that up.

lnicola: Yeah, I'm not sure what's up, I think it should generate them.

jdroenner: is bindgen able to handle defines? i can't find any examples.

lnicola: Yeah, it is: https://github.com/rust-lang/rust-bindgen/issues/1185

jdroenner: so what is wrong with our bindgen config? And i guess we would need to regenerate all bindings -_-

PR #289, opened 2022-08-24
wrapper for gdalmdimtranslate

ChristianBeilschmidt:
---

I created a wrapper around `GDALMultiDimTranslate`. I used part of the logic for the wrapper for the VRT program.

### DIFF
--- a/src/programs/raster/mdimtranslate.rs
+++ b/src/programs/raster/mdimtranslate.rs
@@ -18,11 +18,11 @@ use std::{
 ///
 /// [GDALMultiDimTranslateOptions]: https://gdal.org/api/gdal_utils.html#_CPPv428GDALMultiDimTranslateOptions
 ///
-pub struct MultiDimTranslateOption {
+pub struct MultiDimTranslateOptions {
     c_options: *mut GDALMultiDimTranslateOptions,
 }
 
-impl MultiDimTranslateOption {
+impl MultiDimTranslateOptions {
     /// See [GDALMultiDimTranslateOptionsNew].
     ///
     /// [GDALMultiDimTranslateOptionsNew]: https://gdal.org/api/gdal_utils.html#_CPPv431GDALMultiDimTranslateOptionsNewPPcP37GDALMultiDimTranslateOptionsForBinary
@@ -65,7 +65,7 @@ impl MultiDimTranslateOption {
     }
 }
 
-impl Drop for MultiDimTranslateOption {
+impl Drop for MultiDimTranslateOptions {
     fn drop(&mut self) {
         unsafe {
             gdal_sys::GDALMultiDimTranslateOptionsFree(self.c_options);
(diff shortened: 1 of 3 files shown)


src/programs/raster/mdimtranslate.rs
```diff
@@ -0,0 +1,266 @@
...
+        unsafe {
+            gdal_sys::GDALMultiDimTranslateOptionsFree(self.c_options);
+        }
+    }
+}
+
+impl TryFrom<Vec<&str>> for MultiDimTranslateOption {
+    type Error = GdalError;
+
+    fn try_from(value: Vec<&str>) -> std::result::Result<Self, Self::Error> {
```
lnicola, line 79: Does this work?

```suggestion
    fn try_from(value: Vec<&str>) -> Result<Self> {
```

src/programs/raster/mdimtranslate.rs
```diff
@@ -0,0 +1,266 @@
...
+    mem::ManuallyDrop,
+    path::Path,
+    ptr::{null, null_mut},
+};
+
+/// Wraps a [GDALMultiDimTranslateOptions] object.
+///
+/// [GDALMultiDimTranslateOptions]: https://gdal.org/api/gdal_utils.html#_CPPv428GDALMultiDimTranslateOptions
+///
+pub struct MultiDimTranslateOption {
```
lnicola, line 21: Should probably be called `MultiDimTranslateOptions`, since there can be multiple options inside.

ChristianBeilschmidt: [23957df](https://github.com/georust/gdal/pull/289/commits/23957df6195fc5557422ffd81aa724c7aa4aaaf1)

src/programs/raster/mdimtranslate.rs
ChristianBeilschmidt: [23957df](https://github.com/georust/gdal/pull/289/commits/23957df6195fc5557422ffd81aa724c7aa4aaaf1)

lnicola (approved):

ChristianBeilschmidt:
I forgot to replace the `Cell<bool>` with a plain boolean, I guess it can be this simple :see_no_evil: .

lnicola:
```
thread 'config::tests::test_error_handler' panicked at 'assertion failed: `(left == right)`
  left: `[(Warning, 1, "GPKG: bad application_id=0x00000000 on '/tmp/.tmpPTiAzK.gpkg'"), (Failure, 42, "foo"), (Warning, 1, "bar")]`,
 right: `[(Failure, 42, "foo"), (Warning, 1, "bar")]`', src/config.rs:230:9
```

Is this some sort of new GDAL warning?

ChristianBeilschmidt:
> ```
> thread 'config::tests::test_error_handler' panicked at 'assertion failed: `(left == right)`
>   left: `[(Warning, 1, "GPKG: bad application_id=0x00000000 on '/tmp/.tmpPTiAzK.gpkg'"), (Failure, 42, "foo"), (Warning, 1, "bar")]`,
>  right: `[(Failure, 42, "foo"), (Warning, 1, "bar")]`', src/config.rs:230:9
> ```
> 
> Is this some sort of new GDAL warning?

Since it appeared in the LTS-CI, I guess not.

What could have happened:
The tests run in parallel. Some test sets the global error handler. Then some other test fails… which lands in the error handler. And this test gets some strange results.

I'm not sure, but we could get races on all the global stuff.
Maybe we need some global Mutexes on some test stuff that must not interfere with something else :shrug:.

What do you think?

jdroenner:
bors merge

lnicola:
I'm a bit worried about the test failure here.

jdroenner:
i guess the problem is not really related to this PR but to gdal and cargo test in general. Cargo runs all the tests in parallel and when two tests fail at the same time we have a problem with global state.

lnicola:
I think we can move them to an integration test.

PR #295, opened 2022-08-31
Enabled non-consuming fallible conversion from GDAL to geo-types Geometry

### LINKED ISSUE #276: Exposing `gdal_to_geo`
Is there any reason why `gdal_to_geo` can't be made public? Either through the `vector` module, or through some other "utils" module?

https://github.com/georust/gdal/blob/fb6e260d46409cb5842c7c2f4978bb5d7bb2ebb0/src/vector/mod.rs#L21

metasim:
Also added `Geometry::to_geo` method, for symmetry with `ToGdal::to_gdal`.

---

Closes #276


src/vector/gdal_to_geo.rs
```diff
@@ -110,3 +110,10 @@ impl TryFrom<Geometry> for geo_types::Geometry<f64> {
         }
     }
 }
+
+impl TryFrom<Geometry> for geo_types::Geometry<f64> {
+    type Error = GdalError;
+    fn try_from(value: Geometry) -> Result<Self, Self::Error> {
+        TryFrom::<&Geometry>::try_from(&value)
```
lnicola, line 117:
```suggestion
        Self::try_from(&value)
```

lnicola: Is my suggestion somehow wrong?

I'm a bit confused, you :+1:ed it, marked it as resolved, but didn't change the code.

metasim: @lnicola I forgot to `push` 🤦 . I'm recovering from _The Covid_, and not only forgot, but did them in the wrong order.  My apologies.

lnicola: No worries, and good luck with your recovery!

Issue #268, opened 2022-04-18
docs.rs build is failing for 0.12.0


metasim:
<img width="500" alt="image" src="https://user-images.githubusercontent.com/131013/163739362-b174651f-b3da-40b3-a504-592b52bcc3fa.png">

Looks like some sort of linker error:

```
/opt/rustwide/workdir/build.rs:13: undefined reference to `GDALVersionInfo'
```

Build log:

https://docs.rs/crate/gdal/0.12.0/builds/494924

metasim:
Perhaps `DOCS_RS` is no longer defined? 🤷 

https://github.com/georust/gdal/blob/492e0c8e3f14c26e0e3442e42a04c97c14ff627a/gdal-sys/build.rs#L63-L66

rmanoka:
related to #240 ?

metasim:
Closed by #281

PR #296, opened 2022-09-01
Increased documentation content of README and root rustdoc page.

metasim:
---

I wanted to kickstart some work to level-up the documentation. Wanting to take an incremental approach, making it feel a bit less intimidating to make progress in the future. I don't see this as the only PR I do on this, just a start. All feedback welcome :). 

Related:
* #268
* #281


Readme.md
```diff
@@ -1,18 +1,44 @@
...
+
+Building this crate assumes a compatible version of GDAL is installed with requisite header files and shared libraries. This repository includes pre-generated bindings for GDAL 2.4 through 3.5 (see the`gdal-sys/prebuilt-bindings` directory). If you're compiling against a later version of GDAL, you can enable the `bindgen` feature flag to have new bindings generated on the fly.  
+
+## Community
+
+This crate is part of the expansive (and expanding!) [`georust`](https://georust.org/) organization. Come join our discussions on [Discord](https://discord.gg/Fp2aape)!
+
+## Contributing
+
+This crate continues to evolve, and PRs are welcome. Make sure you are comfortable with the [Code of Conduct](CODE_OF_CONDUCT.md) and [License](LICENSE.txt) and before submitting a PR.
```
michaelkirk, line 40:
```suggestion
This crate continues to evolve, and PRs are welcome. Make sure you are comfortable with the [Code of Conduct](CODE_OF_CONDUCT.md) and [License](LICENSE.txt) before submitting a PR.
```

src/lib.rs
```diff
@@ -1,25 +1,165 @@
...
+//! use gdal::{Dataset, Metadata};
+//! // The `LayerAccess` trait enables reading of vector specific fields from the `Dataset`.
+//! use gdal::vector::LayerAccess;
+//! # fn main() -> gdal::errors::Result<()> {
+//! use gdal::errors::GdalError;
+//! use gdal::vector::geometry_type_to_name;
+//! let dataset = Dataset::open("fixtures/roads.geojson")?;
+//! println!("Dataset description: {}", dataset.description()?);
+//! let layer_count = dataset.layer_count();
+//! println!("Number of layers: {layer_count}" );
```
michaelkirk, line 104:
```suggestion
//! println!("Number of layers: {layer_count}");
```

michaelkirk (commented):
This seems like a great improvement! I'm not a frequent user of this crate though, so I'll wait for someone else to weigh in before merging.

jdroenner:
This is really a great improvement! I agree with the incremental approach for better docs. So we should merge this PR and then add more docs later.

src/lib.rs
```diff
@@ -30,6 +170,7 @@ mod driver;
 pub mod errors;
 mod gdal_major_object;
 mod metadata;
+#[doc(hidden)]
```
ChristianBeilschmidt, line 173: Why do we need to hide this?

Readme.md
```diff
@@ -1,18 +1,44 @@
...
-* read and write raster data
-* warp between datasets
-* read and write vector data
-* access metadata
+## Capabilities
 
-Many raster and vector functions are not available. Patches welcome :)
+GDAL is an incredibly powerful library. For a general understanding of its capabilities, a good place to get started is the [GDAL User-oriented documentation](https://gdal.org/user/index.html). These features include:
+
+* Open a raster dataset for reading/writing
```
lnicola, line 14: Nit: I think the continuous form sounds better here and below:

```suggestion
* Opening a raster dataset for reading/writing
```

Readme.md
```diff
@@ -1,18 +1,44 @@
...
+* Get the raster size and number of bands
+* Get and set the projection and geo-transform
+* Warp (resample and reproject) between datasets
+* Read and write vector data
+* Read and write metadata
+* Translate between formats
+
+## Documentation
+
+This crate's [API documentation](https://docs.rs/crate/gdal) is hosted on [docs.rs](https://docs.rs). The Rust documentation is currently a work in progress, and may not cover requisite details on parameter semantics, value interpretation, etc.. Therefore, the authoritative documentation for the [GDAL C API](https://gdal.org/api/index.html#c-api) may be a helpful reference.
```
lnicola, line 24: It might be worth linking to the C++ API too, since it's matches the Rust one better and usually has more details.

Readme.md
```diff
@@ -1,18 +1,44 @@
...
+
+## Usage
+
+This crate provides idiomatic Rust bindings for the GDAL C library. To do that, it internally uses [`gdal-sys`](gdal-sys), which is a  [`bindgen`](https://rust-lang.github.io/rust-bindgen/)-generated sub-crate, rendered directly from the GDAL C library. Normally one would not need to interact with the `gdal-sys` crate. However, not every single feature of the GDAL C API has been fully exposed, so in some cases the `gdal-sys` crate may be necessary to accessing more esoteric features of GDAL.
```
lnicola, line 28:
```suggestion
This crate provides high-level, idiomatic Rust bindings for GDAL.
To do that, it internally uses [`gdal-sys`](gdal-sys), a  low-level interface to the GDAL C library, generated using [`bindgen`](https://rust-lang.github.io/rust-bindgen/).
Using the `gdal-sys` crate is normally not needed, but it can be useful in order to call APIs that have not yet been exposed in `gdal`.
```

Readme.md
```diff
@@ -1,18 +1,44 @@
...
+
+> _Note_: GDAL offers both C and C++ APIs. This crate only makes use of the GDAL C API.
+
+Building this crate assumes a compatible version of GDAL is installed with requisite header files and shared libraries. This repository includes pre-generated bindings for GDAL 2.4 through 3.5 (see the`gdal-sys/prebuilt-bindings` directory). If you're compiling against a later version of GDAL, you can enable the `bindgen` feature flag to have new bindings generated on the fly.
```
lnicola, line 32:
```suggestion
Building this crate assumes a compatible version of GDAL is installed with the corresponding header files and shared libraries.
This repository includes pre-generated bindings for GDAL 2.4 through 3.5 (see the`gdal-sys/prebuilt-bindings` directory).
If you're compiling against a later version of GDAL, you can enable the `bindgen` feature flag to have new bindings generated on the fly.
```

src/lib.rs
```diff
@@ -1,25 +1,165 @@
...
-//! for feature in layer.features() {
-//!     let highway_field = feature.field("highway").unwrap().unwrap();
-//!     let geometry = feature.geometry();
-//!     println!("{} {}", highway_field.into_string().unwrap(), geometry.wkt().unwrap());
+//! ### Vector
+//!
+//! > _Note_: Describing the sophisticated structure of GDAL vector data is outside the scope of this example, so
+//! > please see [Vector Data Model](https://gdal.org/user/vector_data_model.html) section of the GDAL user manual for details.
+//!
+//! ```rust
```
lnicola, line 94: Maybe add a couple of newlines to spread this out a little and make it nicer to read?

lnicola, line 94:
```suggestion
//! ```rust, no_run
```

src/lib.rs
metasim: Oops. Left over from an aborted experiment with documenting reexports.

Readme.md
metasim: @lnicola I made this point about C vs C++ because when I was new to the crate, I found it confusing on where to go for semantic details, especially when some of the names are the same or similar. I assumed we only link to C because that's what `bindgen` provides for us, and then there's larger complexities with name-mangling, memory model, libraries, etc. in C++. 

Were you thinking I should strike this section, or refine it, or was this just a general point?

lnicola: Of course, we're calling the C API, but ours is more similar in spirit with the C++ one. Anyway, my point was about the documentation: the C GDAL docs are pretty bare, while the C++ ones have more details.

Compare https://gdal.org/doxygen/gdal_8h.html#a764cb93a266ada902b13c2bdad3f7037 with https://gdal.org/doxygen/classGDALDataset.html#ad2a643880bf8b3c492783e1ab5fbc667.

So I was thinking we could put it more like:

```
Therefore, the authoritative documentation is that of GDAL in the form of its [C](https://gdal.org/api/index.html#c-api) and [C++](https://gdal.org/api/index.html#id3) APIs.
The latter is usually better documented.
```

metasim: I think I misread your comment... more clear now that I came back to it. My point is somewhat relevant still; when I was just coming to this crate, I would try to translate from the C++ docs to Rust, and hit a wall because we don't implement the C++ functionality, and didn't to save folks from that pothole.

lnicola: > because we don't implement the C++ functionality

Sure, but our API doesn't look like the C one either :smile:.

lnicola (approved):

lnicola:
Pro tip: when writing Markdown, it's better to keep it at one sentence/phrase per line. This makes it easier to view diffs.

CHANGES.md
```diff
@@ -101,6 +105,7 @@
 - Added a workaround in multi-dim tests to not access files multiple times
 
   - <https://github.com/georust/gdal/pull/302>
+>>>>>>> 604c41fbf62b96d0b44a14606b6f83b37db8059e
```
lnicola, line 108: Missed a conflict marker here.

README.md
```diff
@@ -0,0 +1,49 @@
...
+This crate's [API documentation](https://docs.rs/crate/gdal) is hosted on [docs.rs](https://docs.rs). 
+
+The Rust documentation is currently a work in progress, and may not cover requisite details on parameter semantics, value interpretation, etc. 
+Therefore, the authoritative documentation is that of GDAL in the form of its [C](https://gdal.org/api/index.html#c-api) and [C++](https://gdal.org/api/index.html#id3) APIs.
+The former is technically what this crate calls, but the latter is usually more clear and better documented.
+
+## Usage
+
+This crate provides high-level, idiomatic Rust bindings for GDAL.
+To do that, it internally uses [`gdal-sys`](gdal-sys), a  low-level interface to the GDAL C library, generated using [`bindgen`](https://rust-lang.github.io/rust-bindgen/).
```
lnicola, line 32:
```suggestion
To do that, it uses [`gdal-sys`](gdal-sys) internally, a low-level interface to the GDAL C library, which is generated using [`bindgen`](https://rust-lang.github.io/rust-bindgen/).
```

CHANGES.md
lnicola, line 108:
```suggestion
```

metasim: 🤦 so sorry!

PR #300, opened 2022-09-02
Require matching data types in MDArray::read_into_slice

### LINKED ISSUE #290: MDArray: Different read type compared to array type leads to ASAN issues
From https://github.com/georust/gdal/pull/284:

@lnicola:
> Try this under ASAN:
>
>             let string_array = root_group
>                .open_md_array("uint_var", CslStringList::new())
>                .unwrap();
>
>             let values = string_array.read_as::<u8>(vec![0, 0], vec![1, 2]).unwrap();
>
> This happens because GDALMDArrayRead takes the buffer type as argument, but we're passing GDALMDArrayGetDataType(self.c_mdarray), which is the array type. Normally, GDAL does some format conversions when needed. When we call GDALRasterIOEx, we pass T::gdal_type() instead.

Maybe we can also make sure that the read type aligns with the actual requested and returned type.
However, we should keep in mind that we can get strings and complex types as return values.

lnicola:
---

We don't support `GEDTC_COMPOUND` anyway because of the `GdalType` bound. ~~This is stricter than it could be, but should be fine as a quick workaround for #290. We should pass in the data type of the buffer, but it wasn't obvious to me how to do that.~~

Fixes #290

### DIFF
--- a/src/raster/mdarray.rs
+++ b/src/raster/mdarray.rs
@@ -9,14 +9,15 @@ use gdal_sys::{
     GDALAttributeReadAsIntArray, GDALAttributeReadAsString, GDALAttributeReadAsStringArray,
     GDALAttributeRelease, GDALDataType, GDALDatasetH, GDALDimensionGetIndexingVariable,
     GDALDimensionGetName, GDALDimensionGetSize, GDALDimensionHS, GDALDimensionRelease,
-    GDALExtendedDataTypeClass, GDALExtendedDataTypeGetClass, GDALExtendedDataTypeGetName,
-    GDALExtendedDataTypeGetNumericDataType, GDALExtendedDataTypeH, GDALExtendedDataTypeRelease,
-    GDALGroupGetAttribute, GDALGroupGetDimensions, GDALGroupGetGroupNames,
-    GDALGroupGetMDArrayNames, GDALGroupGetName, GDALGroupH, GDALGroupOpenGroup,
-    GDALGroupOpenMDArray, GDALGroupRelease, GDALMDArrayGetAttribute, GDALMDArrayGetDataType,
-    GDALMDArrayGetDimensionCount, GDALMDArrayGetDimensions, GDALMDArrayGetNoDataValueAsDouble,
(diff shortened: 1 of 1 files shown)


src/raster/mdarray.rs
```diff
@@ -162,6 +162,16 @@ impl<'a> MDArray<'a> {
...
 
+            // HACK: this is too strict, but see https://github.com/georust/gdal/issues/290
+            if !self.datatype().class().is_numeric()
+                || self.datatype().numeric_datatype() != T::gdal_type()
+            {
+                return Err(GdalError::UnsupportedMdDataType {
+                    data_type: self.datatype().class(),
+                    method_name: "GDALMDArrayRead",
+                });
+            }
```
ChristianBeilschmidt, line 173: Maybe we can have the call of `datatype` only once.
Now, you have one C call and two fn-calls that wrap the C call.
(we could move this out of the unsafe block, then)

Moreover, can you elaborate why this is too strict?

ChristianBeilschmidt (changes requested):

src/raster/mdarray.rs
lnicola: It's too strict if it works like the legacy API. There you could read an `u16` band into an `u8` array, or the other way around, and GDAL would do a reasonable conversion (by saturating the values). This implementation prevents that.

The solution would be to pass the buffer data type, but I don't know how to get a `GDALExtendedDataTypeH` from a `GDALDataType`.

ChristianBeilschmidt (approved):

ChristianBeilschmidt:
Looks more robust now.
If this resolves the ASAN issue https://github.com/georust/gdal/issues/290, I would like to merge it.

jdroenner:
bors merge

lnicola:
> If this resolves the ASAN issue https://github.com/georust/gdal/issues/290

It does, I included that part in `test_read_string_array`.

It actually didn't need ASAN to crash, it was a blatant buffer overrun.

PR #304, opened 2022-09-02
Enabled docs.rs feature gate flag rendering.

### LINKED ISSUE #228: docs.rs should include `array` feature
It would be nice if we enabled at least the `array` feature for docs.rs by adding this to `Cargo.toml`:
```
[package.metadata.docs.rs]
features = ["array"]
```
Perhaps adding `datetime` to that list would be good as well.

The goal here is that someone looking at docs.rs should know that the array methods actually exist because otherwise, they're very easy to miss.

metasim:
---

Followed approach outlined [here](https://gist.github.com/jam1garner/7d909e1d0e5986b9b8d50830314bec63).

Rendering can be tested locally with:

    RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --features array --open

Not sure how to test end-to-end (as generated by docs.rs) from a PR.

Closes #228

Sample:

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,3 +35,10 @@ trybuild = "1.0"
 
 [workspace]
 members = ["gdal-sys"]
+
+# docs.rs-specific configuration
+[package.metadata.docs.rs]
+# include `array` feature in documentation
+features = ["array"]
+# define attribute `docsrs` for feature badges
+rustdoc-args = ["--cfg", "docsrs"]
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -16,6 +16,7 @@ pub enum GdalError {
     #[error("StrUtf8Error")]
     StrUtf8Error(#[from] std::str::Utf8Error),
     #[cfg(feature = "ndarray")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "array")))]
     #[error("NdarrayShapeError")]
     NdarrayShapeError(#[from] ndarray::ShapeError),
     #[error("CPL error class: '{class:?}', error number: '{number}', error msg: '{msg}'")]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -171,6 +171,9 @@
 //! ...
 //! ```
 
+// Enable `doc_cfg` features when `docsrs` is defined by docs.rs config
+#![cfg_attr(docsrs, feature(doc_cfg))]
+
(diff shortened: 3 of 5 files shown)


michaelkirk:
I started using `#![cfg_attr(docsrs, feature(doc_auto_cfg))]` [in WKT](https://github.com/georust/wkt/pull/98) which seems to do the trick, and be a little less tedious/error prone.

Cargo.toml
```diff
@@ -35,3 +35,10 @@ trybuild = "1.0"
 
 [workspace]
 members = ["gdal-sys"]
+
+# docs.rs-specific configuration
+[package.metadata.docs.rs]
+# include `array` feature in documentation
+features = ["array"]
+# define attribute `docsrs` for feature badges
+rustdoc-args = ["--cfg", "docsrs"]
```
lnicola, line 44: We actually had this before, but it got dropped in #281 (which should probably be reverted at some point).

lnicola (approved):

rmanoka:
@lnicola Note that even though docs.rs now supports the `rustc-args` configuration, this still means that every dependent crate must include this configuration in their Cargo.toml or that crate's docs won't build.  With the dynamic code-gen, we are a bit more accomodating, so we should not revert that PR IMO.

lnicola:
You're right, it's too bad docs.rs doesn't have a solution for this.

rmanoka:
Guess it would be great if they could always set `cfg(docsrs)` as a sort of a standard signal.

metasim:
This is the first time I've worked with this feature, so some of the nuances are going over my head. Based on what's said above, do we need to pause this PR for a while until the "every dependent crate must include this configuration" issue is resolved (so as to be good citizens)?

As a secondary question, I read about the approach @michaelkirk uses in WKT, but wasn't sure if that would have other consequences, but it certainly seems better. Since WKT is very much a foundational library, should a) we go that way, and b) does it also suffer from the "transitive config" issue?

lnicola:
I don't think this affects the other crates? It only enables the `ndarray` feature, which I'd expect to work pretty much anywhere.

The previous problem was bigger, since you actually needed GDAL to build the docs.

rmanoka:
Agreed; I was trying to explain why #281 should not be reverted.  This PR does not affect the dependent crates' docs, so we're good.

PR #305, opened 2022-09-03
Added a more ergonomic means of accessing GDAL version properties.

metasim:
---

This PR is of minor impact, but does provide some quality-of-life benefits when building systems which need to dynamically check and verify the version of GDAL available at runtime. 

GDAL's native query mechanism uses magical tokens that requires either brute memory, or repeated digging through API docs to fetch the particular version property needed. This PR aims to provide a slightly more principled approach to querying this information via a new `enum` called `VersionInfo`. It does not change the existing mechanism, only provides an alternative.


lnicola:
I know that you wanted to make debug-printing to work, but it seems quite convoluted for what it does. I'd be fine with a more lightweight like:

```rust
pub struct VersionInfo(pub &'static str);

impl VersionInfo {
    const RELEASE_NAME: VersionInfo = VersionInfo("RELEASE_NAME");
    // ...

    fn get_version_info(&self) -> String { todo!() }

    fn all() -> String { todo!() }
}
```

Let's see how the others feel about it.

metasim:
@lnicola I turned down the "clever" volume and implemented a more direct approach. What do you think?

lnicola (approved):

lnicola:
Sure, looks good. r? @jdroenner

metasim:
bors r=lnicola

PR #308, opened 2022-09-07
Added `Rasterband::set_no_data` accepting `Option<f64>`.

### LINKED ISSUE #275: Provide ability to delete `no_data` value.
### Issue

There's no way to delete a `no_data` value in a `RasterBand`. 

### Commentary

Symmetry in the read vs. write is confusing. i.e. you can determine if there's no `no_data` value thanks to the return type `Option<f64>`:

https://github.com/georust/gdal/blob/fb6e260d46409cb5842c7c2f4978bb5d7bb2ebb0/src/raster/rasterband.rs#L389

But the write method only accepts `f64`:

https://github.com/georust/gdal/blob/fb6e260d46409cb5842c7c2f4978bb5d7bb2ebb0/src/raster/rasterband.rs#L400

Changing this parameter to `Option<f64>` would be a breaking change, but would improve the API. The other option is to have a `fn delete_no_data(&mut self) -> Result<()>` method which would be non-breaking, but less idiomatic on the Rust side.

metasim:
---

Closes #275

### DIFF
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -443,15 +443,13 @@ impl<'a> RasterBand<'a> {
     pub fn set_no_data(&mut self, no_data: Option<f64>) -> Result<()> {
         let rv = if let Some(no_data) = no_data {
             unsafe { gdal_sys::GDALSetRasterNoDataValue(self.c_rasterband, no_data) }
-        }
-        else {
+        } else {
             unsafe { gdal_sys::GDALDeleteRasterNoDataValue(self.c_rasterband) }
         };
 
         if rv != CPLErr::CE_None {
             Err(_last_cpl_err(rv))
-        }
-        else {
+        } else {
             Ok(())
         }
     }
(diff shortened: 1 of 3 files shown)


src/raster/rasterband.rs
```diff
@@ -431,12 +431,27 @@ impl<'a> RasterBand<'a> {
     }
 
     /// Set the no data value of this band.
+    ///
+    /// See [`set_no_data`](Self::set_no_data) for similar function
+    /// where _no-data_ can also be deleted.
     pub fn set_no_data_value(&mut self, no_data: f64) -> Result<()> {
-        let rv = unsafe { gdal_sys::GDALSetRasterNoDataValue(self.c_rasterband, no_data) };
+        self.set_no_data(Some(no_data))
+    }
```
metasim, line 439: If backward compatibility weren't an option, I would have changed this signature of this to accept `Option<f64>` over creating a new method and delegating to it. Should we consider a breaking change here?

metasim:
@lnicola FYI: Went with breaking change approach.

PR #309, opened 2022-09-09
Made mapping between `ResampleAlg` and `GDALRIOResampleAlg` more direct.

metasim:
Before the private function `map_resample_alg` was the only way to get the expected GDAL enumeration value.

---

Before this change, you had to copy `map_resample_alg` (because it was private) when calling into `gdal-sys` functions needing the resampling algorithm (e.g. `GDALAutoCreateWarpedVRT`). Instead of making `pub fn map_resample_alg ...`, took the approach that more tightly binds the C-side types to the Rust enum. Makes the association more clear, and less error prone in the future.

### DIFF
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -5,8 +5,8 @@ use crate::raster::{GDALDataType, GdalType};
 use crate::utils::{_last_cpl_err, _last_null_pointer_err, _string};
 use gdal_sys::{
     self, CPLErr, GDALColorEntry, GDALColorInterp, GDALColorTableH, GDALComputeRasterMinMax,
-    GDALGetRasterStatistics, GDALMajorObjectH, GDALPaletteInterp, GDALRWFlag, GDALRasterBandH,
-    GDALRasterIOExtraArg,
+    GDALGetRasterStatistics, GDALMajorObjectH, GDALPaletteInterp, GDALRIOResampleAlg, GDALRWFlag,
+    GDALRasterBandH, GDALRasterIOExtraArg,
 };
 use libc::c_int;
 use std::ffi::CString;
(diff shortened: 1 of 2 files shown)


src/raster/rasterband.rs
```diff
@@ -17,37 +17,49 @@ use ndarray::Array2;
...
+    Average = GDALRIOResampleAlg::GRIORA_Average,
     /// Mode (selects the value which appears most often of all the sampled points)
-    Mode,
+    Mode = GDALRIOResampleAlg::GRIORA_Mode,
     /// Gauss blurring
-    Gauss,
+    Gauss = GDALRIOResampleAlg::GRIORA_Gauss,
 }
 
-fn map_resample_alg(alg: &ResampleAlg) -> u32 {
```
metasim, line 41: Note: this was not private before, so there was no easy way to map values when calling into lower level `gdal-sys` functions.

src/raster/rasterband.rs
```diff
@@ -17,37 +17,49 @@ use ndarray::Array2;
...
+/// let stats = band1.get_statistics(true, false)?.unwrap();
+/// // Down-sample a image using cubic-spline interpolation
+/// let buf = band1.read_as::<f64>((0, 0), ds.raster_size(), (2, 2), Some(ResampleAlg::CubicSpline))?;
+/// // In this particular image, resulting data should be close to the overall average.
+/// assert!(buf.data.iter().all(|c| (c - stats.mean).abs() < stats.std_dev / 2.0));
+/// # Ok(())
+/// # }
+/// ```
 #[derive(Debug, Copy, Clone)]
+#[repr(u32)]
```
metasim, line 39: Not entirely sure about this value, or even if it should be specified.
`GDALRIOResampleAlg::Type` is `libc::c_uint`, which is `u32` on *nix. The [repr options](https://rust-lang.github.io/unsafe-code-guidelines/layout/enums.html#repr-annotations-accepted-on-enums) maybe suggest `#[repr(C)]` should be used instead, and the right value will be selected based on the discriminant values? Advice appreicated.

src/raster/rasterband.rs
metasim: Also, the return type should have been `GDALRIOResampleAlg::Type`.

src/raster/rasterband.rs
```diff
@@ -17,37 +17,49 @@ use ndarray::Array2;
...
+/// let ds = Dataset::open("fixtures/tinymarble.tif")?;
+/// let band1 = ds.rasterband(1)?;
+/// let stats = band1.get_statistics(true, false)?.unwrap();
+/// // Down-sample a image using cubic-spline interpolation
+/// let buf = band1.read_as::<f64>((0, 0), ds.raster_size(), (2, 2), Some(ResampleAlg::CubicSpline))?;
+/// // In this particular image, resulting data should be close to the overall average.
+/// assert!(buf.data.iter().all(|c| (c - stats.mean).abs() < stats.std_dev / 2.0));
+/// # Ok(())
+/// # }
+/// ```
```
metasim, line 37: Bonus documentation while I was in there :)

src/raster/rasterband.rs
lnicola: I don't think we need the `#[repr]` because of the cast in `to_gdal`.

lnicola (approved):

PR #311, opened 2022-09-20
Quality-of-life additions to `CplStringList`.

metasim:
---

New methods:
* `len`
* `is_empty`

New traits: 
* `Debug`
* `Iterator`

New tests:
* `has_length`
* `can_be_empty`
* `invalid_keys`
* `basic_list`
* `debug_fmt`
* `has_iterator`
 

Also tested under:

    ASAN_OPTIONS=detect_stack_use_after_return=1,sanitize-address-use-after-return=always RUSTFLAGS=-Zsanitizer=address cargo +nightly test cpl --target aarch64-apple-darwin

### DIFF
--- a/src/cpl.rs
+++ b/src/cpl.rs
@@ -73,7 +73,7 @@ impl CslStringList {
 
     /// Determine the number of entries in the list.
     pub fn len(&self) -> isize {
-        unsafe { CSLCount(self.as_ptr()) as isize }
+        (unsafe { CSLCount(self.as_ptr()) }) as isize
     }
 
     /// Determine if the list has any values
@@ -86,7 +86,7 @@ impl CslStringList {
         CslStringListIterator::new(self)
     }
 
-    /// Get the raw pointer behind list's data.
+    /// Get the raw pointer to the underlying data.
     pub fn as_ptr(&self) -> gdal_sys::CSLConstList {
         self.list_ptr
     }
--- a/src/cpl.rs
+++ b/src/cpl.rs
@@ -7,7 +7,7 @@ use std::ffi::CString;
 use std::fmt::{Debug, Formatter};
 use std::ptr;
 
-use gdal_sys::{CSLCount, CSLDestroy, CSLFetchNameValue, CSLGetField, CSLSetNameValue};
+use gdal_sys::{CSLCount, CSLDestroy, CSLFetchNameValue, CSLSetNameValue};
 use libc::c_char;
 
 use crate::errors::{GdalError, Result};
@@ -72,13 +72,13 @@ impl CslStringList {
     }
 
     /// Determine the number of entries in the list.
-    pub fn len(&self) -> isize {
-        (unsafe { CSLCount(self.as_ptr()) }) as isize
+    pub fn len(&self) -> usize {
+        (unsafe { CSLCount(self.as_ptr()) }) as usize
     }
 
     /// Determine if the list has any values
     pub fn is_empty(&self) -> bool {
-        self.len() <= 0
+        self.len() == 0
     }
 
     /// Get an iterator over the name/value elements of the list.
@@ -106,8 +106,8 @@ impl Default for CslStringList {
 
 pub struct CslStringListIterator<'a> {
     list: &'a CslStringList,
-    idx: isize,
-    count: isize,
+    idx: usize,
+    count: usize,
 }
 
 impl<'a> CslStringListIterator<'a> {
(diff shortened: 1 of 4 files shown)


src/utils.rs
```diff
@@ -10,6 +10,14 @@ pub fn _string(raw_ptr: *const c_char) -> String {
     c_str.to_string_lossy().into_owned()
 }
 
+pub(crate) fn _string_tuple(raw_ptr: *const c_char, delim: char) -> Option<(String, String)> {
+    let c_str = unsafe { CStr::from_ptr(raw_ptr) };
+    c_str
+        .to_string_lossy()
+        .split_once(delim)
+        .map(|(k, v)| (k.to_string(), v.to_string()))
```
metasim, line 18: Is there a more idiomatic or compact way to go from `(&str, &str)` to `(String, String)`?

src/cpl.rs
```diff
@@ -67,14 +71,30 @@ impl CslStringList {
         Ok(value)
     }
 
+    /// Determine the number of entries in the list.
+    pub fn len(&self) -> isize {
+        unsafe { CSLCount(self.as_ptr()) as isize }
```
lnicola, line 76: Nit:

```suggestion
        unsafe { CSLCount(self.as_ptr()) } as isize
```

src/cpl.rs
```diff
@@ -67,14 +71,30 @@ impl CslStringList {
...
+    pub fn is_empty(&self) -> bool {
+        self.len() <= 0
+    }
+
+    /// Get an iterator over the name/value elements of the list.
+    pub fn iter(&self) -> CslStringListIterator {
+        CslStringListIterator::new(self)
+    }
+
+    /// Get the raw pointer behind list's data.
```
lnicola, line 89:
```suggestion
    /// Get the raw pointer to the underlying data.
```

src/cpl.rs
```diff
@@ -53,12 +56,13 @@ impl CslStringList {
...
-    /// Looks up the value corresponding to a key.
+    /// Looks up the value corresponding to `key`.
     ///
-    /// See `CSLFetchNameValue` for details.
+    /// See [`CSLFetchNameValue`](https://gdal.org/doxygen/cpl__string_8h.html#a4f23675f8b6f015ed23d9928048361a1)
+    /// for details.
     pub fn fetch_name_value(&self, key: &str) -> Result<Option<String>> {
         let key = CString::new(key)?;
-        let c_value = unsafe { gdal_sys::CSLFetchNameValue(self.as_ptr(), key.as_ptr()) };
+        let c_value = unsafe { CSLFetchNameValue(self.as_ptr(), key.as_ptr()) };
```
lnicola, line 65: This seems to be O(N^2). I'm not really familiar with this API, is there way to move to the next element of a `CSLStringList`?

lnicola (approved):

src/cpl.rs
metasim: @lnicola Had to make this `(unsafe { CSLCount(self.as_ptr()) }) as isize` to parse.

src/cpl.rs
```diff
@@ -83,3 +103,125 @@ impl Default for CslStringList {
...
+
+impl<'a> Iterator for CslStringListIterator<'a> {
+    type Item = (String, String);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.is_done() {
+            return None;
+        }
+
+        let field = unsafe { CSLGetField(self.list.as_ptr(), self.idx as libc::c_int) };
```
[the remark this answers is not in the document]
metasim: Ah, yes... [good catch](https://github.com/OSGeo/gdal/blob/3a5039aa47399f72efeb6cd9bf47868385da0796/port/cpl_string.cpp#L175-L188). 

I think the only option is to do the pointer arithmetic on the Rust side. Do you think that's worth the risk?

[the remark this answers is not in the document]
lnicola: I wouldn't worry too much about it, we're full of unsafe code anyway. But I would add a test (if there isn't one already in this PR) to make sure it keeps working.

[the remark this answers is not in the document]
metasim: @lnicola See [6d3081c](https://github.com/georust/gdal/pull/311/commits/6d3081c80cbf07972c8b786d2ec4781d6635a9d5) as an alternative. Which do you prefer?

metasim:
bors r=lnicola

PR #314, opened 2022-09-28
Added ability to create and edit color tables.

metasim:
---

image


src/dataset.rs
```diff
@@ -400,6 +400,13 @@ impl Dataset {
         Ok(Dataset { c_dataset })
     }
 
+    /// Flush all write cached data to disk.
+    ///
+    /// See [`GDALFlushCache`].
+    pub fn flush_cache(&self) {
+        unsafe { GDALFlushCache(self.c_dataset) }
+    }
+
```
metasim, line 409: This is technically not necessary in this PR. I used it to help debug a Linux testing issue and decided to leave it in. Can remove if someone has a problem exposing this.

src/raster/rasterband.rs
```diff
@@ -485,6 +487,13 @@ impl<'a> RasterBand<'a> {
         Some(ColorTable::from_c_color_table(c_color_table))
     }
 
+    /// Set the color table for this band.
+    ///
+    /// See [`ColorTable`] for usage example.
+    pub fn set_color_table(&mut self, colors: &ColorTable) {
+        unsafe { GDALSetRasterColorTable(self.c_rasterband, colors.c_color_table) };
```
metasim, line 494: NB: Per the GDAL docs, `GDALSetRasterColorTable` copies the entries from the given `ColorTable`.

src/raster/rasterband.rs
```diff
@@ -829,13 +844,25 @@ impl PaletteInterpretation {
...
+
+    /// Returns the C API int value of this palette interpretation.
+    pub fn c_int(&self) -> GDALPaletteInterp::Type {
+        match self {
+            Self::Gray => GDALPaletteInterp::GPI_Gray,
+            Self::Rgba => GDALPaletteInterp::GPI_RGB,
+            Self::Cmyk => GDALPaletteInterp::GPI_CMYK,
+            Self::Hls => GDALPaletteInterp::GPI_HLS,
+        }
+    }
```
metasim, line 856: Aside: the `from_c_*` and `c_*` duals are a pattern enough in the code base we may want to consider a unify trait?

src/raster/rasterband.rs
```diff
@@ -852,38 +880,230 @@ pub struct CmykEntry {
...
+/// # Ok(())
+/// # }
+/// ```
 pub struct ColorTable<'a> {
     palette_interpretation: PaletteInterpretation,
     c_color_table: GDALColorTableH,
+    /// If `true`, Rust is responsible for deallocating color table pointed to by
+    /// `c_color_table`, which is the case when instantiated directly, as opposed to
+    /// when read via [`RasterBand::color_table`].
+    rust_owned: bool,
```
metasim, line 1039: To restate from comment above: If created via `new`, Rust deallocates. If retrieved from a `RasterBand`, GDAL deallocates.

src/test_utils.rs
```diff
@@ -12,13 +12,20 @@ impl TempFixture {
...
+        std::fs::copy(&source, &staging.temp_path).unwrap();
+        staging
+    }
 
+    /// Creates a temporary directory and path to a non-existent file with given `name`.
+    /// Useful for writing results to during testing
+    ///
+    /// Returns the struct `TempFixture` that contains the temp dir (for clean-up on `drop`)
+    /// as well as the empty file path.
+    pub fn empty(name: &str) -> Self {
```
metasim, line 26: @ChristianBeilschmidt I evolved this a bit so we could also create empty files that would later be cleaned up after test.

src/dataset.rs
lnicola: No, I think we should have it, but should it take `&self` or `&mut self`. I guess it's mostly academic with `Dataset` being `!Sync`, though.

src/raster/rasterband.rs
lnicola: We could, but I'm not sure it would bring us much. We'd have to make the C representation an associated type, which makes those even more awkward to work with.

src/dataset.rs
metasim: @lnicola This is similar to something I still find really confusing in this library (or maybe Rust?): you can have a non-mutable reference to a `Dataset`, yet get a `RasterBand` from it and change, say, the `no_data` value. I don't think I have the correct mental model for `& mut` yet.

src/dataset.rs
```diff
@@ -400,6 +400,13 @@ impl Dataset {
         Ok(Dataset { c_dataset })
     }
 
+    /// Flush all write cached data to disk.
+    ///
+    /// See [`GDALFlushCache`].
+    pub fn flush_cache(&self) {
```
lnicola, line 406:
```suggestion
    pub fn flush_cache(&self) {
```

Let's take a hint from `std::fs::File` here.

src/raster/rasterband.rs
```diff
@@ -852,38 +880,230 @@ pub struct CmykEntry {
...
+    /// println!("{ct:?}");
+    /// # Ok(())
+    /// # }
+    pub fn color_ramp(
+        start_index: u8,
+        start_color: &ColorEntry,
+        end_index: u8,
+        end_color: &ColorEntry,
+    ) -> Result<ColorTable<'a>> {
+        if start_color.palette_interpretation() != end_color.palette_interpretation() {
```
lnicola, line 1081: I might be wrong, but it feels like this might just as well be an `assert_eq!`, since it's a user error to pass in different types of entries.

But it's fine to leave it like this, if there's prior art or you think it's more user-friendly.

src/raster/rasterband.rs
```diff
@@ -811,15 +820,21 @@ impl ColorInterpretation {
...
+    /// Red, Green, Blue and Alpha
     Rgba,
+    /// Cyan, Magenta, Yellow and Black
     Cmyk,
+    /// Hue, Lightness and Saturation
     Hls,
 }
 
 impl PaletteInterpretation {
+    /// Instantiates Self from the C API int value of [`GDALPaletteInterp`].
```
lnicola, line 837: Nit (also not sure about the hyperlink):

```suggestion
    /// Creates a Rust [`PaletteInterpretation`] from  a C API [`GDALPaletteInterp`] value.
```

src/raster/rasterband.rs
lnicola: I _think_ this looks fine, but I've been wrong before :sweat_smile:.

src/raster/rasterband.rs
```diff
@@ -950,4 +1153,41 @@ impl<'a> ColorTable<'a> {
...
+    }
+}
+
+impl Debug for ColorTable<'_> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let entries = (0..self.entry_count())
+            .filter_map(|i| self.entry(i))
+            .collect::<Vec<_>>();
+
+        f.debug_struct("ColorTable")
```
lnicola, line 1188: Nit-picking again, but I wonder if we shouldn't abbreviate empty (`(0, 0, 0, 0)`) entries at the end or only display the set ones (`[0] => (1, 1, 1, 0), [1] => (2, 2, 2, 0) [...]`).

Totally fine to leave this for another PR though.

lnicola (approved):

src/raster/rasterband.rs
metasim: I lean toward *not* `panic!`-ing on behalf of the user whenever possible. If they want to `unwrap` and get a `panic!` that's their business, but rather give them the option to handle it more gracefully if they want.

metasim: What I would have rather done is make it a compile error :) But that's a rabbit hole.

src/raster/rasterband.rs
metasim: I'll look into that. I'm kinda annoyed in a `gdalinfo` you end up with a long list of zero entries. Before writing/reading the test file, the table is short. At one time I did some work on the Geotrellis GeoTIFF parser, and seem to remember there's legal to have only the entries you specify. Perhaps this is GDAL GTiff implementation thing.

metasim: @lnicola I poked at it a bit an think I'll go with saving it for another PR.

metasim:
bors r=lnicola

PR #318, opened 2022-10-18
Filling out `GDALDataType` metadata and utility support.

metasim:
---

Primarily adds `GdalTypeDescriptor` for enabling runtime inspection of `GDALDataType` properties.

**Note, potentially breaking change**: `GDALDataType` is no longer `pub use` in `gdal::raster`, 
  as `GdalType` and `GdalTypeDescriptor` sufficiently cover use cases in safe code. 
  Still accessible via `gdal_sys::GDALDataType`.

---

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,9 +36,16 @@
 
   - <https://github.com/georust/gdal/pull/324>
 
+- Added `GdalTypeDescriptor` to provide access to metadata and supporting routines around `GDALDataType` ordinals.
+- **Breaking**: `GDALDataType` is no longer `pub use` in `gdal::raster`, 
+  as `GdalType` and `GdalTypeDescriptor` sufficiently cover use cases in safe code. 
+  Still accessible via `gdal_sys::GDALDataType`.
+
+  - <https://github.com/georust/gdal/pull/318>
+
 ## 0.13
 
-- Add prebuild bindings for GDAL 3.5
+- Add prebuilt bindings for GDAL 3.5
 
   - <https://github.com/georust/gdal/pull/277>
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,9 +36,9 @@
 
   - <https://github.com/georust/gdal/pull/324>
 
-- Added `GdalTypeDescriptor` to provide access to metadata and supporting routines around `GDALDataType` ordinals.
+- Added `GdalDataType` to provide access to metadata and supporting routines around `GDALDataType` ordinals.
 - **Breaking**: `GDALDataType` is no longer `pub use` in `gdal::raster`, 
-  as `GdalType` and `GdalTypeDescriptor` sufficiently cover use cases in safe code. 
+  as `GdalType` and `GdalDataType` sufficiently cover use cases in safe code. 
   Still accessible via `gdal_sys::GDALDataType`.
 
   - <https://github.com/georust/gdal/pull/318>
--- a/src/raster/types.rs
+++ b/src/raster/types.rs
@@ -85,7 +85,7 @@ impl GdalDataType {
     /// ```
     pub fn for_value<N: GdalType + Into<f64>>(value: N) -> Self {
         let gdal_type = unsafe { GDALFindDataTypeForValue(value.into(), 0) };
-        Self::try_from(gdal_type).unwrap()
+        Self::try_from(gdal_type).expect("GDALFindDataTypeForValue")
     }
 
     /// Get the name of the [`GDALDataType`].
@@ -113,14 +113,14 @@ impl GdalDataType {
     pub fn bits(&self) -> u8 {
         unsafe { GDALGetDataTypeSizeBits(self.gdal_ordinal()) }
             .try_into()
-            .unwrap()
+            .expect("GDALGetDataTypeSizeBits")
     }
 
     /// Get the [`GDALDataType`] size in **bytes**.
     pub fn bytes(&self) -> u8 {
         unsafe { GDALGetDataTypeSizeBytes(self.gdal_ordinal()) }
             .try_into()
-            .unwrap()
+            .expect("GDALGetDataTypeSizeBytes")
     }
 
     /// Returns `true` if [`GDALDataType`] is integral (non-floating point)
@@ -153,7 +153,7 @@ impl GdalDataType {
     /// ```
     pub fn union(&self, other: Self) -> Self {
         let gdal_type = unsafe { GDALDataTypeUnion(self.gdal_ordinal(), other.gdal_ordinal()) };
-        Self::try_from(gdal_type).unwrap()
+        Self::try_from(gdal_type).expect("GDALDataTypeUnion")
     }
 
     /// Change a given value to fit within the constraints of this [`GDALDataType`].
@@ -310,7 +310,7 @@ pub trait GdalType {
     /// ```
     fn datatype() -> GdalDataType {
         // We can call `unwrap` because existence is guaranteed in this case.
-        Self::gdal_ordinal().try_into().unwrap()
+        Self::gdal_ordinal().try_into().expect("GdalDataType")
     }
 }
 
--- a/src/raster/types.rs
+++ b/src/raster/types.rs
@@ -2,9 +2,9 @@ use crate::errors::{GdalError, Result};
 use crate::utils::{_last_null_pointer_err, _string};
 pub use gdal_sys::GDALDataType;
 use gdal_sys::{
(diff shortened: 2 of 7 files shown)


src/raster/types.rs
```diff
@@ -1,41 +1,482 @@
...
+/// use gdal::raster::{GdalType, GdalTypeDescriptor};
+/// let gdt: GdalTypeDescriptor = 3.try_into().unwrap();
+/// println!("{gdt:#?}")
+/// ```
+impl TryFrom<GDALDataType::Type> for GdalTypeDescriptor {
+    type Error = GdalError;
+
+    fn try_from(value: GDALDataType::Type) -> std::result::Result<Self, Self::Error> {
+        let wrapped = GdalTypeDescriptor(value);
+        if !GdalTypeDescriptor::available_types().contains(&wrapped) {
```
jdroenner, line 237: this could be a match

jdroenner (changes requested):
this PR adds a lot of useful functions. I think an enum would fit a bit better here. What do you think?

metasim:
@jdroenner Thanks for the fantastic feedback! I especially appreciate knowing that the `gdal_sys` leakage isn't actually desired. I'll rework as an `enum`. 🚀

CHANGES.md
```diff
@@ -32,9 +32,16 @@
 
   - <https://github.com/georust/gdal/pull/314>
 
+- Added `GdalTypeDescriptor` to provide access to metadata and supporting routines around `GDALDataType` ordinals.
+- **Breaking** (potentially): `GDALDataType` is no longer `pub use` in `gdal::raster`,
```
lnicola, line 36:
```suggestion
- **Breaking**: `GDALDataType` is no longer `pub use` in `gdal::raster`, 
```

No reason to weasel around :smiley:.

script/doctest-output.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
lnicola, line 1: This seems unrelated, but sure, I can see it being useful.

src/raster/types.rs
```diff
@@ -1,41 +1,482 @@
...
+impl GdalTypeDescriptor {
+    /// Find `GdalTypeDescriptor` by name, as would be returned by [`name`][Self::name].
+    ///
+    /// # Example
+    ///
+    /// ```rust, no_run
+    /// use gdal::raster::{GdalType, GdalTypeDescriptor};
+    /// assert_eq!(GdalTypeDescriptor::from_name("UInt16").unwrap(), <u16>::descriptor())
+    /// ```
+    #[allow(non_snake_case)]
```
lnicola, line 37: What's `non_snake_case` here?

src/raster/types.rs
```diff
@@ -1,41 +1,482 @@
...
+    pub fn from_name(name: &str) -> Result<Self> {
+        let c_name = CString::new(name.to_owned())?;
```
lnicola, line 39:
```suggestion
        let c_name = CString::new(name)?;
```

src/raster/types.rs
```diff
@@ -1,41 +1,482 @@
...
+    /// ```rust, no_run
+    /// use gdal::raster::GdalType;
+    /// assert_eq!(<u16>::descriptor().name(), "UInt16");
+    /// ```
+    pub fn name(&self) -> String {
+        let c_str = unsafe { GDALGetDataTypeName(self.gdal_type()) };
+        if c_str.is_null() {
+            // This case shouldn't happen, because `self` only exists for valid
+            // GDALDataType ordinals.
+            panic!("{}", _last_null_pointer_err("GDALGetDataTypeName"));
```
lnicola, line 86: `GDALGetDataTypeName` doesn't set an error.

src/raster/types.rs
```diff
@@ -1,41 +1,482 @@
...
+                &mut is_clamped,
+                &mut is_rounded,
+            )
+        };
+
+        match (is_clamped > 0, is_rounded > 0) {
+            (false, false) => AdjustedValue::Unchanged(result),
+            (true, false) => AdjustedValue::Clamped(result),
+            (false, true) => AdjustedValue::Rounded(result),
+            (true, true) => panic!("Unexpected adjustment result: clamped and rounded."),
```
lnicola, line 168: Appears to be correct, although the GDAL code is somewhat subtle: for `GDT_Float32` it takes a `double`, casts it to `float`, then back to `double`  :+1:.

src/raster/types.rs
```diff
@@ -1,41 +1,482 @@
...
+    pub fn is_floating(&self) -> bool {
+        (unsafe { GDALDataTypeIsFloating(self.gdal_type()) }) > 0
+    }
+
+    /// Returns `true` if [`GDALDataType`] supports negative values.
+    pub fn is_signed(&self) -> bool {
+        (unsafe { GDALDataTypeIsSigned(self.gdal_type()) }) > 0
+    }
+
+    /// Return the descriptor for smallest [`GDALDataType`] fully contains both data types
```
lnicola, line 120:
```suggestion
    /// Return the descriptor for smallest [`GDALDataType`] that fully contains both data types
```

lnicola (changes requested):

src/raster/types.rs
metasim: Thanks. I need to do a better job of reading the C docs on return type semantics.

script/doctest-output.sh
metasim: I slipped it in instead of creating a new PR. I've needed it during doc development.

src/raster/types.rs
metasim: cruft

src/raster/types.rs
metasim: 🤦

src/raster/types.rs
metasim: @jdroenner I read up on slice patterns, and I still couldn't figure out how to implement `contains` with a `match` 😰

lnicola (approved):

src/raster/types.rs
```diff
@@ -1,41 +1,499 @@
...
+/// ```rust, no_run
+/// use gdal::raster::{GdalType, GdalDataType};
+/// let gdt: GdalDataType = 3.try_into().unwrap();
+/// println!("{gdt:#?}")
+/// ```
+impl TryFrom<u32> for GdalDataType {
+    type Error = GdalError;
+
+    fn try_from(value: u32) -> std::result::Result<Self, Self::Error> {
+        Self::iter()
```
jdroenner, line 264: i think this should be a match on all the known gdal type consts

src/raster/types.rs
```diff
@@ -1,41 +1,499 @@
...
+    /// ```rust, no_run
+    /// use gdal::raster::{GdalType, GdalDataType};
+    /// assert_eq!(GdalDataType::for_value(0), <u8>::datatype());
+    /// assert_eq!(GdalDataType::for_value(256), <u16>::datatype());
+    /// assert_eq!(GdalDataType::for_value(-1), <i16>::datatype());
+    /// assert_eq!(GdalDataType::for_value(<u16>::MAX as f64 * -2.0), <i32>::datatype());
+    /// ```
+    pub fn for_value<N: GdalType + Into<f64>>(value: N) -> Self {
+        let gdal_type = unsafe { GDALFindDataTypeForValue(value.into(), 0) };
+        Self::try_from(gdal_type).unwrap()
```
jdroenner, line 88: please use expect instead of all the unwraps

jdroenner (changes requested):

src/raster/types.rs
```diff
@@ -261,9 +261,25 @@ impl TryFrom<u32> for GdalDataType {
...
+            GDT_UInt32 => Ok(GdalDataType::UInt32),
+            GDT_Int32 => Ok(GdalDataType::Int32),
+            #[cfg(all(major_ge_3, minor_ge_5))]
+            GDT_UInt64 => Ok(GdalDataType::UInt64),
+            #[cfg(all(major_ge_3, minor_ge_5))]
+            GDT_Int64 => Ok(GdalDataType::Int64),
+            GDT_Float32 => Ok(GdalDataType::Float32),
+            GDT_Float64 => Ok(GdalDataType::Float64),
+            GDT_CInt16 | GDT_CInt32 | GDT_CFloat32 | GDT_CFloat64 =>
+                Err(GdalError::BadArgument("Complex data types are not available".into())),
```
metasim, line 280: @jdroenner Do you think this is an appropriate way to handle this case?

jdroenner: yes i think this is a good solution

jdroenner (approved):

Issue #299, opened 2022-09-02
Stack overflow in `programs::raster::mdimtranslate::tests::test_build_tiff_from_path`


lnicola:
```
[14423 stack frames]
#14424 0x00007ffff076ffbc in H5I_inc_ref () from /usr/lib/libhdf5.so.200
#14425 0x00007ffff06e3cdb in H5E__push_stack () from /usr/lib/libhdf5.so.200
#14426 0x00007ffff06e3eef in H5E_printf_stack () from /usr/lib/libhdf5.so.200
#14427 0x00007ffff076ffbc in H5I_inc_ref () from /usr/lib/libhdf5.so.200
#14428 0x00007ffff06e3cdb in H5E__push_stack () from /usr/lib/libhdf5.so.200
#14429 0x00007ffff06e3eef in H5E_printf_stack () from /usr/lib/libhdf5.so.200
#14430 0x00007ffff076ffbc in H5I_inc_ref () from /usr/lib/libhdf5.so.200
#14431 0x00007ffff06e3cdb in H5E__push_stack () from /usr/lib/libhdf5.so.200
#14432 0x00007ffff06e3eef in H5E_printf_stack () from /usr/lib/libhdf5.so.200
#14433 0x00007ffff076ffbc in H5I_inc_ref () from /usr/lib/libhdf5.so.200
#14434 0x00007ffff06e3cdb in H5E__push_stack () from /usr/lib/libhdf5.so.200
#14435 0x00007ffff06e3eef in H5E_printf_stack () from /usr/lib/libhdf5.so.200
#14436 0x00007ffff076ffbc in H5I_inc_ref () from /usr/lib/libhdf5.so.200
#14437 0x00007ffff06e3cdb in H5E__push_stack () from /usr/lib/libhdf5.so.200
#14438 0x00007ffff06e3eef in H5E_printf_stack () from /usr/lib/libhdf5.so.200
#14439 0x00007ffff076ffbc in H5I_inc_ref () from /usr/lib/libhdf5.so.200
#14440 0x00007ffff06e3cdb in H5E__push_stack () from /usr/lib/libhdf5.so.200
#14441 0x00007ffff06e3eef in H5E_printf_stack () from /usr/lib/libhdf5.so.200
#14442 0x00007ffff076ffbc in H5I_inc_ref () from /usr/lib/libhdf5.so.200
#14443 0x00007ffff06e3cdb in H5E__push_stack () from /usr/lib/libhdf5.so.200
[...]

PR #320, opened 2022-10-18
Run NetCDF dependent tests in serial due to crashing on MacOS.

metasim:
---

Attempted workaround for #299. After change, `mdarray` tests no longer seem to stack overflow or `SIGSEGV` on MacOS (`aarch64`).

Edit: not claiming this solves the real problem here, but running tests locally during development is really hampered by this bug.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ semver = "1.0"
 [dev-dependencies]
 tempfile = "3.3"
 trybuild = "1.0"
+serial_test = "0.9.0"
 
 [workspace]
 members = ["gdal-sys"]
--- a/src/raster/mdarray.rs
+++ b/src/raster/mdarray.rs
@@ -797,12 +797,16 @@ impl Attribute {
 
 #[cfg(test)]
 mod tests {
+    // Note: tests herein are run in serial as NetCDF is not thread safe on some platforms
+    // (e.g. MacOS) and crashes when tests are run in parallel.
 
     use super::*;
+    use serial_test::serial;
 
     use crate::{test_utils::TempFixture, Dataset, DatasetOptions, GdalOpenFlags};
 
     #[test]
+    #[serial]
     fn test_root_group_name() {
         let fixture = TempFixture::fixture("byte_no_cf.nc");
 
@@ -819,6 +823,7 @@ mod tests {
     }
 
     #[test]
+    #[serial]
     fn test_array_names() {
         let fixture = TempFixture::fixture("byte_no_cf.nc");
 
@@ -836,6 +841,7 @@ mod tests {
     }
 
     #[test]
+    #[serial]
     fn test_n_dimension() {
         let fixture = TempFixture::fixture("byte_no_cf.nc");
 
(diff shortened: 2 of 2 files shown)


metasim:
@ChristianBeilschmidt Is it worth seeing if #299 goes away if HDF5 files were used in the tests instead of NetCDF files? I'm assuming GDAL's HDF5 support is similar/same to NetCDF here.

lnicola:
Not this again :weary:...

TBH, I'm not too happy about this because of the amount of code that gets pulled in. At least I guess we're already using `thiserror`.

metasim:
> not too happy about this because of the amount of code that gets pulled in

@lnicola It's only pulled in as a dev-dependency... is that what you're referring to?

lnicola:
Well, I'm not going to block merging this PR because of it.

> is only pulled in as a dev-dependency... is that what you're referring to?

Yes, but, but I prefer to avoid what seems like gratuitous dependencies, especially proc macros:

 - they slow down the builds, including CI (which is already slower than it should be); okay, to be fair, we're already pulling in the whole proc macro stack because of `thiserror`. But by default it depends on `futures`?!
 - it has some cognitive overhead. **I can't overstate this.** If there's a mutex in the code, anyone can hit "go to definition" and/or read the doc comment on it. With a proc macro you probably have to look up the crate
 - proc macros don't always interact well with IDEs. For a long time, at least in `rust-analyzer`, completions didn't work inside proc macros because `syn` errors out on invalid code, like `foo.` (which you do all the time). Even when completion works, other IDE features might not (because of unfixed bugs)
 - even so, it's going to be slower, since the IDE needs to serialize the function body and pass it in another process to the proc macro, then read back and deserialize the result -- on every key press. I didn't measure it, but IDE latency is a real concern.
 - people will complain: "why are you pulling in 15 crates instead of using a plain mutex?"
 - what if one of those 15 crates gets compromised in the future?

I'll agree, each of these is a pretty minor concern in isolation. But I don't think using it is worth its weight in this case.

metasim:
@lnicola Yikes! Thanks for explaining. I only looked at the first layer of dependencies shown in crates.io. Lesson learned.

Would be good to not even have this problem. 

Here are the options I see so far:

1. Use a static mutex (I'd need this sketched out... where it lives, where it is acquired in the test, etc).
2. Convert `.nc` files to `.hdf5`. Requires effort from @ChristianBeilschmidt
3. Put these tests under some feature flag? Or target flag?

All in all, I think the goal here is to make testing as easy a possible for contributors. We already have two features (this and the `trybuild` error) that don't happen in CI, but can/do on other hardware. 

If a feature can't be tested on all platforms, should it be behind a feature/target flag?

lnicola:
I'd rather try HDF5 or even Zarr (surely that won't have any threading issues!). We're not testing GDAL here.

lnicola:
@metasim are you on MacOS? It took me quite a few tries to get the crash in #299.

metasim:
@lnicola Yes, MacOS, `aarch64`.

metasim:
Fixed by #321

PR #323, opened 2022-10-24
Utils for getting driver metadata.

metasim:
---


metasim:
Going to rework this into a couple, more focused PRs.

PR #324, opened 2022-10-27
add a driver manager wrapper

ChristianBeilschmidt:
---

I added a wrapper for the `DriverManager.`
I moved some methods to this struct's impl and deprecated them in the `Driver`.
Moreover, I created integration tests that use the manager. They would interfere with other tests.


metasim:
@ChristianBeilschmidt It's interesting that we both were doing work on this at the same time (but I think for different reasons). Here's what I was doing:

https://github.com/georust/gdal/pull/323/files

Why do you think `DriverManager` is necessary over just extending `Driver`?

src/driver.rs
```diff
@@ -366,3 +352,140 @@ impl MajorObject for Driver {
...
+pub struct DriverManager;
+
+impl DriverManager {
+    /// Returns the number of registered drivers.
+    ///
+    /// # Example
+    ///
+    /// ```rust, no_run
+    /// use gdal::Driver;
+    /// println!("{} drivers are registered", Driver::count());
```
metasim, line 367: If `Driver` is to be deprecated, should these examples be updated?

metasim (commented):
I'm interested to know why we need to move the `Driver` methods into `DriverManager`, creating deprecation issues for others. Did you consider just putting the enhanced registration capabilities in `DriverManager` and leaving the `Driver` public API as it is?

src/driver.rs
ChristianBeilschmidt: [b16c050](https://github.com/georust/gdal/pull/324/commits/b16c0501aae2d5e9fb4c357a4ccf567441ad1fcd)

ChristianBeilschmidt:
> I'm interested to know why we need to move the `Driver` methods into `DriverManager`, creating deprecation issues for others. Did you consider just putting the enhanced registration capabilities in `DriverManager` and leaving the `Driver` public API as it is?

As far as I see GDAL's C++-API, it is structured as I did above. It differentiates between DriverManager and Driver. So it makes sense from my point of view to put these methods (previously in Driver) there.

metasim:
@ChristianBeilschmidt Thanks for explaining the correspondence with the C++ implementation. I wasn't aware of that.

jdroenner:
it looks good to me

src/driver.rs
```diff
@@ -366,3 +352,140 @@ impl MajorObject for Driver {
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    /// ```text
+    /// 'VRT' is 'Virtual Raster'
+    /// ```
+    pub fn get_driver(index: usize) -> Result<Driver> {
+        _register_drivers();
+        let c_driver = unsafe { gdal_sys::GDALGetDriver(index.try_into().unwrap()) };
+        if c_driver.is_null() {
```
lnicola, line 402: I don't think `GDALGetDriver` sets the last error: https://github.com/OSGeo/gdal/blob/master/gcore/gdal_priv.h#L1734.

src/driver.rs
```diff
@@ -366,3 +352,140 @@ impl MajorObject for Driver {
...
+    /// ```
+    /// ```text
+    /// Cloud optimized GeoTIFF generator
+    /// ```
+    pub fn get_driver_by_name(name: &str) -> Result<Driver> {
+        _register_drivers();
+        let c_name = CString::new(name)?;
+        let c_driver = unsafe { gdal_sys::GDALGetDriverByName(c_name.as_ptr()) };
+        if c_driver.is_null() {
+            return Err(_last_null_pointer_err("GDALGetDriverByName"));
```
lnicola, line 430: Same as above.

src/driver.rs
```diff
@@ -366,3 +352,140 @@ impl MajorObject for Driver {
...
+        };
+        Ok(Driver { c_driver })
+    }
+
+    /// Register a driver for use.
+    ///
+    /// Wraps [`GDALRegisterDriver()`](https://gdal.org/api/raster_c_api.html#_CPPv418GDALRegisterDriver11GDALDriverH)
+    pub fn register_driver(driver: &Driver) -> usize {
+        let index = unsafe { gdal_sys::GDALRegisterDriver(driver.c_driver) };
```
lnicola, line 439: It's not impossible, but it's pretty hard to end up needing to call this (because we do `_register_drivers` in `Get*`).

src/driver.rs
```diff
@@ -366,3 +352,140 @@ impl MajorObject for Driver {
...
+    /// Wraps [`GDALAllRegister()`](https://gdal.org/api/raster_c_api.html#gdal_8h_1a9d40bc998bd6ed07ccde96028e85ae26)
+    pub fn register_all() {
+        unsafe {
+            gdal_sys::GDALAllRegister();
+        }
+    }
+
+    /// Prevents the automatic registration of all known GDAL drivers when first calling create, open, etc.
+    pub fn prevent_auto_registration() {
+        START.call_once(|| {});
```
lnicola, line 465: :heart:

lnicola (approved):

lnicola:
bors d+

src/driver.rs
ChristianBeilschmidt: [141c8d8](https://github.com/georust/gdal/pull/324/commits/141c8d8f4fab47f28a00010c18dff135bc2fb56e)

src/driver.rs
ChristianBeilschmidt: [141c8d8](https://github.com/georust/gdal/pull/324/commits/141c8d8f4fab47f28a00010c18dff135bc2fb56e)

lnicola (approved):

jdroenner:
bors merge

PR #325, opened 2022-10-30
Call cargo directly instead of using actions-rs/cargo

lnicola:
---

`actions-rs` is unmaintained, and doesn't bring us that much. I think we might lose the nice GitHub integration, but it wasn't really working anyway.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,25 +41,14 @@ jobs:
           override: true
           components: rustfmt, clippy
       - name: Check with Rustfmt
-        uses: actions-rs/cargo@v1
-        with:
-          command: fmt
-          args: --all -- --check
+        run: cargo fmt --all --check
       - name: Check with Clippy
-        uses: actions-rs/clippy-check@v1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-targets ${{ matrix.features }} -- -D warnings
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -41,25 +41,14 @@ jobs:
...
-        with:
-          command: fmt
-          args: --all -- --check
+        run: cargo fmt --all --check
       - name: Check with Clippy
-        uses: actions-rs/clippy-check@v1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-targets ${{ matrix.features }} -- -D warnings
+        run: cargo clippy
```
jdroenner, line 46: I think clippy should get "--all-targets" and "${{ matrix.features }}" otherwise it might skip tests and other code.

lnicola: Oops, not sure how I lost that :grimacing:.

lnicola: Fixed, but now CI fails because we have a ton of clippy warnings.

lnicola: Filed #326. It doesn't actually fail (yet), I was using `beta`.

jdroenner (approved):

jdroenner:
bors merge

lnicola:
bors r=jdroenner

lnicola:
bors retry

lnicola:
bors r=jdroenner

lnicola:
bors retry

PR #328, opened 2022-10-31
Wait for `osgeo/gdal` tests to complete

lnicola:
---

Not sure if this was intentional or not, but it seems like a good idea to wait at least for one of those versions.

And we should probably have a policy for which versions we test on.

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -1,4 +1,12 @@
 status = [
+    "ci gdal-35",
+    "ci gdal-35 (--all-features)",
+    "ci gdal-34",
+    "ci gdal-34 (--all-features)",
+    "ci gdal-33",
+    "ci gdal-33 (--all-features)",
+    "ci gdal-32",
+    "ci gdal-32 (--all-features)",
     "ci ubuntu-lts",
     "ci ubuntu-lts (--all-features)"
 ]


jdroenner:
good catch. Maybe the last three minor versions? My current policy is to include the GDAL version in the latest Ubuntu LTS...

jdroenner (approved):

jdroenner:
bors merge

lnicola:
> My current policy is to include the GDAL version in the latest Ubuntu LTS...

Sorry, what do you mean?

 - that we should test GDAL up to 3.4.1 (from `jammy`) on CI
 - that you try to package GDAL for Debian or Ubuntu so that the LTS picks up the latest version at the time it gets released
 - that you're hoping to package newer GDAL versions for Ubuntu LTS after it gets released (which isn't possible AFAIL)

jdroenner:
oh what i mean is that there should be a CI run that uses the same GDAL version as shipped with the latest Ubuntu LTS. We / I mainly work on Ubuntu and therefore this is the most important CI run for me.

There is no real policy which other versions should be tested. We should discuss this. I guess we could remove the CI for GDAL 3.2 at some point...

PR #334, opened 2022-11-13
Fixes leakage of GDALDataType::Type in RasterBand.

### LINKED ISSUE #333: `gdal::raster::rasterband::RasterBand::band_type` still returns a `GDALDataType::Type`
There seemed to be a change of replacing the `GDALDataType` with a crate enum `GdalDataType`.
However, the `RasterBand` still returns the `u32`: https://github.com/georust/gdal/blob/3eefd647a9d9fdd6d8a11d914f68657800c6fc9f/src/raster/rasterband.rs#L479

Was is intended or forgotten?

metasim:
---

Also fixes related rustdoc, and cleans up `gdal_sys` test warnings.

Closes #333

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- **Breaking**: `RasterBand::band_type` returns `GdalDataType` enum instead of `GDALDataType::Type` ordinal. Fixes [#333](https://github.com/georust/gdal/issues/333)
+- **Breaking**: `RasterBand::band_type` returns the `GdalDataType` enum instead of `GDALDataType::Type` ordinal. Fixes [#333](https://github.com/georust/gdal/issues/333)
 
   - <https://github.com/georust/gdal/pull/334>
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@
 
 - **Breaking**: `RasterBand::band_type` returns `GdalDataType` enum instead of `GDALDataType::Type` ordinal. Fixes [#333](https://github.com/georust/gdal/issues/333)
 
-  - <>
+  - <https://github.com/georust/gdal/pull/334>
 
 ## 0.14
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- **Breaking**: `RasterBand::band_type` returns `GdalDataType` enum instead of `GDALDataType::Type` ordinal. Fixes [#333](https://github.com/georust/gdal/issues/333)
+
+  - <>
+
 ## 0.14
 
 - Added new content to `README.md` and the root docs.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ libc = "0.2"
 geo-types = { version = "0.7" }
 gdal-sys = { path = "gdal-sys", version = "^0.8" }
 ndarray = { version = "0.15", optional = true }
-chrono = { version = "0.4" }
+chrono = { version = "<=0.4.22" }
 bitflags = "1.3"
 once_cell = "1.9"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ libc = "0.2"
 geo-types = { version = "0.7" }
 gdal-sys = { path = "gdal-sys", version = "^0.8" }
 ndarray = { version = "0.15", optional = true }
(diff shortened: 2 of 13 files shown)


src/driver.rs
```diff
@@ -168,7 +168,7 @@ impl Driver {
     /// ds.set_spatial_ref(&SpatialRef::from_epsg(4326)?)?;
     /// assert_eq!(ds.raster_count(), 1);
     /// assert_eq!(ds.raster_size(), (64, 64));
-    /// assert_eq!(ds.rasterband(1)?.band_type(), u8::gdal_ordinal());
+    /// assert_eq!(ds.rasterband(1)?.band_type(), GdalDataType::UInt8);
```
metasim, line 171: This could also be written `u8::datatype()` given the correct imports, but using the enum directly is more intuitive, IMO.

Cargo.toml
```diff
@@ -21,7 +21,7 @@ libc = "0.2"
 geo-types = { version = "0.7" }
 gdal-sys = { path = "gdal-sys", version = "^0.8" }
 ndarray = { version = "0.15", optional = true }
-chrono = { version = "0.4" }
+chrono = { version = "<=0.4.22" }
```
metasim, line 24: This will be reverted when #335 is addressed.

michaelkirk: I'm not super fluent in my cargo versioning constraints, but should this be: `>= 0.4.0 <=0.4.22`

gdal-sys/src/lib.rs
```diff
@@ -3,5 +3,11 @@
 #![allow(non_snake_case)]
 #![allow(clippy::upper_case_acronyms)]
 #![allow(rustdoc::bare_urls)]
+// bindgen test code generates lots of warnings when testing
+// sizes of types due to using constructs such as:
+// `unsafe { &(*(::std::ptr::null::<__sbuf>()))._base as *const _ as usize }`
+// This disables those warnings.
+#![cfg_attr(test, allow(deref_nullptr))]
```
lnicola, line 10: I think we should upgrade `bindgen` instead. This was probably fixed back in May, in https://github.com/rust-lang/rust-bindgen/pull/2203.

src/raster/rasterband.rs
```diff
@@ -476,8 +475,9 @@ impl<'a> RasterBand<'a> {
     }
 
     /// Returns the pixel datatype of this band.
-    pub fn band_type(&self) -> GDALDataType::Type {
-        unsafe { gdal_sys::GDALGetRasterDataType(self.c_rasterband) }
+    pub fn band_type(&self) -> GdalDataType {
+        let ordinal = unsafe { gdal_sys::GDALGetRasterDataType(self.c_rasterband) };
+        ordinal.try_into().unwrap_or(GdalDataType::Unknown)
```
lnicola, line 480: Just looking at the code in the browser, can we call `GdalType::datatype()` here instead?

CHANGES.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- **Breaking**: `RasterBand::band_type` returns `GdalDataType` enum instead of `GDALDataType::Type` ordinal. Fixes [#333](https://github.com/georust/gdal/issues/333)
```
```suggestion
- **Breaking**: `RasterBand::band_type` returns the `GdalDataType` enum instead of `GDALDataType::Type` ordinal. Fixes [#333](https://github.com/georust/gdal/issues/333)
```

src/raster/rasterband.rs
metasim: @lnicola Not sure what you mean... `GdalType::datatype()` would be to get the datatype most closely associated with a static primitive. In this case, we're getting a `c_int` from C-GDAL and need to map that into the `GdalDataType` enum. But perhaps I'm missing what you're suggesting?

lnicola: Basically `ordinal.datatype()` instead of `ordinal.try_into().unwrap_or(GdalDataType::Unknown)`. Maybe it doesn't work for `c_int`, though?

metasim: Yeh, it doesn't work. `GdalType` is a (wierd) type-level trait, with no `self` parameter:

https://github.com/georust/gdal/blob/029b95af44096d34b1ce3dbb80fe43698b16ef70/src/raster/types.rs#L330

In Scala I'd have called this an "evidence" trait. So `<u8>::datatype()` exists, but `3u8.datatype()` does not.

PR #340, opened 2022-11-19
Use tempfile::tempdir instead of /tmp (fix Windows tests)

mwm126:
---


mwm126:
Creating PR from other branch

PR #339, opened 2022-11-19
Fix deprecations introduced in geo-types v0.7.8.

metasim:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/gdal/blob/master/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

PR #336, opened 2022-11-14
Fix deprecations from chrono 0.4.23.

metasim:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/gdal/blob/master/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Fixes #335

PR #218, opened 2021-09-15
Add Windows CI job

pka:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/gdal/blob/master/CODE_OF_CONDUCT.md).
- [ ] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Test failures are ignored for now, because we have a few failures which are listed in ci.yml.

PR #341, opened 2022-11-20
Fix windows tests

mwm126:
---

### DIFF
--- a/src/raster/tests.rs
+++ b/src/raster/tests.rs
@@ -10,7 +10,6 @@ use crate::vsi::unlink_mem_file;
 use crate::DriverManager;
 use gdal_sys::GDALDataType;
 use std::path::Path;
-use tempfile::tempdir;
 
 #[cfg(feature = "ndarray")]
 use ndarray::arr2;
@@ -309,8 +308,7 @@ fn test_create_with_band_type_with_options() {
         },
     ];
 
-    let tmp_dir = tempdir().unwrap();
-    let tmp_filename = tmp_dir.path().join("test.tif");
+    let tmp_filename = TempFixture::empty("test.tif");
     {
         let dataset = driver
             .create_with_band_type_with_options::<u8, _>(&tmp_filename, 256, 256, 1, &options)
--- a/src/raster/tests.rs
+++ b/src/raster/tests.rs
@@ -10,6 +10,7 @@ use crate::vsi::unlink_mem_file;
 use crate::DriverManager;
 use gdal_sys::GDALDataType;
 use std::path::Path;
+use tempfile::tempdir;
 
 #[cfg(feature = "ndarray")]
 use ndarray::arr2;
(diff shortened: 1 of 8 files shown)


src/raster/tests.rs
```diff
@@ -308,17 +309,18 @@ fn test_create_with_band_type_with_options() {
         },
     ];
 
-    let tmp_filename = "/tmp/test.tif";
+    let tmp_dir = tempdir().unwrap();
```
metasim, line 312: Consider using `TempFixture::empty` instead?

https://github.com/georust/gdal/blob/f463ba65592bead0675124702e5887d64710b489/src/test_utils.rs#L21-L33

src/vector/feature.rs
```diff
@@ -749,10 +763,10 @@ impl FieldValue {
     }
 
     /// Interpret the value as `Date`. Returns `None` if the value is something else.
-    pub fn into_date(self) -> Option<Date<FixedOffset>> {
+    pub fn into_date(self) -> Option<NaiveDate> {
         match self {
             FieldValue::DateValue(rv) => Some(rv),
-            FieldValue::DateTimeValue(rv) => Some(rv.date()),
+            FieldValue::DateTimeValue(rv) => Some(rv.date_naive()),
```
metasim, line 769: Now that these are fixed via #336 & #339, can you rebase (or merge) primary focus of PR against `master`?

src/vector/vector_tests/mod.rs
```diff
@@ -851,7 +857,11 @@ mod tests {
             open_flags: GdalOpenFlags::GDAL_OF_UPDATE,
             ..DatasetOptions::default()
         };
-        let tmp_file = "/tmp/test.s3db";
+
+        let tmpdir = tempdir().unwrap();
```
metasim, line 861: Same point about `TempFixture::empty`.

metasim (changes requested):
I'm not actually able to test on a Windows computer, so my review is only source-level. 😄

src/raster/tests.rs
mwm126: done

src/vector/feature.rs
mwm126: Merged `master` into PR 341 branch.  Can squash when merging 341.

mwm126:
Implemented fixes.  I can't test on Windows at the moment, but I can test tomorrow.
Hopefully PR 218 can then be merged!

pka:
#218 disables all broken tests, so it did run fine back then.

mwm126:
The issues with `/tmp/` on Windows are resolved, but I was also getting an error:
```
---- vector::vector_tests::tests::test_write_features stdout ----
thread 'vector::vector_tests::tests::test_write_features' panicked at 'called `Result::unwrap()`
on an `Err` value: Os { code: 32, kind: Uncategorized,
message: "The process cannot access the file because it is being used by another process." }',
 src\vector\vector_tests\mod.rs:793:53
```

My solution was to put the open `ds: Dataset` in it's own scope so that it's closed before being removed (to tidy up after the test.)

metasim (approved):

PR #342, opened 2022-11-20
Consolidated `fixture!` implementation and enabled usage anywhere in module structure.

metasim:
---

There were two versions of `fixture!`, and was working on a PR that needed a third, so this simplifies and consolidates the implementation by making use of relative paths in `[env]` section of `config.toml`.

### DIFF
--- a/src/test_utils.rs
+++ b/src/test_utils.rs
@@ -42,3 +42,10 @@ impl AsRef<Path> for TempFixture {
         self.path()
     }
 }
+
+#[macro_export]
+macro_rules! fixture {
+    ($name:expr) => {
+        std::path::Path::new(env!("FIXTURES_DIR")).join($name)
+    };
+}
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,3 +3,6 @@
 dto = "test --doc -- --show-output"
 # Run clippy, raising warnings to errors
 nowarn = "clippy --all-targets -- -D warnings"
+[env]
+# Used by `fixture!` macro.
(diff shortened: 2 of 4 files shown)


src/test_utils.rs
```diff
@@ -42,3 +42,10 @@ impl AsRef<Path> for TempFixture {
         self.path()
     }
 }
+
+#[macro_export]
+macro_rules! fixture {
+    ($name:expr) => {
+        std::path::Path::new(env!("FIXTURES_DIR")).join($name)
```
metasim, line 49: Should we just convert this into a function? Seems silly now to make it a macro...

lnicola:
Removing the duplicated code and turning the macro into a function are good ideas, but I'm not sure what the extra environment variable is bringing.

I suppose we could use it spread the test fixtures around, but we only have two crates now. It seems to me that the current arrangement (with paths relative to the manifest) is more predictable and easier to understand.

metasim:
@lnicola 
> but I'm not sure what the extra environment variable is bringing

🤔 🤦 I got so wrapped up the old implementation that I missed the "forest for the trees". Yeh, extra indirection is kinda silly.

src/vector/vector_tests/sql.rs
```diff
@@ -1,14 +1,14 @@
 use std::collections::HashSet;
 
+use crate::test_utils::fixture;
```
lnicola, line 3: Nit: we should probably use a consistent style (see line below), but it doesn't matter that much.

lnicola (approved):

metasim:
bors r=lnicola

PR #343, opened 2022-11-21
try Windows CI (from PR 218)

mwm126:
---


mwm126:
I wanted to investigate whether PR 218 would be able to be merged with all tests enabled on Windows, but there still is a failure.  I will investigate locally on my own Windows system.

PR #344, opened 2022-11-21
Metadata iterator

metasim:
---

Provides an iterator over `Metadata`.

Also fleshes out other documentation around `Metadata`.


src/raster/tests.rs
```diff
@@ -173,93 +173,6 @@ fn test_rename_remove_raster() {
...
-    let dataset = driver.create("", 1, 1, 1).unwrap();
-    let mut band = dataset.rasterband(1).unwrap();
-
-    let description = "A merry and cheerful band description";
-    assert_eq!(band.description().unwrap(), "");
-
-    band.set_description(description).unwrap();
-    assert_eq!(band.description().unwrap(), description);
-}
-
```
metasim, line 262: This was all moved to `metadata.rs`.

src/metadata.rs
```diff
@@ -74,23 +121,26 @@ pub trait Metadata: MajorObject {
...
-                    if p.is_null() {
-                        break;
-                    }
-
-                    metadata.push(_string(p));
-                }
+            } else {
+                metadata.append(&mut _string_array(c_res));
             }
         }
```
metasim, line 127: Cleanup along the way. This pattern was already in `_string_array`.

src/metadata.rs
```diff
@@ -128,13 +194,237 @@ pub trait Metadata: MajorObject {
...
+
+/// Standalone metadata entry, as returned by iterator from [`Metadata::metadata_iter`].
+///
+/// Defined by it's parent `domain`, and `key`/`value` pair.
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct MetadataEntry {
+    pub domain: String,
+    pub key: String,
+    pub value: String,
+}
```
metasim, line 246: Question: are there more preferable ways of packaging up the iterator results?

src/metadata.rs
```diff
@@ -128,13 +194,237 @@ pub trait Metadata: MajorObject {
...
+            .any(|e| e == MetadataEntry::new("", "AREA_OR_POINT", "Area")));
+        assert!(ds
+            .metadata_iter()
+            .any(|e| e.domain == "DERIVED_SUBDATASETS"));
+
+        // RasterBand metadata...
+        let ds = Dataset::open(fixture!("labels.tif")).unwrap();
+        let band = ds.rasterband(1).unwrap();
+        assert!(band.metadata_iter().any(|e| e.key == "CLASSES"));
+    }
```
metasim, line 429: New test for iterator.

src/metadata.rs
```diff
@@ -128,13 +194,237 @@ pub trait Metadata: MajorObject {
...
+        };
+    }
+
+    #[test]
+    fn test_get_dataset_driver() {
+        let dataset = Dataset::open(fixture!("tinymarble.png")).unwrap();
+        let driver = dataset.driver();
+        assert_eq!(driver.short_name(), "PNG");
+        assert_eq!(driver.long_name(), "Portable Network Graphics");
+    }
```
metasim, line 328: This and the next several tests were moved from `raster/tests.rs`.

src/driver.rs
```diff
@@ -368,7 +368,7 @@ impl DriverManager {
 
     /// Returns the driver with the given short name or [`Err`] if not found.
     ///
-    /// See also: [`count`](Self::count), [`get`](Self::get)
+    /// See also: [`count`](Self::count), [`get`](Self::get_driver_by_name)
```
metasim, line 371: Documentation bug left over from `DriverManager` PR.

src/metadata.rs
```diff
@@ -39,6 +42,21 @@ use std::ffi::CString;
...
+    /// use gdal::{Dataset, Metadata};
+    /// # fn main() -> gdal::errors::Result<()> {
+    /// let dataset = Dataset::open("fixtures/tinymarble.tif")?;
+    /// // `description` on a `Dataset` is usually the file name.
+    /// let desc = dataset.description()?;
+    /// assert_eq!(desc, "fixtures/tinymarble.tif");
+    /// # Ok(())
+    /// # }
+    /// ```
     fn description(&self) -> Result<String> {
```
metasim, line 60: New docs for this and several other pre-existing methods.

src/metadata.rs
```diff
@@ -47,25 +65,54 @@ pub trait Metadata: MajorObject {
...
 
         if !c_res.is_null() {
-            for i in 0.. {
-                let p = unsafe { *c_res.offset(i) };
-                if p.is_null() {
-                    break;
-                }
-
-                domains.push(_string(p));
-            }
```
metasim, line 62: Prior duplicated code cleanup. This pattern was already in `_string_array`.

src/metadata.rs
```diff
@@ -128,13 +194,237 @@ pub trait Metadata: MajorObject {
...
+impl<'a> MetadataIter<'a> {
+    fn new<P: Metadata>(parent: &'a P) -> Self {
+        let stream = parent
+            .metadata_domains()
+            .into_iter()
+            .flat_map(move |domain| {
+                let keyvals = parent.metadata_domain(&domain).unwrap_or_default();
+                keyvals.into_iter().filter_map(move |keyval| {
+                    keyval
+                        .split_once('=')
```
metasim, line 283: [GDAL API Docs](https://gdal.org/api/gdaldataset_cpp.html#_CPPv4N11GDALDataset11GetMetadataEPKc) state this is how key/value pairs are delimited.

src/metadata.rs
```diff
@@ -47,25 +65,54 @@ pub trait Metadata: MajorObject {
...
         domains
     }
 
+    /// Get all the metadata values within the given `domain`. Returns `None` if
+    /// domain is not defined.
+    /// Entries in the returned `Vec<String>` are formatted as “Name=value” pairs
+    ///
+    /// # Arguments
+    ///
+    /// * domain – the domain of interest. Use `“”` for the default domain.
```
lnicola, line 104: Let's not use smart quotes here, they might cause confusion.

lnicola, line 104:
```suggestion
    /// * domain – the domain of interest. Use `""` for the default domain.
```

src/metadata.rs
```diff
@@ -74,23 +121,26 @@ pub trait Metadata: MajorObject {
...
-                }
+            } else {
+                metadata.append(&mut _string_array(c_res));
             }
         }
 
         Some(metadata)
     }
 
+    /// Get a single metadata entry, as indicated by `key` and `domain`.
```
lnicola, line 132: The other tests are `no_run`, but this one isn't. I'm not sure, but I think we _should_ run them if possible?

src/metadata.rs
```diff
@@ -128,13 +194,237 @@ pub trait Metadata: MajorObject {
...
+
+    /// Determine if this entry is from the default domain, which is named `""`.
+    pub fn is_default_domain(&self) -> bool {
+        self.domain.is_empty()
+    }
+}
+
+/// Metadata iterator state
+pub struct MetadataIter<'a> {
+    stream: Box<dyn Iterator<Item = MetadataEntry> + 'a>,
```
lnicola, line 270: It doesn't matter too much, but could we avoid the `Box` here, for example with https://doc.rust-lang.org/std/iter/fn.from_fn.html (by moving the code into `metadata_iter`).

src/metadata.rs
```diff
@@ -128,13 +194,237 @@ pub trait Metadata: MajorObject {
...
+    /// ```
+    /// Output:
+    /// ```text
+    /// IMAGE_STRUCTURE: INTERLEAVE=PIXEL
+    /// DEFAULT: AREA_OR_POINT=Area
+    /// ...
+    /// DERIVED_SUBDATASETS: DERIVED_SUBDATASET_1_NAME=DERIVED_SUBDATASET:LOGAMPLITUDE:fixtures/tinymarble.tif
+    /// DERIVED_SUBDATASETS: DERIVED_SUBDATASET_1_DESC=log10 of amplitude of input bands from fixtures/tinymarble.tif
+    /// ```
+    fn metadata_iter(&self) -> MetadataIter
```
lnicola, line 230: Would it be more idiomatic to call this `metadata`?

src/metadata.rs
metasim: Editor accident. :/

src/metadata.rs
metasim: @lnicola I've been marking all doctests as `no_run` based on prior feedback. Any thoughts on what should be run and what shouldn't? (In this case, it was an oversight on my part).

lnicola: That's because docs.rs doesn't have GDAL, right? Then I suppose `no_run` makes sense.

metasim: aaaah..... 


Just checked out [this](https://github.com/rust-lang/docs.rs/blob/master/README.md#build-subcommand) test against the `docs.rs` environment without actually publishing to docs.rs... but looks like you have to literally build your own docs.rs and then your docs in a container.

src/metadata.rs
metasim: Wow! I didn't know about `std::iter::from_fn`!!

src/metadata.rs
lnicola: You can make them `no_run`, see https://github.com/georust/gdal/blob/master/gdal-sys/build.rs#L89.

src/metadata.rs
lnicola: Generators can't come too soon :-).

src/metadata.rs
metasim: > 



> Would it be more idiomatic to call this `metadata`?

@lnicola How about just `iter()`?

metasim: > How about just `iter()`?

Nevermind... since this is a trait, it could/would clash with other `iter`s on the other types.

metasim: @lnicola I'm a bit stuck not boxing this and yet keeping the return type tidy. 

Of course I really want this:

    fn metadata(&self) -> impl Iterator<MetadataEntry>


But can't until [_Impl trait initiative_](https://rust-lang.github.io/impl-trait-initiative/explainer/rpit_trait.html) happens.

Expanding (and leaking) the _actual_ type leads to this:

    fn metadata(&self) -> FromFn<fn() -> FlatMap<IntoIter<String>, FilterMap<IntoIter<String>, fn(String) -> Option<MetadataEntry>>, fn(String) -> FilterMap<IntoIter<String>, fn(String) -> Option<MetadataEntry>>>>

🤮 

I then tried this:

    fn metadata(&self) -> FromFn<fn() -> Option<MetadataEntry>>

But get the error 

    closures can only be coerced to `fn` types if they do not capture any variables

(and the closure has to capture _something_ to maintain state).

So the best I'm able to come up with is this:

    fn metadata(&self) -> FromFn<Box::<dyn FnMut() -> Option<MetadataEntry> + '_>>

What else should I be considering?

metasim: There's also this, which I actually like the best because it leaks fewer internals:

    fn metadata(&self) -> Box<dyn Iterator<Item = MetadataEntry> + '_>

src/metadata.rs
metasim: @lnicola Please see comments after `metadata_iter` -> `metadata` comment. I should have added them here.

src/metadata.rs
lnicola: Oh right, it's a trait.

lnicola: Oof, sorry, but now I'm thinking `impl Iterator` isn't ideal here. Most people seem to avoid it in public APIs, and it probably makes thing like storing the iterator in a struct or implementing more traits (e.g. `DoubleEndedIterator`) harder.

lnicola: > My assumption is because it requires nightly to have a trait with return position impl?

Yes, but `std` doesn't use it outside traits either.

metasim: So which approach should we choose?

lnicola: I'm inclined to say that the explicit (`MetadataIter`) version is probably better.

metasim:
@lnicola Thanks for all the constructive feedback. I think it's all incorporated.

lnicola (approved):

metasim:
bors r=lnicola

PR #349, opened 2022-12-05
Handling cases where feature may not have any geometry at all

phayes:
- [✓] I agree to follow the project's [code of conduct](https://github.com/georust/gdal/blob/master/CODE_OF_CONDUCT.md).
- [✓] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

I've run into a few instances where I'm loading data uploaded from users (for example CSV files) where GDAL isn't able to find any geometry.  In instances where you might be working with user-provided data with no geometry, calls to `Feature::geometry()` simply panic. 

This PR documents renders `Feature::geometry` non-panicking by returning an `Option<&Geometry>`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- **Breaking**: `Feature::geometry` returns an `Option<&Geometry>` instead of `&Geometry`. Calls to `Feature::geometry` will no longer panic.
+
+  - <https://github.com/georust/gdal/pull/349>
+
 - **Breaking**: `RasterBand::band_type` returns the `GdalDataType` enum instead of `GDALDataType::Type` ordinal. Fixes [#333](https://github.com/georust/gdal/issues/333)
 
   - <https://github.com/georust/gdal/pull/334>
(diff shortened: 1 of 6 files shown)


lnicola:
It looks like GDAL returns `nullptr` for these. I think I'd prefer returning an `Option<&Geometry>` here.

phayes:
@lnicola , 

I was wondering about returning an `Option<&Geometry>`.  In 99.9% of cases, you're pretty much always working with data that has geometry, so I was feeling uncertain about forcing all users to unwrap and handle an `Option` instead of just documenting the panic and providing a `has_geometry` guard function. 

I'd be happy to rework this into returning an `Option` if that's the consensus.

lnicola:
Agreed, but in more than 99.9% of cases, writing to a file will succeed, yet Rust still makes us check for the outcome. It's just more idiomatic this way.

(and you don't strictly need to `unwrap`, pattern matching (e.g. `let Some(geom) = feat.geometry() else { continue; }`) is usually enough)

metasim:
@phayes 1000% `Option`.

phayes:
Thanks @metasim , @lnicola . I have now reworked this PR so that `Feature::geometry` return an `Option`

lnicola (approved):

metasim:
bors r=lnicola

PR #350, opened 2022-12-10
Remove test failing on GDAL 3.6

lnicola:
---

### DIFF
--- a/src/vsi.rs
+++ b/src/vsi.rs
@@ -291,28 +291,6 @@ mod tests {
     }
 
     #[test]
-    fn unable_to_create() {
-        let file_name = "";
-
-        assert!(matches!(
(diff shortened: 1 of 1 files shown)


lnicola:
I'm not sure what this is testing, but it no longer works. Maybe VSI now allows files without a name.

```
---- vsi::tests::unable_to_create stdout ----
thread 'vsi::tests::unable_to_create' panicked at 'called `Result::unwrap_err()` on an `Ok` value: ()', src/vsi.rs:298:61
```

lnicola:
Yeah, it's intentional: https://github.com/OSGeo/gdal/blob/master/NEWS.md#port.

metasim (approved):

Issue #306, opened 2022-09-04
Provide an API to create a Geometry object from an owned OGRGeometryH


thomas001:
Currently not all of the C Geometry API is wrapped, so users will need to go to `gdal-sys` and raw unsafe C calls from time to time. This is fine, wrapping all of GDAL is a huge undertaking. So right now I have code like this:

```rust
let p = gdal_sys::OGR_G_MakeValid(geom.c_geometry());
let v = Geometry::lazy_feature_geometry();
v.set_c_geometry(p);
v
```

As you can see creating a `Geometry` back from the raw C pointer is cumbersome and it leaks memory since `v` doesn't own the C geometry. I am asking for a function like `Geometry::from_owned_c_geometry` which will own the passed C object. So the above code could become

```rust
unsafe { 
  Geometry::from_owned_c_geometry(gdal_sys::OGR_G_MakeValid(geom.c_geometry()))
}
```

Please consider implementing it in public API. There seems to be a [crate local function](https://github.com/georust/gdal/blob/875d6440b4b9943e67ec8a870c4b423b0f2f73b8/src/vector/geometry.rs#L59) that almost does this, so maybe just change it to public .

metasim:
@thomas001 Is `OGR_G_MakeValid` just one of many calls you're wanting to make, or would the more convenient approach be to add something like `Geometry::make_valid` instead? 

In the future, I wonder if we would we regret letting an internal hack leak out (to never be put back in the box), and wish we'd just go ahead and implement the desired functionality? Not sure what the scope is of what you need, but feel like it's worth asking the question.

thomas001:
@metasim Okay here is my brain dump on this topic, sorry if it is a bit unstructured.

There are a few functions that I missed, `MakeValid` is one of them, but also some set operation (`Union`, `Difference`, ...) or things like `Polygonize` seem to be not wrapped. 

Yes, implementing the missing functions would be one way to go forward here (for a simple case like `MakeValid` I could also send a PR). 

The more general question is: if you agree that not all functionality of GDAL is wrapped in this crate, should we offer users at least a supported way to go back and forth with `gdal-sys` which is more of a complete (since generated) wrapper?

Finally, I think there are some hackish functions already out there. There is at least `set_c_geometry` which is just not taking ownership of the C geometry.

So maybe two possible directions:

- C interop is supported and common, then my ask seems valid
- C interop is considered a hack and users should only interact using safe rust API, then
[...]

PR #356, opened 2022-12-31
Implements `Geometry::make_valid`.

metasim:
---

Partially addresses #306.

cc: @thomas001

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,10 @@
 
   - <https://github.com/georust/gdal/pull/355>
 
+- Exposed various functions on `Geometry`:  `make_valid`, `geometry_name`, and `point_count`. 
+  
+  - <https://github.com/georust/gdal/pull/356>
+
 ## 0.14
 
 - Added new content to `README.md` and the root docs.
--- a/src/cpl.rs
+++ b/src/cpl.rs
@@ -7,13 +7,15 @@ use std::ffi::CString;
 use std::fmt::{Debug, Formatter};
 use std::ptr;
 
-use gdal_sys::{CSLCount, CSLDestroy, CSLFetchNameValue, CSLSetNameValue};
+use gdal_sys::{CSLCount, CSLDestroy, CSLDuplicate, CSLFetchNameValue, CSLSetNameValue};
 use libc::c_char;
 
 use crate::errors::{GdalError, Result};
 use crate::utils::{_string, _string_tuple};
 
-/// Wraps a [`gdal_sys::CSLConstList`]  (a.k.a. `char **papszStrList`)
+/// Wraps a [`gdal_sys::CSLConstList`]  (a.k.a. `char **papszStrList`). This data structure
+/// (a null-terminated array of null-terminated strings) is used throughout GDAL to pass
+/// `KEY=VALUE`-formatted options to various functions.
 ///
 /// See the [`CSL*` GDAL functions](https://gdal.org/api/cpl.html#cpl-string-h) for more details.
 pub struct CslStringList {
@@ -21,6 +23,7 @@ pub struct CslStringList {
 }
 
 impl CslStringList {
+    /// Creates an empty GDAL string list.
     pub fn new() -> Self {
         Self {
             list_ptr: ptr::null_mut(),
@@ -104,6 +107,14 @@ impl Default for CslStringList {
     }
 }
 
+impl Clone for CslStringList {
+    fn clone(&self) -> Self {
+        let list_ptr = unsafe { CSLDuplicate(self.list_ptr) };
+        Self { list_ptr }
+    }
+}
+
+/// State for iterator over [`CslStringList`] entries.
 pub struct CslStringListIterator<'a> {
     list: &'a CslStringList,
     idx: usize,
@@ -152,6 +163,25 @@ impl Debug for CslStringList {
             f.write_fmt(format_args!("{k}={v}\n"))?;
         }
         Ok(())
+    }
+}
+
+/// Convenience shorthand for specifying an empty `CslStringList` to functions accepting
+/// `Into<CslStringList>`.
+impl From<()> for CslStringList {
+    fn from(_: ()) -> Self {
+        CslStringList::default()
+    }
+}
+
+/// Creates a [`CslStringList`] from a slice of _key_/_value_ tuples.
+impl<const N: usize> From<&[(&str, &str); N]> for CslStringList {
+    fn from(pairs: &[(&str, &str); N]) -> Self {
+        let mut result = Self::default();
+        for (k, v) in pairs {
+            result.set_name_value(k, v).expect("valid key/value pair");
+        }
+        result
     }
 }
 
--- a/src/cpl.rs
+++ b/src/cpl.rs
@@ -168,13 +168,35 @@ impl Debug for CslStringList {
 
 /// Convenience shorthand for specifying an empty `CslStringList` to functions accepting
 /// `Into<CslStringList>`.
+///
+/// # Example
+///
+/// ```rust, no_run
+/// use gdal::cpl::CslStringList;
+/// fn count_opts<O: Into<CslStringList>>(opts: O) -> usize {
+///     opts.into().len()
+/// }
+///
+/// assert_eq!(count_opts(()), 0);
+/// ```
 impl From<()> for CslStringList {
     fn from(_: ()) -> Self {
         CslStringList::default()
     }
 }
 
-/// Creates a [`CslStringList`] from a slice of _key_/_value_ tuples.
+/// Convenience for creating a [`CslStringList`] from a slice of _key_/_value_ tuples.
+///
+/// # Example
+///
+/// ```rust, no_run
+/// use gdal::cpl::CslStringList;
+/// fn count_opts<O: Into<CslStringList>>(opts: O) -> usize {
+///     opts.into().len()
+/// }
+///
+/// assert_eq!(count_opts(&[("One", "1"), ("Two", "2"), ("Three", "3")]), 3);
+/// ```
 impl<const N: usize> From<&[(&str, &str); N]> for CslStringList {
     fn from(pairs: &[(&str, &str); N]) -> Self {
         let mut result = Self::default();
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -235,7 +235,7 @@ pub trait Metadata: MajorObject {
     }
 }
 
-/// Standalone metadata entry, as returned by iterator from [`Metadata::metadata_iter`].
+/// Standalone metadata entry, as returned by iterator from [`Metadata::metadata`].
 ///
 /// Defined by it's parent `domain`, and `key`/`value` pair.
 #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
--- a/src/vector/geometry.rs
+++ b/src/vector/geometry.rs
@@ -478,7 +478,7 @@ impl Geometry {
     /// ```rust, no_run
     /// use gdal::vector::Geometry;
     /// # fn main() -> gdal::errors::Result<()> {
-    /// let src = Geometry::from_wkt("POLYGON ((0 0,10 10,0 10,10 0,0 0))")?;
+    /// let src = Geometry::from_wkt("POLYGON ((0 0, 10 10, 0 10, 10 0, 0 0))")?;
     /// let dst = src.make_valid(())?;
     /// assert_eq!("MULTIPOLYGON (((10 0,0 0,5 5,10 0)),((10 10,5 5,0 10,10 10)))", dst.wkt()?);
     /// # Ok(())
--- a/src/vector/geometry.rs
+++ b/src/vector/geometry.rs
@@ -229,7 +229,7 @@ impl Geometry {
     ///
     /// `index` is the line string vertex index, from 0 to `point_count()-1`, or `0` when a point.
     ///
-    /// Refer: [`OGR_G_GetPoint`](https://gdal.org/api/vector_c_api.html#_CPPv414OGR_G_GetPoint12OGRGeometryHiPdPdPd)
+    /// See: [`OGR_G_GetPoint`](https://gdal.org/api/vector_c_api.html#_CPPv414OGR_G_GetPoint12OGRGeometryHiPdPdPd)
     pub fn get_point(&self, index: i32) -> (f64, f64, f64) {
         let mut x: c_double = 0.;
         let mut y: c_double = 0.;
@@ -300,14 +300,14 @@ impl Geometry {
 
     /// Get the geometry type ordinal
     ///
-    /// Refer: [OGR_G_GetGeometryType](https://gdal.org/api/vector_c_api.html#_CPPv421OGR_G_GetGeometryType12OGRGeometryH)
+    /// See: [OGR_G_GetGeometryType](https://gdal.org/api/vector_c_api.html#_CPPv421OGR_G_GetGeometryType12OGRGeometryH)
     pub fn geometry_type(&self) -> OGRwkbGeometryType::Type {
         unsafe { gdal_sys::OGR_G_GetGeometryType(self.c_geometry()) }
     }
 
     /// Get the WKT name for the type of this geometry.
     ///
-    /// Refer: [`OGR_G_GetGeometryName`](https://gdal.org/api/vector_c_api.html#_CPPv421OGR_G_GetGeometryName12OGRGeometryH)
+    /// See: [`OGR_G_GetGeometryName`](https://gdal.org/api/vector_c_api.html#_CPPv421OGR_G_GetGeometryName12OGRGeometryH)
     pub fn geometry_name(&self) -> String {
         // Note: C API makes no statements about this possibly returning null.
         // So we don't have to result wrap this,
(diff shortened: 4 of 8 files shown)


src/metadata.rs
```diff
@@ -235,7 +235,7 @@ pub trait Metadata: MajorObject {
     }
 }
 
-/// Standalone metadata entry, as returned by iterator from [`Metadata::metadata_iter`].
+/// Standalone metadata entry, as returned by iterator from [`Metadata::metadata`].
```
metasim, line 238: Stray bug.

src/vector/geometry.rs
```diff
@@ -388,21 +389,18 @@ impl Geometry {
...
 
-    /// May or may not contain a reference to a SpatialRef: if not, it returns
-    /// an `Ok(None)`; if it does, it tries to build a SpatialRef. If that
-    /// succeeds, it returns an Ok(Some(SpatialRef)), otherwise, you get the
-    /// Err.
+    /// Get the spatial reference system for this geometry.
     ///
+    /// Returns `Some(SpatialRef)`, or `None` if one isn't defined.
+    ///
+    /// Refer: [OGR_G_GetSpatialReference](https://gdal.org/doxygen/ogr__api_8h.html#abc393e40282eec3801fb4a4abc9e25bf)
```
metasim, line 396: Slipping in documentation synchronization from #355.

src/vector/geometry.rs
```diff
@@ -388,21 +389,18 @@ impl Geometry {
...
         let c_spatial_ref = unsafe { gdal_sys::OGR_G_GetSpatialReference(self.c_geometry()) };
 
         if c_spatial_ref.is_null() {
             None
         } else {
-            match unsafe { SpatialRef::from_c_obj(c_spatial_ref) } {
-                Ok(sr) => Some(sr),
-                Err(_) => None,
-            }
+            unsafe { SpatialRef::from_c_obj(c_spatial_ref) }.ok()
```
metasim, line 403: Consistency.

CHANGES.md
```diff
@@ -22,6 +22,10 @@
 
   - <https://github.com/georust/gdal/pull/355>
 
+- Implemented `Geometry::make_valid` and `Geometry::make_valid_ex`.
```
metasim, line 25: When reviewing the following changes, I'm interested in knowing what our policy should be here: 

a. Mirror `foo` & `foo_ex` patterns in C/C++ API
b. Have only `foo`, but always require `FooOpts`, providing an appropriate `FooOpts::default()` to minimize additional ceremony.

Question will make more sense in the context of what is to come.

src/metadata.rs
lnicola: Why can't rustdoc warn about this, I wonder?

src/vector/geometry.rs
```diff
@@ -593,4 +663,38 @@ mod tests {
         // We don't care what it returns when passed an invalid value, just that it doesn't crash.
         geometry_type_to_name(4372521);
     }
+
+    #[test]
+    pub fn test_make_valid() {
```
metasim, line 668: Tests ported from [GDAL](https://github.com/OSGeo/gdal/blob/9f541e8193bd7b1b59ef1aa9ada3bc235fdb56e6/autotest/ogr/ogr_geom.py#L3949).

src/vector/geometry.rs
```diff
@@ -480,9 +514,45 @@ impl Debug for GeometryRef<'_> {
...
+    Linework,
+    /// First makes all rings valid, then merges shells and subtracts holes from shells to generate valid result.
+    ///
+    /// Assumes holes and shells are correctly categorized.
+    Structure {
+        /// If `false`, collapses are converted to empty geometries.
+        /// If `true`, collapses are converted to a valid geometry of lower dimension.
+        keep_collapsed: bool,
+    },
+}
```
metasim, line 531: Purpose here is to try to encode possible options in a way that only expresses valid combination of options. I'd like to think our attempting to provide these ergonomics are in-scope for this library, but I could see arguments against it too.

w.r.t. the opening question: should we just have a single `make_valid` that always requires this (e.g. `geom.make_valid(MakeValidOpts::default())`) or is it better have two separate methods (as this implementation has)?

src/vector/geometry.rs
```diff
@@ -480,9 +514,45 @@ impl Debug for GeometryRef<'_> {
     }
 }
 
+/// Invocation options for [`Geometry::make_valid_ex`].
+#[derive(Debug, Clone, Default)]
+pub enum MakeValidOpts {
```
lnicola, line 519: I feel (not that strongly, though) we should allow the caller to pass in their own options in case a future version of GDAL adds an option we don't know about yet,

CHANGES.md
lnicola: I'm inclined to prefer single functions instead of pairs like these. Note that GDAL already has some `Ex` functions, so us making up new ones might end up being quite confusing. But I guess that's not such a great risk, GDAL would have to add a new parameter that can't be expressed as an existing `CSLStringList`.

src/vector/geometry.rs
lnicola: Yeah, but the valid combinations today might be insufficient tomorrow. We could either allow passing a `CSLStringList`, or add a `Custom` variant here -- but then what happens when GDAL adds another unrelated option?

src/vector/geometry.rs
metasim: Thoughts on how that might be reflected in the implementation while not falling back to everything accepting `CslStringList` (which I think misses an opportunity for safer usage).

src/metadata.rs
metasim: It does, if you remember to run it! 😄.

I guess the more specific question is: should we add this as a CI check?

lnicola: I'd say yes. Are we clean on those warnings now?

CHANGES.md
metasim: What if the signature was something like `fn make_valid<O: Into<CslStringList>>(&self, opts: O)`? That way you could use `MakeValidOpts` if you want, or do it all yourself manually? 🤔 

(NB: I think figuring out the right pattern here is important, because looking ahead I see more of this sort of things coming up).

src/metadata.rs
metasim: Yes, should be. Note, I use [this](https://github.com/georust/gdal/blob/master/.cargo/config.toml#L3) (`cargo dto`) when writing docs to help me catch errors in the examples. `rustdoc` isn't enought. Seems we need a combination of that, disallowing `rustdoc` warnings (similar to [this](https://github.com/georust/gdal/blob/master/.cargo/config.toml#L5)).

src/vector/geometry.rs
metasim: How about this approach?

https://github.com/georust/gdal/pull/356#discussion_r1059668374

It's in times like these I wish Rust had default arguments...

CHANGES.md
lnicola: Yeah, I thought about that, but it has some disadvantages:

 - it makes the function generic, increasing the compile times and bloating the code if we're not careful to split it in two
 - it's less discoverable, you're ready to use the function, then you realize you have to see what implements `Into<CslStringList>`. Examples can help here and I think rustdoc shows these implementations when you look at the trait, but it's still a bit distracting.

src/metadata.rs
lnicola: This isn't an error in an example, it's a reference to an undefined symbol. But I'll take your word for it :-).

metasim: Sorry... was conflating documentation error contexts 🤦 . Both are potential issues.

CHANGES.md
metasim: @lnicola Would you mind taking a look at this latest version? [Here's how it looks in use](https://github.com/georust/gdal/pull/356/files#diff-4597c3ebc0daa4f8f578c4684bc8c616cb21d6ec4328936ff017c23444d7330cR691-R737).

TBH, because I don't have as much of an aversion to generics/monomorphism (when compared to the benefits), and think documentation can assist with discoverability, I much prefer the `Into<CslStringList>` approach to what is in this version. But I admittedly I have a bias toward ergonomics and have never had to deal with the consequences of code bloat. 

That said, I'll defer to you on what's ultimately best for the project; happy to follow your lead here. 😃 Note: If we *do* go with the generic `CslStringList` parameter version, I'll probably remove `MakeValidOpts` until we come up with a consistent approach to string-list-option construction.

metasim: PS: I'll fix docs and move this out of draft once we settle on a design.

metasim: PS: Just re-read this, and maybe should have drawn attention earlier to the fact that GDAL already has `OGR_G_MakeValid` and `OGR_G_MakeValidEx` variants. So our doubling with `Ex` functions in this case wouldn't technically be making up a new one.

> I'm inclined to prefer single functions instead of pairs like these. Note that GDAL already has some `Ex` functions, so us making up new ones might end up being quite confusing. But I guess that's not such a great risk, GDAL would have to add a new parameter that can't be expressed as an existing `CSLStringList`.

metasim: @lnicola After sleeping on it, I'm now leaning toward this being the API, given the mirroring of `OGR_G_MakeValid`, and `OGR_G_MakeValidEx`.

```
pub fn make_valid(&self) -> Result<Geometry>
```
and
```
pub fn make_valid_ex(&self, opts: &CslStringList) -> Result<Geometry>
```

For now I'll drop `MakeValidOpts`, but possibly consider in the future a separate module of types for idiomatically constructing valid argument combinations and converting them to `CslStringList`, for methods that accept them.

WDYT?

lnicola: Sorry, I'll try to take another look tomorrow.

metasim: > Yeah, but the C++ API only has

huh... you know, until now I just assumed the C API was the core with C++ built on top of that, and not the other way around! 🤦

lnicola: Nope, GDAL is C++, but it exposes a C API. I don't know if the C++ one is actually public.

metasim: @jdroenner @rmanoka @ChristianBeilschmidt or anyone else: do you have opinions about the API approach we should take now an moving forward when optional arguments are specified with `CslStringList`? We're trying to decide a consistent approach as we expose more functions that use a list of `"KEY=VALUE"` strings to specify options.

jdroenner: i think the `O: Into<X> ` pattern is really nice in many cases. However, in this case i think we can go with `Option<X>` or  `CslStringList::default()`. Since it is a list an option is somehow redundant. but maybe more ergonomic.

lnicola: @jdroenner does that trade-off change for you once we bring in `MakeValidOpts`? `make_valid(MakeValidOpts::Structure)` looks a bit nicer than `make_valid(MakeValidOpts::Structure.into())`.

src/vector/vector_tests/mod.rs
```diff
@@ -826,14 +827,17 @@ mod tests {
...
+                let _nolog = SuppressGDALErrorLog::new();
+                // force error
+                assert!(matches!(
+                    layer.set_attribute_filter("foo = bar").unwrap_err(),
+                    GdalError::OgrError {
+                        err: gdal_sys::OGRErr::OGRERR_CORRUPT_DATA,
+                        method_name: "OGR_L_SetAttributeFilter",
+                    }
+                ));
+            }
```
metasim, line 840: PR scope creep... I got tired of seeing this error in test output.

src/cpl.rs
```diff
@@ -155,6 +166,47 @@ impl Debug for CslStringList {
...
+///     opts.into().len()
+/// }
+///
+/// assert_eq!(count_opts(&[("One", "1"), ("Two", "2"), ("Three", "3")]), 3);
+/// ```
+impl<const N: usize> From<&[(&str, &str); N]> for CslStringList {
+    fn from(pairs: &[(&str, &str); N]) -> Self {
+        let mut result = Self::default();
+        for (k, v) in pairs {
+            result.set_name_value(k, v).expect("valid key/value pair");
```
lnicola, line 204:
```suggestion
            result.set_name_value(k, v).expect("invalid key/value pair");
```

I think the message would be something like `valid key/value pair: "Invalid characters in name: '#'"`, which doesn't make sense.

And going back a little, I'm not sure what to say about this impl. `from` should be infallible, so `try_from` would be better, which this clearly isn't. On the other hand, it's probably going to be used with compile-constant keys and values. Is that why you've implemented it for fixed-size arrays (as opposed to `&[(&str, &str)]`)?

src/vector/geometry.rs
```diff
@@ -416,6 +454,61 @@ impl Geometry {
...
+    ///
+    /// When GEOS < 3.8, this method will return `Ok(self.clone())` if it is valid, or `Err` if not.
+    ///
+    /// Refer: [OGR_G_MakeValidEx](https://gdal.org/api/vector_c_api.html#_CPPv417OGR_G_MakeValidEx12OGRGeometryH12CSLConstList)
+    ///
+    /// # Example
+    /// ```rust, no_run
+    /// use gdal::vector::Geometry;
+    /// # fn main() -> gdal::errors::Result<()> {
+    /// let src = Geometry::from_wkt("POLYGON ((0 0,10 10,0 10,10 0,0 0))")?;
```
lnicola, line 481: (nit) these are probably nicer to read as below, but you don't need to change them:

```suggestion
    /// let src = Geometry::from_wkt("POLYGON ((0 0, 10 10, 0 10, 10 0, 0 0))")?;
```

lnicola (changes requested):
The error handling changes seem like a really bad idea.

I'm also unsure about the `From` implementation for `CslStringList`, but that's probably less important. From (sic!) [the docs](https://doc.rust-lang.org/std/convert/trait.From.html):

> Note: This trait must not fail. The From trait is intended for perfect conversions. If the conversion can fail or is not perfect, use [TryFrom](https://doc.rust-lang.org/std/convert/trait.TryFrom.html).

Other than that, LGTM.

src/cpl.rs
metasim: > Is that why you've implemented it for fixed-size arrays

No. More of a result of me playing type golf; Without the `const` parameter I get this error. Probably me being dumb.

image


The fact that `set_name_value` is fallible is really annoying, but understandable. Going with `try_from` pretty much eliminates the benefits of having this. So if you're in doubt about it, I think I should just remove it all together.

src/vector/layer.rs
```diff
@@ -251,7 +251,8 @@ pub trait LayerAccess: Sized {
     }
 
     /// Set a feature on this layer layer.
-    /// Refer[SetFeature](https://gdal.org/doxygen/classOGRLayer.html#a681139bfd585b74d7218e51a32144283)
+    ///
+    /// Refer [SetFeature](https://gdal.org/doxygen/classOGRLayer.html#a681139bfd585b74d7218e51a32144283)
```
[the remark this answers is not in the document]
metasim: I don't like `Refer` either... but was trying to be consistent (but forgot the colon here).

metasim:
> The error handling changes seem like a really bad idea.

I detailed my rationale [here](https://github.com/georust/gdal/pull/356#discussion_r1066216433).  Only doing it because a) I believe successful tests should not report error messages, and b) according to the gdal docs, the implementation should be thread-safe.

src/vector/geometry.rs
```diff
@@ -482,13 +575,14 @@ impl Debug for GeometryRef<'_> {
...
+    use super::*;
     use crate::spatial_ref::SpatialRef;
-
-    use super::{geometry_type_to_name, Geometry};
+    use crate::test_utils::SuppressGDALErrorLog;
 
     #[test]
     #[allow(clippy::float_cmp)]
     pub fn test_area() {
+        let _nolog = SuppressGDALErrorLog::new();
```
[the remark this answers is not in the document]
lnicola: Ah, I missed it being thread-local. Should be fine, then 👍.

metasim:
> I'm also unsure about the `From` implementation for `CslStringList`, but that's probably less important. 

I'm also unsure about this. Maybe I should just drop this `impl` until we figure out a more ergonomic way to construct static `CslStringList`s?

src/vector/geometry.rs
metasim, line 585: Without this specific suppression, the log message is

> Warning 1: OGR_G_Area() called against non-surface geometry type.

src/cpl.rs
lnicola: Oof, yes, I think you need to cast it, like in https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0c8e96e6b31a1d7ac28a76cce7abbd1b. Anyway, I suppose it's fine.

lnicola:
We could also implement TryFrom, which seems like a good thing to have anyway.

bors d+

src/cpl.rs
metasim: > I think the message would be something like `valid key/value pair: "Invalid characters in name: '#'"`, which doesn't make sense.


Are you 100% sure? (I'm not) 😄 

Per [the docs](https://doc.rust-lang.org/core/result/enum.Result.html#recommended-message-style)

> We recommend that expect messages are used to describe the reason you expect the Result should be Ok.

metasim:
@lnicola Thanks for slogging through this with me! I'm really happy with the result. Wouldn't be half as good without your feedback! 😃

src/cpl.rs
lnicola: Can you test? 😀 I can't right now, sorry.

My reading of the docs is something like `CslStringList::try_from(&["FOO", "BAR"]).expect("obviously correct")`, that is, user code. It's different in library code which can and will actually panic.

metasim: I'm not sure I'm following your line of thinking here. Happy to submit a new PR revisiting this.

lnicola: Indeed, this produces:

```
thread 'cpl::tests::from_impl' panicked at 'valid key/value pair: BadArgument("Invalid characters in name: 'TWO#'")', src/cpl.rs:204:41
```

compared to a random example from `libcore`:

```
self.checked_add(rhs).expect("overflow when adding durations")
```

PR #359, opened 2023-01-07
Adds c_rasterband method to Rasterband

Gert-JanPeeters:
---


lnicola:
Can you add a changelog entry? The users might be interested in knowing about this.

bors d+

PR #362, opened 2023-01-27
Replace `CslStringList` `From` implementation with `TryFrom`

lnicola:
---

This isn't _that_ ugly, and I think it's better than panicking in a `From` implementation.

### DIFF
--- a/src/cpl.rs
+++ b/src/cpl.rs
@@ -164,44 +164,25 @@ impl Debug for CslStringList {
     }
 }
 
-/// Convenience shorthand for specifying an empty `CslStringList` to functions accepting
-/// `Into<CslStringList>`.
-///
-/// # Example
-///
-/// ```rust, no_run
-/// use gdal::cpl::CslStringList;
-/// fn count_opts<O: Into<CslStringList>>(opts: O) -> usize {
-///     opts.into().len()
-/// }
-///
-/// assert_eq!(count_opts(()), 0);
-/// ```
-impl From<()> for CslStringList {
-    fn from(_: ()) -> Self {
-        CslStringList::default()
-    }
-}
-
 /// Convenience for creating a [`CslStringList`] from a slice of _key_/_value_ tuples.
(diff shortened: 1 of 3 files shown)


src/cpl.rs
```diff
@@ -166,44 +166,25 @@ impl Debug for CslStringList {
...
-///
-/// ```rust, no_run
-/// use gdal::cpl::CslStringList;
-/// fn count_opts<O: Into<CslStringList>>(opts: O) -> usize {
-///     opts.into().len()
-/// }
-///
-/// assert_eq!(count_opts(()), 0);
-/// ```
-impl From<()> for CslStringList {
```
metasim, line 182: 😢

src/vector/geometry.rs
```diff
@@ -692,7 +688,7 @@ mod tests {
     /// Simple clone case.
     pub fn test_make_valid_clone() {
         let src = Geometry::from_wkt("POINT (0 0)").unwrap();
-        let dst = src.make_valid(());
+        let dst = src.make_valid(&CslStringList::new());
```
metasim, line 691: 😢

metasim (approved):
Looks like 1.67 came with new clippy lints...

metasim:
bors d+

lnicola:
There's another PR for the lints.

src/cpl.rs
lnicola: This isn't so bad, is it? I mean, it's `from(&[])` vs. `from(())`..

metasim: Nah, I'm just being silly.  I was bothered by the verbosity of `src.make_valid(&CslStringList::new())` (vs. `src.make_valid(())`), but fell into being overly clever and not really common/idiomatic. I understand why you got rid of it.

lnicola:
bors d=metasim

lnicola:
bors r=metasim

PR #364, opened 2023-02-05
Add CslStringList::add_string

jshrake:
---

This method wraps gdal_sys::CSLAddString.

This method is useful when constructing argument lists for option structs for the various utility functions found in gdal_utils, such as GDALWarp and GDALTranslate.

### DIFF
--- a/src/cpl.rs
+++ b/src/cpl.rs
@@ -7,7 +7,9 @@ use std::ffi::CString;
 use std::fmt::{Debug, Formatter};
 use std::ptr;
 
-use gdal_sys::{CSLCount, CSLDestroy, CSLDuplicate, CSLFetchNameValue, CSLSetNameValue};
+use gdal_sys::{
+    CSLAddString, CSLCount, CSLDestroy, CSLDuplicate, CSLFetchNameValue, CSLSetNameValue,
+};
 use libc::c_char;
 
 use crate::errors::{GdalError, Result};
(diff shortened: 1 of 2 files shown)


src/cpl.rs
```diff
@@ -57,6 +59,17 @@ impl CslStringList {
         Ok(())
     }
 
+    /// Adds the string `value` to the list.
+    ///
+    /// Returns `Ok<()>` on success, `Err<GdalError>` if `value` cannot be converted to a C string.
```
metasim, line 64:
```suggestion
    /// Returns `Ok<()>` on success, `Err<GdalError>` if `value` cannot be converted to a C string.
    ///
    /// See [`CSLAddString`](https://gdal.org/api/cpl.html#_CPPv412CSLAddStringPPcPKc)

```

metasim (commented):
LGTM. @lnicola?

src/cpl.rs
```diff
@@ -57,6 +59,20 @@ impl CslStringList {
...
+    ///
+    /// See: [`CSLAddString`](https://gdal.org/api/cpl.html#_CPPv412CSLAddStringPPcPKc)
+
+    pub fn add_string(&mut self, value: &str) -> Result<()> {
+        let v = CString::new(value)?;
+        unsafe {
+            self.list_ptr = CSLAddString(self.list_ptr, v.as_ptr());
```
lnicola, line 71: Please move the assignment out of the `unsafe` block (move it after the `=`), it's easier to see what exactly is unsafe that way.

And if it's not too hard, can you squash at the end?

lnicola:
There is already an easier way to do this (see the example for `Geometry::make_valid`), but it seems reasonable to expose this too.

lnicola (approved):

metasim:
> There is already an easier way to do this (see the example for `Geometry::make_valid`), but it seems reasonable to expose this too.

Agreed. I'm of multiple minds about `CslStringList` as it provides so many ways to mis-specify settings. But I suppose it's not the job of this library to work around _all_ of GDAL's legacy shortcomings.

lnicola:
We can add type-safe wrappers for those, but it's pretty hard to catch and keep up with GDAL. Think of all those create and open options it has.

src/cpl.rs
jshrake: Done and done!

metasim:
Nice work @jshrake !

PR #370, opened 2023-02-11
Added `envelope*` methods to `Geometry`.

metasim:
---

### DIFF
--- a/src/vector/geometry.rs
+++ b/src/vector/geometry.rs
@@ -2,6 +2,7 @@ use std::cell::RefCell;
 use std::ffi::CString;
 use std::fmt::{self, Debug, Formatter};
 use std::marker::PhantomData;
+use std::mem::MaybeUninit;
 use std::ops::Deref;
 use std::ptr::null_mut;
 
@@ -13,6 +14,7 @@ use gdal_sys::{self, OGRErr, OGRGeometryH, OGRwkbGeometryType};
 use crate::errors::*;
 use crate::spatial_ref::{CoordTransform, SpatialRef};
 use crate::utils::{_last_null_pointer_err, _string};
+use crate::vector::{Envelope, Envelope3D};
 
 /// OGR Geometry
 pub struct Geometry {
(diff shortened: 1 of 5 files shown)


gdal-sys/src/lib.rs
```diff
@@ -6,3 +6,27 @@
...
+        Self {
+            MinX: 0.0,
+            MaxX: 0.0,
+            MinY: 0.0,
+            MaxY: 0.0,
+            MinZ: 0.0,
+            MaxZ: 0.0,
+        }
+    }
+}
```
metasim, line 32: Is this controversial? Can we get bindgen to do it for us (on selected structs)?

src/vector/geometry.rs
```diff
@@ -424,10 +425,30 @@ impl Geometry {
         Ok(unsafe { Geometry::with_c_geometry(new_c_geom, true) })
     }
 
+    /// Compute geometry area in square units of the spatial reference system in use.
+    ///
+    /// Supported for on`LinearRing`, `Polygon` or `MultiPolygon`.
```
lnicola, line 430:
```suggestion
    /// Supported for `LinearRing`, `Polygon` or `MultiPolygon`.
```

src/vector/geometry.rs
```diff
@@ -424,10 +425,30 @@ impl Geometry {
...
+    ///
+    /// See: [`OGR_G_Area`](https://gdal.org/api/vector_c_api.html#_CPPv410OGR_G_Area12OGRGeometryH)
     pub fn area(&self) -> f64 {
         unsafe { gdal_sys::OGR_G_Area(self.c_geometry()) }
     }
 
+    /// Computes and returns the axis-aligned 2D bounding envelope for this geometry.
+    pub fn envelope(&self) -> Envelope {
+        let mut envelope = Envelope::default();
+        unsafe { gdal_sys::OGR_G_GetEnvelope(self.c_geometry(), &mut envelope) };
```
lnicola, line 441: Nit: do you want to give https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html a try here?

gdal-sys/src/lib.rs
lnicola: I don't think so, looks good to me.

src/vector/geometry.rs
metasim: Do you mean using `MaybeUninit` instead of `Default`, and skip the initialization?

lnicola: Yes. It's not very important, just slightly nice to skip the initialization.

lnicola (approved):
The `MaybeUninit` uses look good, will merge in a day or so.

src/vector/geometry.rs
```diff
@@ -424,10 +426,38 @@ impl Geometry {
...
+    /// See: [`OGR_G_GetEnvelope`](https://gdal.org/api/vector_c_api.html#_CPPv417OGR_G_GetEnvelope12OGRGeometryHP11OGREnvelope)
+    pub fn envelope(&self) -> Envelope {
+        let mut envelope = MaybeUninit::uninit();
+        unsafe {
+            gdal_sys::OGR_G_GetEnvelope(self.c_geometry(), envelope.as_mut_ptr());
+            envelope.assume_init()
+        }
+    }
+
+    /// Computes and returns the axis aligned 3D bounding envelope for this geometry.\
```
lnicola, line 450:
```suggestion
    /// Computes and returns the axis aligned 3D bounding envelope for this geometry.
```

src/vector/geometry.rs
```diff
@@ -424,10 +426,38 @@ impl Geometry {
...
+    ///
+    /// See: [`OGR_G_GetEnvelope3D`](https://gdal.org/api/vector_c_api.html#_CPPv419OGR_G_GetEnvelope3D12OGRGeometryHP13OGREnvelope3D)
+    pub fn envelope3d(&self) -> Envelope3D {
```
lnicola, line 453: Nit:

```suggestion
    pub fn envelope_3d(&self) -> Envelope3D {
```

Not sure about this one, what do you think?

PR #371, opened 2023-02-11
Refactoring `vector` module for better maintainability and extensibility.

metasim:
- [NA] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Refactoring to improve maintainability of vector module.

Changes include:
* Grouping methods of similar purpose into separate files
* Putting associated tests in the same file as the implementation
* Adding or expanding documentation on existing methods
* Normalizing documentation patterns and format.

Intent was to not have any breaking changes.


lnicola (approved):
LGTM overall, but I can't say for sure :sweat_smile:.

One naming nit: should we use something like `set` instead of `binary`? And `formats` (:bike: :house:) instead of `text`?

metasim:
> One naming nit

Not a nit... I've already renamed those 5 times. 🙈

jdroenner (approved):

metasim:
@lnicola @jdroenner I'm going to merge this, and then in a separate PR implement `union`, `difference`, `symmetric_difference`, etc. in the new `set` module.

lnicola:
> [NA] I added an entry to CHANGES.md if knowledge of this change could be valuable to users.

image

Anyway, we can probably merge this.

bors d+

lnicola:
Actually, maybe there's no breaking changes?

bors r+

PR #372, opened 2023-02-15
Implement CoordTransformOptions and CoordTransform::new_with_options constructor

ttencate:
---

I only needed this for the ballpark option, because I found it surprising that it's possible by default to transform between CRSes that are only valid on opposite sides of the world. I didn't change this default though.

The other three options are wrapped, but I don't understand them well enough to add tests for them.

### DIFF
--- a/src/spatial_ref/srs.rs
+++ b/src/spatial_ref/srs.rs
@@ -7,6 +7,145 @@ use std::str::FromStr;
 
 use crate::errors::*;
 
+/// Options for [`CoordTransform::new_with_options`].
+#[derive(Debug)]
+pub struct CoordTransformOptions {
+    inner: gdal_sys::OGRCoordinateTransformationOptionsH,
+}
+
+impl Drop for CoordTransformOptions {
+    fn drop(&mut self) {
+        unsafe { gdal_sys::OCTDestroyCoordinateTransformationOptions(self.inner) };
+    }
+}
+
+impl CoordTransformOptions {
+    /// Creation options for [`CoordTransform`].
+    pub fn new() -> Result<CoordTransformOptions> {
+        let c_obj = unsafe { gdal_sys::OCTNewCoordinateTransformationOptions() };
+        if c_obj.is_null() {
+            return Err(_last_null_pointer_err(
+                "OCTNewCoordinateTransformationOptions",
+            ));
+        }
+        Ok(CoordTransformOptions { inner: c_obj })
+    }
+
+    /// Sets an area of interest.
+    ///
+    /// The west longitude is generally lower than the east longitude, except for areas of interest
+    /// that go across the anti-meridian.
+    ///
+    /// For more information, see:
+    /// <https://gdal.org/tutorials/osr_api_tut.html#advanced-coordinate-transformation>
+    ///
+    /// # Arguments
+    ///
+    /// - `west_longitude_deg` – West longitude (in degree). Must be in [-180,180]
+    /// - `south_latitude_deg` – South latitude (in degree). Must be in [-90,90]
+    /// - `east_longitude_deg` – East longitude (in degree). Must be in [-180,180]
+    /// - `north_latitude_deg` – North latitude (in degree). Must be in [-90,90]
+    pub fn set_area_of_interest(
+        &mut self,
+        west_longitude_deg: f64,
+        south_latitude_deg: f64,
+        east_longitude_deg: f64,
+        north_latitude_deg: f64,
+    ) -> Result<()> {
+        let ret_val = unsafe {
+            gdal_sys::OCTCoordinateTransformationOptionsSetAreaOfInterest(
+                self.inner,
+                west_longitude_deg,
+                south_latitude_deg,
+                east_longitude_deg,
+                north_latitude_deg,
+            )
+        };
+        if ret_val == 0 {
+            return Err(_last_cpl_err(CPLErr::CE_Failure));
+        }
+        Ok(())
+    }
+
+    /// Sets the desired accuracy for coordinate operations.
+    ///
+    /// Only coordinate operations that offer an accuracy of at least the one specified will be
+    /// considered.
+    ///
+    /// An accuracy of 0 is valid and means a coordinate operation made only of one or several
+    /// conversions (map projections, unit conversion, etc.) Operations involving ballpark
+    /// transformations have a unknown accuracy, and will be filtered out by any dfAccuracy >= 0
+    /// value.
+    ///
+    /// If this option is specified with PROJ < 8, the `OGR_CT_OP_SELECTION` configuration option
+    /// will default to `BEST_ACCURACY`.
+    #[cfg(any(major_ge_4, all(major_ge_3, minor_ge_3)))]
+    pub fn desired_accuracy(&mut self, accuracy: f64) -> Result<()> {
+        let ret_val = unsafe {
+            gdal_sys::OCTCoordinateTransformationOptionsSetDesiredAccuracy(self.inner, accuracy)
+        };
+        if ret_val == 0 {
+            return Err(_last_cpl_err(CPLErr::CE_Failure));
+        }
+        Ok(())
+    }
+
+    /// Sets whether ballpark transformations are allowed.
+    ///
+    /// By default, PROJ may generate "ballpark transformations" (see
+    /// <https://proj.org/glossary.html>) when precise datum transformations are missing. For high
+    /// accuracy use cases, such transformations might not be allowed.
+    ///
+    /// If this option is specified with PROJ < 8, the `OGR_CT_OP_SELECTION` configuration option
+    /// will default to `BEST_ACCURACY`.
+    #[cfg(any(major_ge_4, all(major_ge_3, minor_ge_3)))]
+    pub fn set_ballpark_allowed(&mut self, ballpark_allowed: bool) -> Result<()> {
+        let ret_val = unsafe {
+            gdal_sys::OCTCoordinateTransformationOptionsSetBallparkAllowed(
+                self.inner,
+                ballpark_allowed as libc::c_int,
+            )
+        };
+        if ret_val == 0 {
+            return Err(_last_cpl_err(CPLErr::CE_Failure));
+        }
+        Ok(())
+    }
+
+    /// Sets a coordinate operation.
+    ///
+    /// This is a user override to be used instead of the normally computed pipeline.
+    ///
+    /// The pipeline must take into account the axis order of the source and target SRS.
+    ///
+    /// The pipeline may be provided as a PROJ string (single step operation or multiple step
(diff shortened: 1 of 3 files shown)


src/spatial_ref/srs.rs
```diff
@@ -7,6 +7,146 @@ use std::str::FromStr;
...
+    ///
+    /// This is a user override to be used instead of the normally computed pipeline.
+    ///
+    /// The pipeline must take into account the axis order of the source and target SRS.
+    ///
+    /// The pipeline may be provided as a PROJ string (single step operation or multiple step
+    /// string starting with +proj=pipeline), a WKT2 string describing a CoordinateOperation, or a
+    /// "urn:ogc:def:coordinateOperation:EPSG::XXXX" URN.
+    ///
+    /// Parameters:
```
metasim, line 128: nit:

```suggestion
    /// # Arguments
```

src/spatial_ref/srs.rs
```diff
@@ -7,6 +7,146 @@ use std::str::FromStr;
...
+        }
+        Ok(CoordTransformOptions { inner: c_obj })
+    }
+
+    /// Sets an area of interest.
+    ///
+    /// The west longitude is generally lower than the east longitude, except for areas of interest
+    /// that go across the anti-meridian.
+    ///
+    /// Parameters:
```
metasim, line 40: Nit:
```suggestion
    /// # Arguments
```

src/spatial_ref/srs.rs
```diff
@@ -7,6 +7,146 @@ use std::str::FromStr;
...
+    /// Only coordinate operations that offer an accuracy of at least the one specified will be
+    /// considered.
+    ///
+    /// An accuracy of 0 is valid and means a coordinate operation made only of one or several
+    /// conversions (map projections, unit conversion, etc.) Operations involving ballpark
+    /// transformations have a unknown accuracy, and will be filtered out by any dfAccuracy >= 0
+    /// value.
+    ///
+    /// If this option is specified with PROJ < 8, the OGR_CT_OP_SELECTION configuration option
+    /// will default to BEST_ACCURACY.
```
metasim, line 80: nit: 
```suggestion
    /// If this option is specified with PROJ < 8, the `OGR_CT_OP_SELECTION` configuration option
    /// will default to `BEST_ACCURACY`.
```

src/spatial_ref/srs.rs
```diff
@@ -7,6 +7,146 @@ use std::str::FromStr;
...
+    }
+
+    /// Sets whether ballpark transformations are allowed.
+    ///
+    /// By default, PROJ may generate "ballpark transformations" (see
+    /// <https://proj.org/glossary.html>) when precise datum transformations are missing. For high
+    /// accuracy use cases, such transformations might not be allowed.
+    ///
+    /// If this option is specified with PROJ < 8, the OGR_CT_OP_SELECTION configuration option
+    /// will default to BEST_ACCURACY.
```
metasim, line 101: nit:
```suggestion
    /// If this option is specified with PROJ < 8, the `OGR_CT_OP_SELECTION` configuration option
    /// will default to `BEST_ACCURACY`.
```

src/spatial_ref/srs.rs
```diff
@@ -7,6 +7,146 @@ use std::str::FromStr;
...
+
+    /// Sets a coordinate operation.
+    ///
+    /// This is a user override to be used instead of the normally computed pipeline.
+    ///
+    /// The pipeline must take into account the axis order of the source and target SRS.
+    ///
+    /// The pipeline may be provided as a PROJ string (single step operation or multiple step
+    /// string starting with +proj=pipeline), a WKT2 string describing a CoordinateOperation, or a
+    /// "urn:ogc:def:coordinateOperation:EPSG::XXXX" URN.
```
metasim, line 126: Is there a canonical description on these operation definitions you could link to?

src/spatial_ref/tests.rs
```diff
@@ -245,6 +245,25 @@ fn failing_transformation() {
...
+    let trafo = CoordTransform::new(&ma, &nl);
+    assert!(trafo.is_ok());
+
+    let mut options = CoordTransformOptions::new().unwrap();
+    options.set_ballpark_allowed(false).unwrap();
+    let trafo = CoordTransform::new_ex(&ma, &nl, &options);
+    let err = trafo.unwrap_err();
+    assert!(matches!(err, GdalError::NullPointer { .. }), "{:?}", err);
+}
+
```
metasim, line 266: Would it be viable to provide a test for `set_coordinate_operation`?

metasim (commented):
A few nits, but overall LGTM. However, I suggest a review from one of the more experienced team members.

src/spatial_ref/srs.rs
ttencate: I have absolutely no idea what any of this even means, I just copied the docs from upstream GDAL.

src/spatial_ref/tests.rs
ttencate: If somebody could tell me input and expected output, sure. I only implemented it for completeness. If it's not cool to have untested (but also rarely used!) code in there, I'm also happy to remove it again.

src/spatial_ref/srs.rs
```diff
@@ -7,6 +7,148 @@ use std::str::FromStr;
...
+/// Options for [`CoordTransform::new_ex`].
+#[derive(Debug)]
+pub struct CoordTransformOptions {
+    inner: gdal_sys::OGRCoordinateTransformationOptionsH,
+}
+
+impl Drop for CoordTransformOptions {
+    fn drop(&mut self) {
+        unsafe { gdal_sys::OCTDestroyCoordinateTransformationOptions(self.inner) };
+        self.inner = ptr::null_mut();
```
lnicola, line 19: No need to do this AFAIK.

src/spatial_ref/srs.rs
```diff
@@ -7,6 +7,148 @@ use std::str::FromStr;
...
+            gdal_sys::OCTCoordinateTransformationOptionsSetAreaOfInterest(
+                self.inner,
+                west_longitude_deg,
+                south_latitude_deg,
+                east_longitude_deg,
+                north_latitude_deg,
+            )
+        };
+        if ret_val == 0 {
+            return Err(GdalError::OctError {
```
lnicola, line 63: I'm not sure how our errors are supposed to work, but this function returns an error message, e.g. `Invalid dfWestLongitudeDeg`. You can get it using `utils::_last_cpl_err`, but of course, that will return a `CplError` instead. So I don't really know what would be the best.

src/spatial_ref/srs.rs
```diff
@@ -7,6 +7,148 @@ use std::str::FromStr;
...
+    /// considered.
+    ///
+    /// An accuracy of 0 is valid and means a coordinate operation made only of one or several
+    /// conversions (map projections, unit conversion, etc.) Operations involving ballpark
+    /// transformations have a unknown accuracy, and will be filtered out by any dfAccuracy >= 0
+    /// value.
+    ///
+    /// If this option is specified with PROJ < 8, the `OGR_CT_OP_SELECTION` configuration option
+    /// will default to `BEST_ACCURACY`.
+    #[cfg(all(major_ge_3, minor_ge_3))]
```
lnicola, line 82: This is wrong, it will fail in `4.0`. We want `major > 3 || major == 3 && minor >= 3`. There are some places where we do this correctly.

But honestly, I wish we could improve on these somehow.

src/spatial_ref/srs.rs
```diff
@@ -7,6 +7,148 @@ use std::str::FromStr;
...
+    }
+
+    /// Sets a coordinate operation.
+    ///
+    /// This is a user override to be used instead of the normally computed pipeline.
+    ///
+    /// The pipeline must take into account the axis order of the source and target SRS.
+    ///
+    /// The pipeline may be provided as a PROJ string (single step operation or multiple step
+    /// string starting with +proj=pipeline), a WKT2 string describing a CoordinateOperation, or a
```
lnicola, line 126:
```suggestion
    /// string starting with `+proj=pipeline`), a WKT2 string describing a `CoordinateOperation`, or a
```

src/spatial_ref/srs.rs
```diff
@@ -33,6 +175,24 @@ impl CoordTransform {
         })
     }
 
+    pub fn new_ex(
+        sp_ref1: &SpatialRef,
```
lnicola, line 179: Maybe something like

```suggestion
        source: &SpatialRef,
```

(and `target` below)

src/spatial_ref/srs.rs
```diff
@@ -33,6 +175,24 @@ impl CoordTransform {
         })
     }
 
+    pub fn new_ex(
```
lnicola, line 178:
```suggestion
    pub fn new_with_options(
```

?

src/spatial_ref/tests.rs
lnicola: I think we could call it just to check it doesn't crash.

ttencate: Added a test.

src/spatial_ref/srs.rs
ttencate: Thought so. Smelled like a C habit; I just cargo-culted it. Removed it from `CoordTransform` as well.

src/spatial_ref/srs.rs
ttencate: Done. Yeah, definitely not the only occurrence of this mistake in the codebase, or even in this file.

Merging major and minor into a single `gdal_ge_3_3` option would make these lines shorter, at the expense of adding a bunch of (autogenerated) cfg options.

A custom proc macro with custom syntax could of course also work.

src/spatial_ref/srs.rs
ttencate: Done.

src/spatial_ref/srs.rs
ttencate: I found https://gdal.org/tutorials/osr_api_tut.html#advanced-coordinate-transformation which sheds some light. I don't think our API docs are the right place to expand this fully, so I added a link.

src/spatial_ref/srs.rs
ttencate: Done, but note that it's inconsistent with `Dataset::new_ex`. That one was named after `GDALOpenEx`, however, so there's some argument to be made for the inconsistency.

src/spatial_ref/srs.rs
ttencate: Good call. Done.

src/spatial_ref/srs.rs
ttencate: I don't know why this OCT prefix is even used here (OGR Coordinate Transformation??), so it seems excessive to add another error type for it. `CplError` will do fine for now. Used this for the other setters as well, though these are (currently) infallible.

PR #377, opened 2023-03-04
link and version

jdroenner:
---

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,7 +1,8 @@
 use std::{env, str::FromStr};
 
 fn main() {
-    let gdal_version_string = env::var("DEP_GDAL_VERSION_NUMBER").unwrap();
+    let gdal_version_string = env::var("DEP_GDAL_VERSION_NUMBER")
+        .expect("The GDAL-SYS crate must emit the version of libgdal via cargo:version_number");
     println!("GDAL version string: \"{gdal_version_string}\"");
 
     // this version string is the result of:
--- a/build.rs
+++ b/build.rs
@@ -1,11 +1,7 @@
-use std::str::FromStr;
-
-fn gdal_version_info() -> String {
-    gdal_sys::gdal_version_docs_rs_wrapper()
-}
+use std::{env, str::FromStr};
 
 fn main() {
-    let gdal_version_string = gdal_version_info();
+    let gdal_version_string = env::var("DEP_GDAL_VERSION_NUMBER").unwrap();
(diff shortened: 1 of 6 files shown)


lnicola (approved):
I wonder if we shouldn't emit something like `9090900` in the docs.rs case, so everything gets enabled.

build.rs
```diff
@@ -1,11 +1,7 @@
-use std::str::FromStr;
-
-fn gdal_version_info() -> String {
-    gdal_sys::gdal_version_docs_rs_wrapper()
-}
+use std::{env, str::FromStr};
 
 fn main() {
-    let gdal_version_string = gdal_version_info();
+    let gdal_version_string = env::var("DEP_GDAL_VERSION_NUMBER").unwrap();
```
metasim, line 4: Though very unlikely, consider using `.expect(...)` with a helpful message to indicate it was expected that `gdal-sys/build.rs` would have emitted `cargo:version-number`.

gdal-sys/build.rs
```diff
@@ -59,45 +59,24 @@ fn find_gdal_dll(lib_dir: &Path) -> io::Result<Option<String>> {
...
-                let res_ptr = crate::GDALVersionInfo(c_key.as_ptr());
-                let c_res = std::ffi::CStr::from_ptr(res_ptr);
-                c_res.to_string_lossy().into_owned()
-            }
-        }"#
-        .to_string()
-    };
-    std::fs::write(out_path, docs_helper_code.as_bytes()).unwrap();
-}
-
```
metasim, line 87: ❤️

gdal-sys/src/lib.rs
```diff
@@ -5,4 +5,3 @@
 #![allow(rustdoc::bare_urls)]
 
 include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
-include!(concat!(env!("OUT_DIR"), "/docs_rs_helper.rs"));
```
metasim, line 8: This must have been satisfying to eliminate! 😄

metasim (approved):

build.rs
jdroenner: [4e263c9](https://github.com/georust/gdal/pull/377/commits/4e263c94e97ca2cf1f2d468766125e9531178012)

gdal-sys/src/lib.rs
jdroenner: yes :D

PR #379, opened 2023-03-10
add union operation for geometry

zredb:
---
This is my first time to commit code to this repository. I see that vecotr only has an intersection operation at present, so I try to add a union operation to it. If it can be accepted, I will add other operations in the same way later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 GPATH
 GTAGS
 GRTAGS
+/setenv.bat
--- a/src/vector/ops/set.rs
+++ b/src/vector/ops/set.rs
@@ -32,6 +32,22 @@ impl Geometry {
         }
         Some(unsafe { Geometry::with_c_geometry(ogr_geom, true) })
     }
+
+    pub fn union(&self, other: &Self) -> Option<Self> {
+        if !self.has_gdal_ptr() {
+            return None;
+        }
+        if !other.has_gdal_ptr() {
+            return None;
+        }
+        unsafe {
+            let ogr_geom = gdal_sys::OGR_G_Union(self.c_geometry(), other.c_geometry());
+            if ogr_geom.is_null() {
+                return None;
+            }
+            Some(Geometry::with_c_geometry(ogr_geom, true))
+        }
+    }
 }
 
 #[cfg(test)]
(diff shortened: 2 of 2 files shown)


metasim (changes requested):
@zredb Thank you for the contribution!
Could you please add documentation to `union`, similar to `intersection`?

metasim:
Also, I'm not seeing an update to `CHANGES.md`.

src/vector/ops/set.rs
```diff
@@ -32,6 +32,22 @@ impl Geometry {
...
+        if !self.has_gdal_ptr() {
+            return None;
+        }
+        if !other.has_gdal_ptr() {
+            return None;
+        }
+        unsafe {
+            let ogr_geom = gdal_sys::OGR_G_Union(self.c_geometry(), other.c_geometry());
+            if ogr_geom.is_null() {
+                return None;
```
lnicola, line 46: I haven't tested, but it might be possible to return a proper error here?

But then I'm not sure what we should do in the `!has_gdal_ptr()` case...

src/vector/ops/set.rs
```diff
@@ -84,4 +100,50 @@ mod tests {
...
+                .unwrap();
+        let other = Geometry::from_wkt("POLYGON ((1 -5, 1 1, -5 1, -5 -5, 1 -5))").unwrap();
+
+        let inter = geom.union(&other);
+
+        assert!(inter.is_some());
+
+        let inter = inter.unwrap();
+
+        assert_eq!(inter.area(), 135.0);
```
lnicola, line 118: I think we have some `assert_almost_eq` helper somewhere.

.gitignore
```diff
@@ -8,3 +8,4 @@
 GPATH
 GTAGS
 GRTAGS
+/setenv.bat
```
lnicola, line 11: Do we want this? In the long run every contributor might end up with a script like that, but with a different name.

src/vector/ops/set.rs
zredb: According to gdal's document, when an error occurs, the return value is null, so here we return None, which seems to be consistent with gdal.

.gitignore
zredb: Setenv is just my own configuration. I have removed it from gitignore.

src/vector/ops/set.rs
lnicola: Yeah, but in most cases GDAL will set an error code and message like [here](https://github.com/OSGeo/gdal/blob/master/ogr/ogrgeometry.cpp#L4531), even without documenting it explicitly.

But indeed, this doesn't seem to be done in a consistent fashion.

.gitignore
```diff
@@ -7,4 +7,4 @@
 # gtags
 GPATH
 GTAGS
-GRTAGS
+GRTAGS
```
lnicola, line 10: Nit: your IDE is eating final newlines.

src/vector/ops/set.rs
```diff
@@ -32,6 +32,35 @@ impl Geometry {
...
+    ///
+    /// Generates a new geometry which is the region of union of
+    /// the two geometries operated on.
+    /// Geometry validity is not checked. In case you are unsure of the
+    /// validity of the input geometries, call IsValid() before,
+    /// otherwise the result might be wrong.
+    ///
+    /// # Returns
+    /// Some(Geometry) if both Geometries contain pointers
+    /// None if either geometry is missing the gdal pointer, or there is an error.
```
lnicola, line 46:
```suggestion
    /// None if either geometry is missing the GDAL pointer, or there is an error.
```

metasim, line 46:
```suggestion
    /// `None` if either geometry is missing the gdal pointer, or there is an error.
```

metasim: I note that the documentation for `intersection` isn't that great. The docs in `predicates.rs` is closer to the mark. Would be great if you can bring it to that level, but if you can't I'll do a separate PR to normalize it.

lnicola (approved):
If it's not too much of a hassle, would you mind squashing?

zredb:
I think I would, but what does "squaring" mean? English is not my first language, so I don't quite understand what "squaring" really means.

lnicola:
Squashing, as in merging the 5 commits into a single one. Something like:

```sh
$ git reset @~5 # 5 commits before, or git reset 1dbc60215033caee1f99d933ae11354f70bc5f0e, the last commit before your changes
$ git add -u .
$ git commit -a
$ git status # make sure everything looks fine
$ git log # make sure everything looks fine
$ git ps -f
```

But you don't need to worry about it.

I'll let metasim take another look before merging.

metasim (approved):

PR #384, opened 2023-03-29
Add `Geometry::length`

lnicola:
---

### DIFF
--- a/src/vector/geometry.rs
+++ b/src/vector/geometry.rs
@@ -233,9 +233,19 @@ impl Geometry {
         Ok(())
     }
 
+    /// Compute geometry area in units of the spatial reference system in use.
+    ///
+    /// Supported for `Curve` (including `LineString` and `CircularString`) and `MultiCurve`.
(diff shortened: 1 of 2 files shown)


src/vector/geometry.rs
```diff
@@ -233,9 +233,19 @@ impl Geometry {
...
+    ///
+    /// See: [`OGR_G_Length`](https://gdal.org/api/vector_c_api.html#_CPPv412OGR_G_Length12OGRGeometryH)
+    pub fn length(&self) -> f64 {
+        unsafe { gdal_sys::OGR_G_Length(self.c_geometry()) }
+    }
+
     /// Compute geometry area in square units of the spatial reference system in use.
     ///
-    /// Supported for `LinearRing`, `Polygon` or `MultiPolygon`.
+    /// Supported for `LinearRing`, `Polygon` and `MultiPolygon`.
```
lnicola, line 248: Drive-by nit: I know it came from the GDAL docs, but I think this reads better than "or".

PR #394, opened 2023-05-07
Updated `set.rs` docs to be more consistent with those in `predicates.rs`.

metasim:
---

image

Context: https://github.com/georust/gdal/pull/379#discussion_r1135734190

### DIFF
--- a/src/vector/ops/set.rs
+++ b/src/vector/ops/set.rs
@@ -48,7 +48,7 @@ impl Geometry {
     /// * Geometry validity is not checked, and invalid geometry will generate unpredictable results.
     /// Use [`Geometry::is_valid`] if validity might be in question.
     /// * If GEOS is *not* enabled, this function will always return `None`.
-    /// You  may check for GEOS support with [`VersionInfo::has_geos`][has_geos].
+    /// You may check for GEOS support with [`VersionInfo::has_geos`][has_geos].
     ///
     /// # Returns
     /// * `Some(geometry)`: a new `Geometry` representing the computed union
--- a/src/vector/ops/set.rs
+++ b/src/vector/ops/set.rs
@@ -52,7 +52,7 @@ impl Geometry {
     ///
     /// # Returns
     /// * `Some(geometry)`: a new `Geometry` representing the computed union
-    /// * `None`: when the union was not able to be computed
+    /// * `None`: when the union could not be computed
     ///
     /// See: [`OGR_G_Union`][OGR_G_Union]
     ///
--- a/src/vector/ops/set.rs
+++ b/src/vector/ops/set.rs
@@ -4,20 +4,27 @@ use crate::vector::Geometry;
 ///
 /// These methods provide set operations over two geometries, producing a new geometry.
 impl Geometry {
-    /// Compute intersection.
+    /// Computes the [geometric intersection][intersection] of `self` and `other`.
     ///
-    /// Generates a new geometry which is the region of intersection of
-    /// the two geometries operated on. Call intersects (Not yet implemented)
-    /// to check if there is a region of intersection.
-    /// Geometry validity is not checked. In case you are unsure of the
-    /// validity of the input geometries, call IsValid() before,
-    /// otherwise the result might be wrong.
+    /// Generates a new geometry which is the region of intersection of the two geometries operated on.
+    ///
+    /// # Notes
+    /// * If you only need to determine if two geometries intersect and don't require
+    /// the resultant region, use [`Geometry::intersects`].
+    /// * Geometry validity is not checked, and invalid geometry will generate unpredictable results.
+    /// Use [`Geometry::is_valid`] if validity might be in question.
+    /// * If GEOS is *not* enabled, this function will always return `None`.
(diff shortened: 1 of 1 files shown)


src/vector/ops/set.rs
```diff
@@ -33,19 +40,25 @@ impl Geometry {
...
+    /// * Geometry validity is not checked, and invalid geometry will generate unpredictable results.
+    /// Use [`Geometry::is_valid`] if validity might be in question.
+    /// * If GEOS is *not* enabled, this function will always return `None`.
+    /// You  may check for GEOS support with [`VersionInfo::has_geos`][has_geos].
     ///
     /// # Returns
-    /// `Some(geometry)` if both Geometries contain pointers.
-    /// `None` if either geometry is missing the GDAL pointer, or there is an error.
+    /// * `Some(geometry)`: a new `Geometry` representing the computed union
+    /// * `None`: when the union was not able to be computed
```
metasim, line 55: I'm really frustrated by this, as the C/C++ docs say: 

> a new geometry representing the union or NULL if an error occurs.

which would imply to me that we should be returning `Result<Geometry, GDALError>`.

src/vector/ops/set.rs
```diff
@@ -4,20 +4,27 @@ use crate::vector::Geometry;
...
+    /// * Geometry validity is not checked, and invalid geometry will generate unpredictable results.
+    /// Use [`Geometry::is_valid`] if validity might be in question.
+    /// * If GEOS is *not* enabled, this function will always return `None`.
+    /// You  may check for GEOS support with [`VersionInfo::has_geos`][has_geos].
     ///
     /// # Returns
-    /// `Some(geometry)` if both Geometries contain pointers
-    /// `None` if either geometry is missing the GDAL pointer, or there is an error.
+    /// * `Some(geometry)`: a new `Geometry` representing the computed intersection
+    /// * `None`: when the geometries do not intersect or result could not be computed
```
metasim, line 21: Along with [this](https://github.com/georust/gdal/pull/394/files#r1186926701), frustrated to see the overloading of `Option` to handle both the empty geometry case as well as the error case. Not sure if we should try to fix this on the Rust side.

src/vector/ops/set.rs
```diff
@@ -4,20 +4,27 @@ use crate::vector::Geometry;
...
-    /// otherwise the result might be wrong.
+    /// Generates a new geometry which is the region of intersection of the two geometries operated on.
+    ///
+    /// # Notes
+    /// * If you only need to determine if two geometries intersect and don't require
+    /// the resultant region, use [`Geometry::intersects`].
+    /// * Geometry validity is not checked, and invalid geometry will generate unpredictable results.
+    /// Use [`Geometry::is_valid`] if validity might be in question.
+    /// * If GEOS is *not* enabled, this function will always return `None`.
+    /// You  may check for GEOS support with [`VersionInfo::has_geos`][has_geos].
```
lnicola, line 17: Nit:

```suggestion
    /// You may check for GEOS support with [`VersionInfo::has_geos`][has_geos].
```

src/vector/ops/set.rs
```diff
@@ -33,19 +40,25 @@ impl Geometry {
...
-    /// Geometry validity is not checked. In case you are unsure of the
-    /// validity of the input geometries, call IsValid() before,
-    /// otherwise the result might be wrong.
+    /// Generates a new geometry which is the union of the two geometries operated on.
+    ///
+    /// # Notes
+    /// * Geometry validity is not checked, and invalid geometry will generate unpredictable results.
+    /// Use [`Geometry::is_valid`] if validity might be in question.
+    /// * If GEOS is *not* enabled, this function will always return `None`.
+    /// You  may check for GEOS support with [`VersionInfo::has_geos`][has_geos].
```
lnicola, line 51:
```suggestion
    /// You may check for GEOS support with [`VersionInfo::has_geos`][has_geos].
```

lnicola (approved):

src/vector/ops/set.rs
lnicola: GDAL sets an error message when GEOS or SFCGAL are not available, but then the "correct" return type would be `Result<Option<Geometry>, GDALError>`, right? Which is a little annoying.

It might be a moot point. I suspect it's impossible to get a GEOS-less GDAL unless you're compiling it yourself, but then not everyone might know that GEOS is important to have when you're using vector data.

Also, I wonder if we should mention SFCGAL, which is required for [some geometry types](https://github.com/OSGeo/gdal/blob/e708608cf2466580db2690fa93cb3823ded6d4a3/ogr/ogrgeometry.cpp#L7469).

lnicola, line 55:
```suggestion
    /// * `None`: when the union could not be computed
```

src/vector/ops/set.rs
lnicola: > when the geometries do not intersect

At least GEOS (didn't test with SFCGAL) returns an empty geometry (not `NULL`) here.

src/vector/ops/set.rs
metasim: 🤔

PR #395, opened 2023-05-09
Split up `spatial_ref` module into feature-focused submodules.

metasim:
---

Also added small bits of documentation and fixed clippy warnings.

**Aside**: _I originally started this refactor because I'm about to expose the GCP API and erroneously thought it would be a part of the `spatial_ref` API. But now I see it's only `raster`-related. Therefore, this is basically some simple house-keeping..._


src/dataset.rs
```diff
@@ -686,6 +682,10 @@ impl Dataset {
         };
 
         let c_layer = unsafe {
+            let c_srs = match options.srs {
+                Some(srs) => srs.to_c_hsrs(),
+                None => null_mut(),
+            };
```
metasim, line 688: Since `to_c_hsrs` is now `unsafe`, moved from top of function to fall inside the same `unsafe` block as the call the the C API.

lnicola:
I didn't look at the other changes, but I don't think that function should be unsafe. I filed https://github.com/georust/gdal/discussions/360 for that, which.. you've already seen? :smile:

metasim:
@lnicola 

> I didn't look at the other changes, but I don't think that function should be unsafe. I filed #360 for that, which.. you've already seen? 😄

Ah! I misread your ticket!... and thought you were saying the opposite 🤦 

I defer to your experience here. I'll remove it here, and we can create a new PR to handle the other cases.... _or_ find an all around better way.

CHANGES.md
```diff
@@ -81,6 +81,10 @@
 
   - <https://github.com/georust/gdal/pull/370>
 
+- Split `spatial_ref` module into more feature-specific submodules. (Public API is unchanged.)
```
lnicola, line 84: Not sure if users will care about this :-).

lnicola (approved):
Thanks for these refactoring and documentation PRs, I think they're great!

PR #404, opened 2023-05-19
Add GCP access API

lnicola:
---

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 # Changes
 
 ## Unreleased
-- Added `Dataset::gcps` and `Dataset::set_gcps` APIs
+- Added `Dataset::gcp_projection`, `Dataset::gcps`, `Dataset::set_gcps` APIs
 
   - <https://github.com/georust/gdal/pull/404>
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 # Changes
 
 ## Unreleased
+- Added `Dataset::gcps` and `Dataset::set_gcps` APIs
+
+  - <https://github.com/georust/gdal/pull/404>
+
+
 - Added pre-built bindings for GDAL 3.7
 
   - <https://github.com/georust/gdal/pull/401>
--- a/src/gcp.rs
+++ b/src/gcp.rs
@@ -155,6 +155,11 @@ impl Dataset {
     ///
     /// See: [`GDALDataset::SetGCPs(int, const GDAL_GCP *, const OGRSpatialReference *)`](https://gdal.org/api/gdaldataset_cpp.html#_CPPv4N11GDALDataset7SetGCPsEiPK8GDAL_GCPPK19OGRSpatialReference)
     pub fn set_gcps(&self, gcps: Vec<Gcp>, spatial_ref: &SpatialRef) -> Result<()> {
+        assert!(
+            gcps.len() <= libc::c_int::MAX as usize,
+            "only up to `INT_MAX` GCPs are supported"
+        );
+
         struct CGcp {
             id: CString,
             info: CString,
@@ -196,7 +201,7 @@ impl Dataset {
         let rv = unsafe {
             gdal_sys::GDALSetGCPs2(
                 self.c_dataset(),
-                gdal_gcps.len() as i32,
+                gdal_gcps.len() as libc::c_int,
                 gdal_gcps.as_ptr(),
                 spatial_ref.to_c_hsrs() as *mut _,
             )
--- a/src/gcp.rs
+++ b/src/gcp.rs
@@ -124,6 +124,21 @@ impl Dataset {
         unsafe { SpatialRef::from_c_obj(c_ptr) }.ok()
     }
 
+    /// Get the projection definition string for the GCPs in this dataset.
+    ///
+    /// # Notes
+    /// * This is separate and distinct from [`Dataset::projection`], and only applies to
+    /// embedded GCPs.
+    ///
+    ///  See: [`GDALGetGCPProjection`](https://gdal.org/api/raster_c_api.html#gdal_8h_1a85ffa184d3ecb7c0a59a66096b22b2ec)
+    pub fn gcp_projection(&self) -> Option<String> {
+        let cc_ptr = unsafe { gdal_sys::GDALGetGCPProjection(self.c_dataset()) };
+        if cc_ptr.is_null() {
+            return None;
+        }
+        Some(_string(cc_ptr))
+    }
+
     /// Fetch GCPs.
     ///
     /// See: [`GDALDataset::GetGCPs`](https://gdal.org/api/gdaldataset_cpp.html#_CPPv4N11GDALDataset7GetGCPsEv)
@@ -208,6 +223,14 @@ mod tests {
         let gcp_srs = dataset.gcp_spatial_ref();
         let auth = gcp_srs.and_then(|s| s.authority().ok());
         assert_eq!(auth.unwrap(), "EPSG:4326");
+    }
+
+    #[test]
+    fn test_gcp_projection() {
+        let dataset = Dataset::open(fixture("gcp.tif")).unwrap();
+        let gcp_proj = dataset.gcp_projection();
+        assert!(gcp_proj.is_some());
+        assert!(gcp_proj.unwrap().contains("WGS 84"));
     }
 
     #[test]
--- a/src/gcp.rs
+++ b/src/gcp.rs
@@ -1,7 +1,110 @@
 //! Raster ground control point support
 
+use std::ffi::{CStr, CString};
+use std::marker::PhantomData;
+
+use gdal_sys::CPLErr;
+
+use crate::errors::Result;
 use crate::spatial_ref::SpatialRef;
+use crate::utils::{_last_cpl_err, _string};
 use crate::Dataset;
+
+/// An owned Ground Control Point.
+#[derive(Clone, Debug, PartialEq, PartialOrd)]
+pub struct Gcp {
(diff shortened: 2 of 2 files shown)


src/raster/gcp.rs
```diff
@@ -20,11 +123,83 @@ impl Dataset {
 
         unsafe { SpatialRef::from_c_obj(c_ptr) }.ok()
     }
+
```
metasim, line 126:
```suggestion
    /// Get the projection definition string for the GCPs in this dataset.
    ///
    /// # Notes
    /// * This is separate and distinct from [`Dataset::projection`], and only applies to
    /// embedded GCPs.
    ///
    ///  See: [`GDALGetGCPProjection`](https://gdal.org/api/raster_c_api.html#gdal_8h_1a85ffa184d3ecb7c0a59a66096b22b2ec)
    pub fn gcp_projection(&self) -> Option<String> {
        let cc_ptr = unsafe { gdal_sys::GDALGetGCPProjection(self.c_dataset()) };
        if cc_ptr.is_null() {
            return None;
        }
        Some(_string(cc_ptr))
    }
```

If you merge this I'll close https://github.com/georust/gdal/pull/402

src/raster/gcp.rs
```diff
@@ -34,4 +209,62 @@ mod tests {
         let auth = gcp_srs.and_then(|s| s.authority().ok());
         assert_eq!(auth.unwrap(), "EPSG:4326");
     }
+
```
metasim, line 212:
```suggestion
    #[test]
    fn test_gcp_projection() {
        let dataset = Dataset::open(fixture("gcp.tif")).unwrap();
        let gcp_proj = dataset.gcp_projection();
        assert!(gcp_proj.is_some());
        assert!(gcp_proj.unwrap().contains("WGS 84"));
    }
```

src/raster/gcp.rs
```diff
@@ -1,8 +1,111 @@
...
+    pub pixel: f64,
+    /// Line (y) location of GCP on raster.
+    pub line: f64,
+    /// X position of GCP in georeferenced space.
+    pub x: f64,
+    /// Y position of GCP in georeferenced space.
+    pub y: f64,
+    /// Elevation of GCP, or zero if not known.
+    pub z: f64,
+}
```
metasim, line 30: Is there not a way to avoid duplicating the structure of`GDAL_GCP`? Or are the potential options just not worth the effort, or the ergonomics are just bad?

src/raster/gcp.rs
```diff
@@ -1,8 +1,111 @@
...
+
+/// A wrapper over a Ground Control Point, borrowed from an existing dataset.
+#[derive(Clone, Debug)]
+#[repr(transparent)]
+pub struct GcpRef<'a> {
+    inner: gdal_sys::GDAL_GCP,
+    _phantom: PhantomData<&'a gdal_sys::GDAL_GCP>,
+}
```
metasim, line 38: While this is called a `GcpRef`, in reality this looks more like a "newtype plus lifetime", which feels more "owned" than a "ref". So why is `inner` a `GDAL_GCP`, rather than a `* const GDAL_GCP`?

Also, why does the `repr` matter here given this is claiming to be a "ref"?

src/raster/gcp.rs
metasim: Also feels like some impedance mismatching between `Gcp::id` as a field, and `GcpRef::id(&self)` as a function. Can those be made consistent?

lnicola: > Is there not a way to avoid duplicating the structure ofGDAL_GCP? Or are the potential options just not worth the effort, or the ergonomics are just bad?

The easier option would be to add the accessors to GDAL_GCP, but I'm not sure I like mixing types from the two crates (exposing gdal-sys from the gdal API).

lnicola: > Also feels like some impedance mismatching between Gcp::id as a field, and GcpRef::id(&self) as a function. Can those be made consistent?

I guess so, but `GcpRef` is what you consume as a user (so it doesn't really make sense to change it), while `Gcp` is what you set on the dataset, so it's important to be able to create one without a lot of effort. Making the fields private would probably mean something like a builder API.

src/raster/gcp.rs
metasim: Gotcha. I think I"m reading too much in the name, and have gotten myself embedded too deeply in the `foreign_types` semantics, where the borrowed `...Ref` type wraps an opaque pointer, and the owned type wraps (or can wrap) the actual structure.

src/raster/gcp.rs
metasim: >  not sure I like mixing types from the two crates (exposing gdal-sys from the gdal API).

💯 Agree.

src/raster/gcp.rs
```diff
@@ -20,11 +123,83 @@ impl Dataset {
...
+        unsafe { std::slice::from_raw_parts(data as *const GcpRef, len as usize) }
+    }
+
+    /// Assign GCPs.
+    ///
+    /// This method assigns the passed set of GCPs to this dataset, as well as
+    /// setting their coordinate system.
+    ///
+    /// See: [`GDALDataset::SetGCPs(int, const GDAL_GCP *, const OGRSpatialReference *)`](https://gdal.org/api/gdaldataset_cpp.html#_CPPv4N11GDALDataset7SetGCPsEiPK8GDAL_GCPPK19OGRSpatialReference)
+    pub fn set_gcps(&self, gcps: Vec<Gcp>, spatial_ref: &SpatialRef) -> Result<()> {
```
metasim, line 142: Devil's advocate question: Any arguments for *not* consuming `gcps`? I've not learned best practice here; when to choose from:

* `gcps: Vec<Gcp>`
* `gcps: &Vec<Gcp>`
* `gcps: &[Gcp]`

The last one seems most flexible (since we end up copying the data anyway), but as I said, I'm still learning API best practices.

src/raster/gcp.rs
```diff
@@ -1,8 +1,111 @@
...
+    /// Y position of GCP in georeferenced space.
+    pub y: f64,
+    /// Elevation of GCP, or zero if not known.
+    pub z: f64,
+}
+
+/// A wrapper over a Ground Control Point, borrowed from an existing dataset.
+#[derive(Clone, Debug)]
+#[repr(transparent)]
+pub struct GcpRef<'a> {
```
metasim, line 35: Is the lifetime parameter necessary? `inner` will be valid even if the originating `Dataset` goes away.

lnicola: It won't, it's the dataset's memory IIRC.

src/raster/gcp.rs
```diff
@@ -20,11 +123,83 @@ impl Dataset {
...
     }
+
+    /// Fetch GCPs.
+    ///
+    /// See: [`GDALDataset::GetGCPs`](https://gdal.org/api/gdaldataset_cpp.html#_CPPv4N11GDALDataset7GetGCPsEv)
+    pub fn gcps(&self) -> &[GcpRef] {
+        let len = unsafe { gdal_sys::GDALGetGCPCount(self.c_dataset()) };
+        let data = unsafe { gdal_sys::GDALGetGCPs(self.c_dataset()) };
+        unsafe { std::slice::from_raw_parts(data as *const GcpRef, len as usize) }
+    }
```
metasim, line 134: What's the argument against returning `Vec<Gcp>`, which (I think) would eliminate the need for `GcpRef`. Because of the data copies that would result? 

I partly ask because going from Rust to C we do a data copy (from `Gcp` to `GDAL_GCP`), but from C to Rust we have an extra `transparent`, castable structure `GcpRef` to avoid that copy, which seems asymmetric.

src/raster/gcp.rs
metasim: Ahhh... because there's a cast happening here:

https://github.com/georust/gdal/pull/404/files#diff-19ab0c93d77643f78caf7ad8ff89235da2a204d84593581fa208cb70d51caca7R133

Thank you.

src/raster/gcp.rs
lnicola: 2N+1 allocations. That's what I originally had.

It's asymmetric because of the string conversions. We can do C to Rust string conversions on request (in the getter), but we can't to do it the other way around because C will always read the strings.

src/raster/gcp.rs
lnicola: String to CString conversion should reuse the allocation of there's space for an extra byte.

Whether that's a good enough reason or not, that's arguable.

src/raster/gcp.rs
lnicola: Done, though we should have probably merged #402 directly instead :smile:.

src/raster/gcp.rs
lnicola: Maybe `BorrowedGcp` or `NonOwnedGcp` would be clearer?

src/raster/gcp.rs
metasim: 🙏

metasim:
bors d+

metasim:
@lnicola Do you want to include a fix to https://github.com/georust/gdal/issues/405 also?

lnicola:
In a separate PR, maybe. I'm not sure this one will get consensus.

metasim:
> In a separate PR

Sounds fine. Happy to do it if you want me to, but I think we should do it before cutting 0.15.

jdroenner:
i think this PR looks fine. We should merge it :)

lnicola:
bors r=metasim,jdroenner

lnicola:
Final bike-shedding question: should we go for `Dataset::ground_control_points` instead?

PR #406, opened 2023-05-22
Mapping of additional methods for SpatialRef.

tombieli:
---

Hello,
I wrote additional methods for `SpatialRef` to access some properties from `OGRSpatialReference`.
I needed those properties in project which I'm prototyping :) .
Of course I've added also relevant tests.

Best Regards,
Tomasz Bieliński


metasim:
@tombieli Thanks for the PR! Would you be willing to document the methods you added?

Here are some examples of what we're looking for:

https://github.com/georust/gdal/blob/master/src/vector/ops/transformations.rs

tombieli:
> @tombieli Thanks for the PR! Would you be willing to document the methods you added?
> 
> Here are some examples of what we're looking for:
> 
> https://github.com/georust/gdal/blob/master/src/vector/ops/transformations.rs
@metasim:
Yes. A commit is in the branch.
Best Regards

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,120 @@ impl SpatialRef {
             None
         }
     }
+
+    /// Get spheroid semi major axis. It returns [`Err`] variant in case of non [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE) error result from orginal method.
```
metasim, line 455: Adding a blank line between the summary and details helps the rustdoc render more cleanly.

```suggestion
    /// Get spheroid semi major axis. 
    ///
    /// Returns [`Err`] variant in case of non [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE) error result from the C library.
```

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,120 @@ impl SpatialRef {
...
+                err: rv,
+                method_name: "OSRGetSemiMajor",
+            });
+        }
+        Ok(a)
+    }
+
+    /// Get spheroid semi minor axis. It returns [`Err`] variant in case of non [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE) error result from orginal method.
+    /// 
+    /// See: [`OSRGetSemiMinor`](https://gdal.org/api/ogr_srs_api.html#_CPPv415OSRGetSemiMinor20OGRSpatialReferenceHP6OGRErr)
```
metasim, line 472:
```suggestion
    /// Get spheroid semi minor axis. 
    ///
    /// Returns [`Err`] variant in case of non [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE) error result from the C library.
    /// 
    /// See: [`OSRGetSemiMinor`](https://gdal.org/api/ogr_srs_api.html#_CPPv415OSRGetSemiMinor20OGRSpatialReferenceHP6OGRErr)
```

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,120 @@ impl SpatialRef {
...
+                err: rv,
+                method_name: "OSRSetProjParm",
+            });
+        }
+        Ok(())
+    }
+
+    /// Fetch a projection parameter value. It returns [`Err`] variant in case of non [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE) error result from orginal method.
+    /// 
+    /// See: [`OSRGetProjParm`](https://gdal.org/api/ogr_srs_api.html#_CPPv414OSRGetProjParm20OGRSpatialReferenceHPKcdP6OGRErr)
```
metasim, line 502:
```suggestion
    /// Fetch a projection parameter value. 
    ///
    /// Returns [`Err`] variant in case of non [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE) error result from the C library.
    /// 
    /// See: [`OSRGetProjParm`](https://gdal.org/api/ogr_srs_api.html#_CPPv414OSRGetProjParm20OGRSpatialReferenceHPKcdP6OGRErr)
```

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,120 @@ impl SpatialRef {
...
+                method_name: "OSRGetProjParm",
+            });
+        }
+        Ok(p)
+    }
+
+    /// Fetch a projection parameter value. In case of any error returns defualt value.
+    /// This associated function is variant of [`SpatialRef::get_proj_param`] which incorporates default fallback mechanism from orginal library.
+    /// 
+    /// See: [`OSRGetProjParm`](https://gdal.org/api/ogr_srs_api.html#_CPPv414OSRGetProjParm20OGRSpatialReferenceHPKcdP6OGRErr)
```
metasim, line 519:
```suggestion
    /// Fetch a projection parameter value. 
    ///
    /// In case of any error returns default value.
    /// This associated function is variant of [`SpatialRef::get_proj_param`] which incorporates default fallback mechanism from the C library.
    /// 
    /// See: [`OSRGetProjParm`](https://gdal.org/api/ogr_srs_api.html#_CPPv414OSRGetProjParm20OGRSpatialReferenceHPKcdP6OGRErr)
```

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,120 @@ impl SpatialRef {
...
+        if rv != OGRErr::OGRERR_NONE {
+            return Err(GdalError::OgrError {
+                err: rv,
+                method_name: "OSRGetSemiMinor",
+            });
+        }
+        Ok(b)
+    }
+
+    /// Set a projection parameter value. It returns [`Err`] variant in case of non [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE) result from orginal method.
```
metasim, line 485:
```suggestion
    /// Set a projection parameter value. 
    ///
    /// Returns [`Err`] variant in case of non [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE) result from the C library.
```

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,120 @@ impl SpatialRef {
...
+    pub fn get_proj_param_or_default(&self, name: &str, default: f64) -> f64 {
+        match CString::new(name) {
+            Ok(c_name) => unsafe { gdal_sys::OSRGetProjParm(self.0, c_name.as_ptr(), default, ptr::null_mut()) },
+            Err(_) => default
+        }
+    }
+
+    /// Set attribute value in spatial reference. It returns [`Err`] variant in case of non [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE) result from orginal method.
+    /// 
+    /// See: [`OSRSetAttrValue`](https://gdal.org/api/ogr_srs_api.html#_CPPv415OSRSetAttrValue20OGRSpatialReferenceHPKcPKc)
```
metasim, line 529:
```suggestion
    /// Set attribute value in spatial reference. 
    ///
    /// It returns [`Err`] variant in case of non [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE) result from the C library.
    /// 
    /// See: [`OSRSetAttrValue`](https://gdal.org/api/ogr_srs_api.html#_CPPv415OSRSetAttrValue20OGRSpatialReferenceHPKcPKc)
```

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,120 @@ impl SpatialRef {
...
+                err: rv,
+                method_name: "OSRSetAttrValue",
+            });
+        }
+        Ok(())
+    }
+
+    /// Fetch indicated attribute of named node. It returns [`Err`] variant if orginal method returns `nullptr`.
+    /// 
+    /// See: [`OSRGetProjParm`](https://gdal.org/api/ogr_srs_api.html#_CPPv415OSRGetAttrValue20OGRSpatialReferenceHPKci)
```
metasim, line 545:
```suggestion
    /// Fetch indicated attribute of named node. 
    ///
    /// Returns [`Err`] variant if the C library return `nullptr`.
    /// 
    /// See: [`OSRGetProjParm`](https://gdal.org/api/ogr_srs_api.html#_CPPv415OSRGetAttrValue20OGRSpatialReferenceHPKci)
```

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,120 @@ impl SpatialRef {
...
+        if c_ptr_value.is_null() {
+            return Err(_last_null_pointer_err("OSRGetAttrValue"));
+        }
+        Ok(_string(c_ptr_value))
+    }
+
+
+    /// Make a duplicate of the GEOGCS node of this [`SpatialRef`]. It returns [`Err`] variant if orginal method returns `nullptr`.
+    /// 
+    /// Seff: [OSRCloneGeogCS](https://gdal.org/api/ogr_srs_api.html#_CPPv414OSRCloneGeogCS20OGRSpatialReferenceH)
```
metasim, line 558:
```suggestion
    /// Make a duplicate of the GEOGCS node of this [`SpatialRef`]. 
    ///
    /// Returns [`Err`] variant if the C library returns `nullptr`.
    /// 
    /// Seff: [OSRCloneGeogCS](https://gdal.org/api/ogr_srs_api.html#_CPPv414OSRCloneGeogCS20OGRSpatialReferenceH)
```

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,141 @@ impl SpatialRef {
             None
         }
     }
+
+    /// Get spheroid semi major axis.
```
lnicola, line 455:
```suggestion
    /// Get spheroid semi-major axis.
```

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,141 @@ impl SpatialRef {
...
+    ///
+    /// Returns [`Err`] variant in case of non [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE) error result from the C library.
```
lnicola, line 457:
```suggestion
    /// Returns [`Err`] variant in case the C library returns an error code different from [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE).
```

lnicola:
Thanks, I'll take a look in a bit.

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,141 @@ impl SpatialRef {
             None
         }
     }
+
+    /// Get spheroid semi-major axis.
+    ///
+    /// Returns [`Err`] variant in case the C library returns an error code different from [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE).
```
lnicola, line 457: To be honest, I'm not really a fan of these, since they basically say "the function returns `Err` if the C function returned an error", which is true, but useless.

But I don't mind them too much, so it's all right if you like them.

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,141 @@ impl SpatialRef {
...
+        }
+        Ok(())
+    }
+
+    /// Fetch a projection parameter value.
+    ///
+    /// Returns [`Err`] variant in case the C library returns an error code different from [`OGRERR_NONE`](https://gdal.org/api/vector_c_api.html#c.OGRERR_NONE).
+    ///
+    /// See: [`OSRGetProjParm`](https://gdal.org/api/ogr_srs_api.html#_CPPv414OSRGetProjParm20OGRSpatialReferenceHPKcdP6OGRErr)
+    pub fn get_proj_param(&self, name: &str) -> Result<f64> {
```
lnicola, line 511: So ideally, this would return something like `Result<Option<f64>>` and the user could handle the missing case. But since we can do that, should we take the default as an argument?

It's currently ambiguous. Are the parameters never 0 in practice?

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,141 @@ impl SpatialRef {
...
+        }
+        Ok(())
+    }
+
+    /// Fetch indicated attribute of named node.
+    ///
+    /// Returns [`Err`] variant if the C library return `nullptr`.
+    ///
+    /// See: [`OSRGetProjParm`](https://gdal.org/api/ogr_srs_api.html#_CPPv415OSRGetAttrValue20OGRSpatialReferenceHPKci)
+    pub fn get_attr_value(&self, node_path: &str, child: u32) -> Result<String> {
```
lnicola, line 565: `child: usize` (potentially with an `assert!`) might be more idiomatic here, that's what we do with e.g. the dimensions in `GDALCreate`. No strong opinion, though.

src/spatial_ref/srs.rs
```diff
@@ -451,6 +451,141 @@ impl SpatialRef {
...
+    /// Fetch a projection parameter value.
+    ///
+    /// In case of any error returns default value.
+    /// This associated function is variant of [`SpatialRef::get_proj_param`] which incorporates default fallback mechanism from the C library.
+    ///
+    /// See: [`OSRGetProjParm`](https://gdal.org/api/ogr_srs_api.html#_CPPv414OSRGetProjParm20OGRSpatialReferenceHPKcdP6OGRErr)
+    pub fn get_proj_param_or_default(&self, name: &str, default: f64) -> f64 {
+        match CString::new(name) {
+            Ok(c_name) => unsafe {
+                gdal_sys::OSRGetProjParm(self.0, c_name.as_ptr(), default, ptr::null_mut())
```
lnicola, line 534: This drops the error. I think we should return `Result<f64>`, like above (unless we figure GDAL never returns an error).

src/spatial_ref/srs.rs
tombieli: I maybe tried to be to precise and I created to explicit documentation.

src/spatial_ref/srs.rs
tombieli: Behavior of `OSRGetProjParm` is quite tricky, a least from Rust point of view.

In C and C++ if value is  found it is returned normally, but if value is not found, a value from default parameter is returned, but also error code (if pointer is not null) is set to `OGRERR_FAILURE`.  However looking from Rust point of view it may mean not found (therefore is candidate for `Option::None`) or can mean any other failure.
You can see this in GDAL [source code for `OSRGetProjParm`](https://github.com/OSGeo/gdal/blob/51d91973656df7be083a9194780da286c6a2cab7/ogr/ogrspatialreference.cpp#L5667).

To sum up:
* on existing value: we receive the value itself,
* if value does not exist: we receive default value and very general error code.

That's why I used only `Result<f64>` as return type and I also created also `get_proj_param_or_default`.

src/spatial_ref/srs.rs
tombieli: This function was created to allow Rust consumer of library to use it's original default fallback mechanism. Behavior of  `OSRGetProjParm` is quite tricky - see comment above.  In this case we intentionally do not bother if GDAL will return error, because we have default value.  To retain consistency with this behavior an error from  `CString::new(name)` is dropped (maybe this is bad idea).

This function could be candidate to removal from PR.

src/spatial_ref/srs.rs
tombieli: I've decided to change type of that parameter to `usize`.

src/spatial_ref/srs.rs
lnicola: Docs are good, but this is currently neither consistent with the rest of the documentation, nor useful: "the function returns `Err` if an error has occurred" doesn't bring any information at all to someone used to the Rust conventions.

Looking at `GetSemiMajor`, it's a bit weird -- it returns both a valid value (`SRS_WGS84_SEMIMAJOR`) and an error if it can't find the value, which is something I'm not sure we want to emulate.

Can you change this to something like "Returns an error if the semi-major axis can't be found" ("semi-minor" for the other one), "if there the `PROJCS` node is missing" for `set_proj_param`?

`GetProjParm` returns an error if it can't find `PROJCS` or `CONVERSION`, or if the parameter is missing. `OSRSetAttrValue` returns an error if the path is empty. `OSRGetAttrValue` returns an error if the attribute can't be found.

For `geog_cs`, it's a bit complicated, I guess an error means that it's missing.

Also, `OSRSetAttrValue` in GDAL can take a null value, which means that the node is created without a value. Maybe we should take an `Option<&str>` to support that?

lnicola: Docs are good, but this is currently neither consistent with the rest of the documentation, nor useful: "the function returns `Err` if an error has occurred" doesn't bring any information at all to someone used to the Rust conventions.

Looking at `GetSemiMajor`, it's a bit weird -- it returns both a valid value (`SRS_WGS84_SEMIMAJOR`) and an error if it can't find the value, which is something I'm not sure we want to emulate.

Can you change this to something like "Returns an error if the semi-major axis can't be found" ("semi-minor" for the other one), "if there the `PROJCS` node is missing" for `set_proj_param`?

`GetProjParm` returns an error if it can't find `PROJCS` or `CONVERSION`, or if the parameter is missing. `OSRSetAttrValue` returns an error if the path is empty. `OSRGetAttrValue` returns an error if the attribute can't be found.

For `geog_cs`, it's a bit complicated, I guess an error means that it's missing.

These expose some of the GDAL internals, but I don't expect them to change significantly.

Also, `OSRSetAttrValue` in GDAL can take a null value, which means that the node is created without a value. Maybe we should take an `Option<&str>` to support that?

src/spatial_ref/srs.rs
lnicola: Yeah, I'm inclined towards something like a single `pub fn get_proj_param(&self, name: &str, default: f64) -> Result<Option<f64>>`, which returns `Err` on a string conversion error (basically never), and `Ok(None)` on `OGRERR_FAILURE`. It's not that awkward to use, and should be pretty clear.

Both error cases in the implementation are similar, and boil down to "there's no parameter with that name on the `PROJCS`.

src/spatial_ref/srs.rs
tombieli: This function is removed.

src/spatial_ref/srs.rs
tombieli: @lnicola:
I've corrected comments.

I've also decided to change signature of `get_proj_param` to:
```rust
pub fn get_proj_param(&self, name: &str) -> Result<Option<f64>>
```
I did not include a default value option, because in each case this value will be discarded.

I've done similar thing to `get_attr_value`.

There is also problem with `set_attr_value`. I've tried to write some tests which set some silly (like "Spam" or "Eggs") and less silly (like "WGS_1984" and "metre") data, but results were quite strange. I'm not an expert at manipulation of WKT for projection, so I maybe tried to write tests in wrong way. Maybe some of You will have better idea how to write those tests.

Alternatively we can remove this function now and implement it later.

metasim (approved):

src/spatial_ref/srs.rs
lnicola: > but results were quite strange

You're right, I'll try to look into it.

lnicola: I took the liberty of pushing a commit to fix this. It was a pretty basic gotcha -- `Option::map` consumes the value. I would have hoped that clippy would catch this, but alas...

tombieli: Thanks. Rust/C++ edge can be tricky.

lnicola: Everything can be tricky. The Python bindings will segfault if you do something similar with a SpatialRef, so it's not just us.

lnicola:
Added one more commit with an `assert!` in `get_attr_value`.

r? @jdroenner, otherwise I'll r+ tomorrow or so.

tombieli:
> Added one more commit with an `assert!` in `get_attr_value`.

@lnicola:
I have a question: is this a good idea? As far as I know `assert!` will introduce potential `panic!` to production code. Maybe better solution will to return some `Err` or discard check entirely? C++ code should handle this properly.

lnicola:
@tombieli first of all, my bad, I shouldn't have pushed this commit, but rather left it as a review comment. I only did it because I felt bad for letting this PR linger for so long, and wanted to get the new release out of the gate soon. And I actually mentioned it in the same comment where I suggested using `usize`. I would revert it now, but it feels like extra churn.

That said, to address your points:

> C++ code should handle this properly.

The C++ code doesn't handle this properly, because the issue is in the Rust wrapper. C++ takes `i32` (on common platforms), we have an `usize` (basically `u64`). Without the `assert!`, a large value like `0x8000_0000_0000_0000` will end up after conversion as `0`. GDAL won't be able to tell because it will happen in our own code.

> or discard check entirely

That feels like sweeping the problem -- which we introduced, by taking `usize` to make for a nicer API -- under the rug, allowing the caller to make unwanted changes to the SpatialRef, like in my example above.

So the question is whether to panic or return an error. My reasoning is that the caller must have gotten the index from somewhere. As far as I can tell, there are three ways one could get it:

 - the caller knows the structure of the SpatialRef, and changes one specific node (like we do in our test); in this case, passing something like `0x8000_0000_0000_0000` is clearly incorrect (since GDAL can't even represent that), and we should panic since it's a caller bug
 - the caller got the index from `FindProjParm`: in that case, it can't be outside of `int` (since GDAL returns one), so if the index is outside of the `int` limits, it's again a caller bug. We don't currently expose that function, but this is true regardless of that.
 - the caller used `SetNode` (`OSRSetAttrValue`) to create a new child, so they know that e.g. `0` is a valid index. If I'm reading the docs correct, that will create at most one child. And even if there was an `AddNode` method (which I couldn't spot), you wouldn't be able to create more than `c_int::MAX` children, since GDAL clearly can't represent that.

So I don't see any way code that's not hopelessly broken could end up with such a large index there. And I think in Rust it's rather common to panic in such cases, instead of allowing the program to keep running, only to crash or fail in a more obscure way later on.

But if we do decide to keep the panic, we should document it just in case.

tombieli:
@lnicola: Thank You for Your answer.
First of all: I'm not upset that You made this commit - You are maintainer here.

I questioned this move with `assert!`, because I know there are two ideas in Rust programming:
* To represent all good and bad states of application as types. For instance the `Result` type and other enums,
* Not to use `panic!` in production code. I've heard that is even macro or some other CI/CD pipeline plugin that fails entire build, when linking to `panic!` is detected.

But looking from other side: I see on this simple example, that mapping API to Rust from C++ could be very tricky, especially when C++ library has its own non-rust ideas to create function contract (for instance: using signed integer for non-negative values and three-state result/error reporting).

I understand that chances to fire that `assert!` is quite low.

I don't have big experience in Rust, and no experience in shipping production code in Rust. I know that having theoretical bright ideas is not enough, so I don't expect that my opinion on this topic is 100% accurate (it may even less than 50% ;) ). If You as maintainer want this `assert!` I accept it.  

> But if we do decide to keep the panic, we should document it just in case.

I will add relevant info in docs.

src/spatial_ref/srs.rs
```diff
@@ -562,6 +562,8 @@ impl SpatialRef {
     /// * `Ok(None)` - if node or attribute are not found (C library will return `nullptr`) or attribute contains no value,
     /// * `Err(_)` - if there is a string conversion error.
     ///
+    /// **Note:** this fuction panics when child > [libc::c_int::MAX].
```
lnicola, line 565: Can you format it [like this](https://doc.rust-lang.org/src/core/cell.rs.html#791-793)? Also, should it have some backquotes (`child > `[`libc::c_int::MAX`])? I'm not sure how the links work here.

tombieli: I made a correction.

lnicola:
Can you rebase (or merge) to fix the conflict?

tombieli:
> Can you rebase (or merge) to fix the conflict?

Yes. Merge commit is in the branch.

lnicola:
bors r=metasim,lnicola

PR #410, opened 2023-05-31
Fix GDAL 3.7 bindings

lnicola:
---


src/dataset.rs
```diff
@@ -405,8 +405,23 @@ impl Dataset {
     /// Flush all write cached data to disk.
     ///
     /// See [`GDALFlushCache`].
-    pub fn flush_cache(&mut self) {
-        unsafe { GDALFlushCache(self.c_dataset) }
+    ///
+    /// Note: on GDAL versions older than 3.7, this function always succeeds.
+    pub fn flush_cache(&mut self) -> Result<()> {
+        #[cfg(any(all(major_ge_3, minor_ge_7)))]
```
jdroenner, line 411: i think we also need to add `major_ge_4`  to the any block. and in the other case we could add `major_lt_3` but i guess we dont support GDAL versions < 3.

lnicola: Yeah, added it now, but I don't think we're really consistent with these.

jdroenner: i never thought about a GDAL version > 3 when i "invented" this :see_no_evil: . There will be some trouble when the version really switches to 4...  
Maybe we could write a macro or something to handle this better.

jdroenner (approved):

lnicola:
bors r=jdroenner

lnicola:
Just for the record, these were broken because I generated the 3.7 bindings on GDAL 3.6.3 😅. Thanks @latot for spotting it.

PR #420, opened 2023-07-05
Fix use-after-free in Dataset::close

lnicola:
---


jdroenner (approved):

src/dataset.rs
```diff
@@ -174,6 +174,7 @@ impl GeoTransformEx for GeoTransform {
 #[derive(Debug)]
 pub struct Dataset {
     c_dataset: GDALDatasetH,
+    closed: bool,
```
lnicola, line 177: We could also have a wrapper around `c_dataset` and use `ManuallyDrop` on that instead.

src/dataset.rs
```diff
@@ -428,8 +432,10 @@ impl Dataset {
     ///
     /// See [`GDALClose`].
     ///
-    /// Note: on GDAL versions older than 3.7, this function always succeeds.
-    pub fn close(&mut self) -> Result<()> {
+    /// Note: on GDAL versions older than 3.7.0, this function always succeeds.
+    pub fn close(mut self) -> Result<()> {
```
lnicola, line 436: We could keep the old signature, but we'd have to check for `closed` on every public method. Or we could yank 0.15.0, which might be a good idea anyway.

jdroenner: yes i think thats the way to go

lnicola: Which way? 😅

jdroenner: change the signature

jdroenner: oh sorry. I think we should change it to ` pub fn close(mut self) -> Result<()> `

src/dataset.rs
jdroenner: `ManuallyDrop` sounds nice

lnicola: I switched to `ManuallyDrop` in https://github.com/georust/gdal/commit/2e4b75ee6c974febd7da09cd8859aecbb77674c5, but I'm not sure it's much better.

lnicola: Scratch that, it doesn't really work, because the inner type's `drop` still runs, and we can't get the return value of `GDALClose` out of there. I'll revert to the previous version.

jdroenner:
sorry i forgot to merge this. Should we create a new release tomorrow?

PR #422, opened 2023-07-06
Use latest minor releases and fix Arrow GPG errors

lnicola:
---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,17 +16,17 @@ jobs:
     strategy:
       matrix:
         version:
-          - 3.7
-          - 3.6
-          - 3.5
-          - 3.4
-          - 3.3
-          - 3.2
-          - 3.1
+          - 3.7.0
+          - 3.6.4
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -36,6 +36,10 @@ jobs:
       - name: Install build deps
         shell: bash
         run: |
+          curl -LO https://apache.jfrog.io/artifactory/arrow/ubuntu/apache-arrow-apt-source-latest-$(grep -F VERSION_CODENAME /etc/os-release | cut -d= -f2).deb
```
jdroenner, line 39: why, do we need the arrow source? Is it for the build with bindgen?

lnicola: It's not the source, it's the APT repository: https://arrow.apache.org/install/. We need it because otherwise `apt update` fails with a GPG error on some GDAL images.

jdroenner: well then i guess we have no choice...

PR #429, opened 2023-08-22
prepare 0.16.0 changelog + version

jdroenner:
- [x ] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,22 +1,26 @@
 # Changes
 
 ## Unreleased
+
+## 0.16
+
 - **Breaking**: `Dataset::close` now consumes `self`
 
- - <https://github.com/georust/gdal/pull/420>
+- <https://github.com/georust/gdal/pull/420>
 
 - `Gcp` and `GcpRef` are now public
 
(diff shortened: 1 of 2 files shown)


lnicola:
The whitespace changes break the formatting:

image

(not that I find this layout great, but at least it's consistent)

And maybe we should publish it as 0.15.1 and yank 0.15.0?

jdroenner:
0.15.1 vs 0.16.0 is a good question. Since there is a breaking change we should name it 0.16.0. Using 0.15.1 could cause some code to stop compiling but who ever uses that code should change it anyway :sweat:  I'm open for suggestions what version number to pick.

lnicola:
I think it's fine if we yank it and publish 0.15.1, but let's just skip to 0.16.0, as it seems that a lot of people don't really like yanking. I had some more planned breaking changes, but version numbers are free :-).

jdroenner:
yes will do. We need to bump gdal-sys to 0.9.1 because we changed the bindgen version. Should we also bump the dependencies for the gdal crate?

PR #434, opened 2023-09-06
Make raster index usize

lnicola:
---

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,25 +2,29 @@
 
 ## Unreleased
 
+- **Breaking**: `Dataset::raster_count` now returns an `usize` and `Dataset::rasterband` now takes `usize` instead of `isize`
+
+  - <https://github.com/georust/gdal/pull/434>
+
  - **Breaking**: `CslStringListIterator` returns a `CslStringListEntry` instead of `(String, String)` in order to differentiate between `key=value` entries vs `flag` entries.
  - **Breaking**: `CslStringList::fetch_name_value` returns `Option<String>` instead of `Result<Option<String>>`, better reflecting the semantics of GDAL C API.
  - Added `CslStringList::get_field`, `CslStringList::find_string`, `CslStringList::partial_find_string`, `CslStringList::find_string_case_sensitive`, `CslStringList::into_ptr`, `CslStringList::add_name_value`.
- 
+
    - <https://github.com/georust/gdal/pull/455>
 
- - **Breaking**: `ExtendedDataType` no longer implements `Clone`, `PartialEq` and `Eq`
+- **Breaking**: `ExtendedDataType` no longer implements `Clone`, `PartialEq` and `Eq`
 
-   - <https://github.com/georust/gdal/pull/451>
+  - <https://github.com/georust/gdal/pull/451>
 
- - **Breaking**: Moved `LayerIterator`, `LayerOptions` and `Transaction` to `crate::vector`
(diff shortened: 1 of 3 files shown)


jdroenner:
I'm not against this change but i always worry that there is a hidden reason why it is an int and not an uint in C/C++...

lnicola:
[They're 1-based](https://gdal.org/doxygen/classGDALDataset.html#ad96adcf07f2979ad176e37a7f8638fb6). I think it's just that `int` is the "default" data type in C.

> nBandId	the index number of the band to fetch, from 1 to GetRasterCount().

src/dataset.rs
```diff
@@ -544,7 +544,10 @@ impl Dataset {
     ///
     /// Applies to raster datasets, and fetches the
     /// rasterband at the given _1-based_ index.
-    pub fn rasterband(&self, band_index: isize) -> Result<RasterBand> {
+    pub fn rasterband(&self, band_index: usize) -> Result<RasterBand> {
+        assert!(band_index > 0);
```
jdroenner, line 548: i think we should return an error here.

jdroenner (changes requested):

src/dataset.rs
jdroenner: The method already returns a Result so why panic when we can also return a matching error?

lnicola: Because it's not an error the caller can handle, it's a logic bug in the code. So it makes sense to report it right away instead of waiting until later, when it might be less clear what failed.

jdroenner: thats a good idea. thanks!

lnicola: It reports an error like `ERROR 5: foo.tif: GDALDataset::GetRasterBand(2) - Illegal band #`.

For 0 we could just call `GDALGetRasterBand`, but for a band number larger than `c_int::MAX` we'd need to synthetize the message or add a new error variant.

jdroenner: if it reports an error for 0 then its fine for me

metasim:
If it wouldn't cause a riot, I'd suggest we change it to `NonZeroUsize`. But it does add a bunch of `unwrap` and `try_into` friction.

src/dataset.rs
lnicola: Updated.

lnicola:
Yeah, I've considered that, but it seems annoying. `raster_count()` would still return `usize`.

metasim:
I'm guessing this would be a non-starter here, but in my day job, because I've seen so many off-by-one band errors, I've gone to the extreme of creating a type for band indexes to force us to think about what we're doing and where. 

https://gist.github.com/metasim/3539e492f0cfa2c50ac0fe4e3d0ca9a5

This works in concert with convenience extension methods on `Dataset`: 

* `fn band_indexes(&self) -> impl Iterator<Item = BIdx>`
* `fn band(&self, idx: BIdx) -> Result<Rasterband>`

lnicola:
I can see where you're coming from, but it feels too opinionated. You could drop either of the two variants and it wouldn't matter too much.

metasim (approved):

metasim:
@jdroenner FWIW I'm in favor of approving and merging this.

jdroenner (approved):

jdroenner:
i guess we need to resolve the changelog conflicts again

lnicola:
bors r=metasim,jdroenner

PR #440, opened 2023-09-27
Use tinymarble.tif instead of the PNG

lnicola:
---

We already have the TIFF in the repo, and this fixes the [block size](https://github.com/metasim/gdal/blob/0a50ba27b21ebe6dbe05d1ba1342dfa3c9af0fe1/src/raster/tests.rs#L507-L510) tests, where the official Docker images use `ENABLE_WHOLE_IMAGE_OPTIMIZATION`, which [causes](https://github.com/OSGeo/gdal/blob/bb931fe41aa2c348c6d40f4d3b0e3612246ba21e/frmts/png/pngdataset.cpp#L110-L119) the block size to be equal to the image size, while most distributions of GDAL don't.

I regenerated the existing TIFF, which has reversed (?), using:

```sh
gdal_translate tinymarble.png tinymarble.tif
gdal_edit.py -a_srs EPSG:4326 tinymarble.tif
gdaladdo -ro -r bilinear tinymarble.tif 2 4
```


src/metadata.rs
```diff
@@ -320,26 +320,26 @@ mod tests {
...
     fn test_get_metadata_domains() {
-        let dataset = Dataset::open(fixture("tinymarble.png")).unwrap();
+        let dataset = Dataset::open(fixture("tinymarble.tif")).unwrap();
         let mut domains = dataset.metadata_domains();
-        if domains[0].is_empty() {
-            domains.remove(0);
-        }
+        domains.retain(|d| !d.is_empty());
+        domains.sort();
+        domains.dedup();
```
lnicola, line 327: `ubuntu-lts` returns two `xml:XMP` entries.

src/raster/tests.rs
```diff
@@ -759,17 +734,17 @@ fn test_raster_stats() {
     assert_eq!(
         rb.get_statistics(true, false).unwrap().unwrap(),
         StatisticsAll {
-            min: 12.0,
+            min: 0.0,
             max: 255.0,
-            mean: 89.2526,
-            std_dev: 90.99835379412092,
+            mean: 68.4716,
+            std_dev: 83.68444773934999,
```
metasim, line 740: Any idea why the statistics changed? Is it computing approx stats via overviews?

metasim:
bors d+

src/raster/tests.rs
lnicola: No, the original TIFF was somehow different from the PNG:

image

Reversed, on top, is the original TIFF, on bottom is the PNG converted to TIFF. Notice the darker area at the South pole and what looks like JPEG artifacts (but might be something else). I think both versions are derived from the real Blue Marble, but perhaps not in an ideal way.

lnicola:
bors r=metasim

lnicola:
bors retry

Issue #365, opened 2023-02-05
Debug-printing geometries segfaults


lnicola:
Something like:

```rust
fn main() -> Result<()> {
    let args = std::env::args().collect::<Vec<_>>();
    let point_ds = Dataset::open(&args[1])?;
    let mut layer = point_ds.layer(0)?;
    for feature in layer.features() {
        dbg!(feature);
    }
    Ok(())
}
```

crashes with:

```
[src/main.rs:17] feature = Feature {
    _defn: Defn {
        c_defn: 0x0000562e86c81c80,
    },
    c_feature: 0x0000562e86c85d60,
    geometry: [
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', ~/.cargo/registry/src/github.com-1ecc6299db9ec823/gdal-0.14.0/src/vector/geometry.rs:177:38
stack backtrace:
   0: rust_begin_unwind
             at /rustc/efd27454ae94386e11997b1158f60d3a79deeb7d/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/efd27454ae94386e11997b1158f60d3a79deeb7d/library/core/src/panicking.rs:64:14
   2: core::panicking::panic
             at /rustc/efd27454ae94386e11997b1158f60d3a79deeb7d/library/core/src/panicking.rs:114:5
   3: gdal::vector::geometry::Geometry::wkt
   4: <&T as core::fmt::Debug>::fmt
```

metasim:
@lnicola Was going to replicate on MacOS. Can you share the input file?

lnicola:
It happens with any input, like the sample from https://github.com/ngageoint/GeoPackage/blob/master/docs/examples/java/example.gpkg.

metasim:
Not really the problem, but calling `unwrap` seems like a bad idea?

https://github.com/georust/gdal/blob/c7f88b56a855cfdff44267009f2ad1730cd03b14/src/vector/geometry.rs#L177

metasim:
@lnicola I wonder if [this (`foreign_types`)](https://docs.rs/foreign-types/latest/foreign_types/index.html) or something like it might be worth considering. I think we need some principled semantics around these GDAL pointers, Rust has all the type machinery we need, we just need to find good examples of how to do this properly or rely on a library that does it for us.

lnicola:
Still haven't looked into it, but I think we handle layers in a better way, that could serve as inspiration.

metasim:
[...]

Issue #168, opened 2021-02-23
RFC: handle getters (`c_dataset` and friends) should be safe


lnicola:
The Rust convention is to place `unsafe` at the usage site, e.g. `as_ptr` is safe, but dereferencing the pointer is not. We should follow the same convention.

rmanoka:
I would be a bit more confident with changing the `pub(crate)` ones to begin with.

metasim:
Closed by https://github.com/georust/gdal/pull/447

PR #445, opened 2023-10-01
A Consistent and Principled Approach to GDAL C Pointers

metasim:
---

> _Note_: This PR is likely to be abandoned. While the `foreign-types` changes work well with the `raster`-related types, the ownership model behind the `vector` and `mdarray` types is sufficiently complicated that a transition would require more interested participants than currently available, and is likely not worth the effort at this time.

---

_Remaining Tasks_:

* [x] Convert `Dataset` (dependency https://github.com/georust/gdal/pull/447)
* [ ] Convert `GCP` (possibly; it has a Rust type mirroring C type)
* [ ] Figure out what to do with `{Owned}Layer` (it's [really strange](https://discord.com/channels/598002550221963289/1149380155949711450/1158848863592001568))
* [x] Convert `FieldDefn`
* [ ] Convert `mdarray::MDArray`
* [ ] Convert `mdarray::Group`
* [ ] Convert `mdarray::Dimension`
* [ ] Convert `mdarray::ExtendedDataType`
* [ ] Convert `mdarray::Attribute`
* [ ] Rename `SpatialRef` to `SpatialReference` to avoid `SpatialRefRef`
* [ ] Convert any relevant secondary APIs, e.g. things from `vsi`, `cpl`, etc.
* [ ] Unify mutable reference handling (see below) (dependency https://github.com/georust/gdal/issues/448)
* [ ] Extended CHANGELOG documentation to help users migrate
* [ ] Run under [Sanitizer workflow](https://github.com/metasim/gdal/actions/workflows/sanitizer.yml)
* [X] I agree to follow the project's [code of conduct](https://github.com/georust/gdal/blob/master/CODE_OF_CONDUCT.md).
* [ ] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.

## Background

The Rust bindings to the GDAL library have been developed and maintained by many contributors over a time span of almost a decade (the first commit on record is March 25, 2014). As is normal in such a situation, inconsistencies and diverging principles arise. The scope of this PR is to address inconsistencies in the handling of pointers to C-instantiated GDAL types, making use use of the [`foreign-types`](https://crates.io/crates/foreign-types) crate as a means.

The `foreign-types` crate is mature and [widely used](https://crates.io/crates/foreign-types/reverse_dependencies), with 45 public dependents (including `openssl`), approximately 120,000 daily downloads, and > 65 million total downloads. It is dual licensed under Apache and MIT licenses. It provides an opinionated set of types and traits for wrapping pointers to C memory, with support for borrowed vs owned references, `Clone` and `Drop`, with optional proc-macros to eliminate some boilerplate. The included documentation and examples are somewhat sparse, but the maintainer [is responsive](https://github.com/sfackler/foreign-types/issues/23) to questions. If you dig deep enough (including the very large projects that use the crate), you can find good examples. While the names and abstractions declared in `foreign-types` may not be to the preferences of all users, maturity, consistency and friction-lowering were given primacy in considering `foreign-types`.

_Note_: For simplicity I use the term "pointer" to include memory addresses, handles, or any other identifier to a C-allocated resource in GDAL.

## Goals

The impetus for this PR is several logged issues around pointer handling:

* https://github.com/georust/gdal/issues/4
* https://github.com/georust/gdal/issues/168
* https://github.com/georust/gdal/issues/306
* https://github.com/georust/gdal/issues/365
* https://github.com/georust/gdal/discussions/360

This PR aims to provide foundational work helpful in addressing these issues, primarily focusing on "safe consistency". Specifically:

* Visibility: There's inconsistency as to when pointer access functions are `pub`. Because `georust/gdal` is (currently) a subset of the functionality covered by `OSGeo/gdal`, it's important to enable advanced users to access GDAL pointers in a principled but enabling way.
* `unsafe`: There's inconsistency in when pointer-related methods are marked as `unsafe`. It is not [idiomatic](https://doc.rust-lang.org/std/os/fd/trait.IntoRawFd.html#tymethod.into_raw_fd) to treat pointer reading, manipulation or arithmetic as `unsafe`. Rust considers it safe to manipulate pointers and other handles, while dereferencing or turning them back into something useable *is* `unsafe`. 
* Naming : 
  - Accessor methods: Currently we have names like `c_dataset`, `to_c_hsrs`, `c_options`, `to_c_hct`, `c_geometry`, etc. to access the C pointer.
  - Constructor methods: Similarly, there are currently various mechanisms for constructing Rust types from C pointers
* Owned vs borrowed values: GDAL assumes a delineation between owned vs. borrowed pointers, as declared in the documentation of most functions. With `foreign-types` we can lift those documented concepts into compiler-checked constraints, handling the allocation, deallocation, cloning, ownership transfer, and lifetime tracking aspects of the resource lifecycle. 

## Benefits

The intended benefits are:

* Improved user ergonomics via consistency
* Improved efficiency when binding to additional GDAL features (e.g. VRTs, Warp API, etc.)
* Reduced maintenance burden via lower cognitive load
* Better ergonomics for advanced users implementing alternative or bespoke bindings to unexposed GDAL features
* The Big One: Full compiler-checked resource lifecycle handling, including allocation, deallocation, cloning, ownership transfer, and lifetimes

## Scope

This PR migrates the following types to use `foreign-types`

* `Defn`
* `Feature`
* `Geometry`
* `SpatialRef`
* `CoordTransform`
* `CoordTransformOptions`
* `RasterBand`

(See _Remaining Tasks_ below)

_Note_: Because of this scope, many APIs will break, necessitating clear documentation and appropriate semantic version handling (albeit we're still < 1.0).

---

## Notes on Migration

These are some thoughts I had about the migration experience, some of which are copied from the [related discussion](https://github.com/georust/gdal/discussions/360) on this topic. I include them here for those interested in the deeper deails behind certain implementation decisions.

* At first I didn't love the idea of using the `foreign_type!` macros, but after implementing `ForeignType` and `ForeignTypeRef` manually, it's definitely worth it _when it's appropriate_. it is not always appropriate, such as the case for `RasterBand` where we are _always_ working with shared references (always owned by `Dataset`). See below.
* When using the macro, looking at the generated code can be helpful in learning how to properly use the types. See [here](https://github.com/georust/gdal/discussions/360#discussioncomment-5878136) for an example.
* You can handle both generics as well as lifetime parameters, but takes deeper understanding of `foreign-types`.
* `ForeignType::from_ptr`, which is `unsafe`, is how you create instances.
* `ForeignType::as_ptr`, which is "safe" (as we desire here), lets you get the pointer.
* `ForeignType::into_ptr` consumes/moves ownership, helpful for C functions that take over ownership.
* Except for constructors and other methods that do not take a `&self` parameter, most methods should be implemented on the borrowed `XYZRef` type. This is because the owned `XYZ` type auto de-refs to it, enabling methods to work on both types.
* If `Clone` is supported, then `ToOwned` is implemented by the macro, which is nice. This is how you go from `XYZRef` to `XYZ` (via `XYZRef::to_owned`), i.e. how you convert from a borrowed reference to an owned instance.
* (I think) The value of `type CType` is implicitly expected to be `Sized`. Therefore (and I could be missing something), the type alias in `bindgen` can't be explicitly used. For example `bindgen` gives us `type OGRSpatialReferenceH = *mut libc::c_void`, but in `ForeignType` implementation, `type CType = OGRSpatialReferenceH` will cause errors. Instead it ends up needing to be `type CType = libc::c_void`, which seems unfortunate when using `foreign-types` with `bindgen`. Might be worth a question to the maintainer.
* There are several cases (such as `Layer`) where the code can be made significantly more principled through `foreign_types`: 
  - e.g.: `Layer` was [holding onto a `&Defn`](https://github.com/georust/gdal/commit/10a1310dbb89ec3c87848efb4cf34b46290f0030#diff-65fa51ecf72a12889cd0a4ddeb0a68691e480f41b5cfac001c4fb5dc7c85bee8L17-R24) solely for the lifetime parameter tracking, which wasn't technically needed originally, and easier to eliminate with `foreign_types`.
  - `OwnedLayer` can go away.
  - We can eliminate `owned: bool` fields that exist in some types because this is explicit in the type; e.g. `Geometry::owned`.
* In cases where the type is **always** owned by the C library and never managed directly in Rust, the implementation is quite simple if you use `Opaque` and implement `ForeignTypeRef`. e.g. [`RasterBand`](https://github.com/georust/gdal/commit/877c8ea1b55074a612d5d6e95f1489b2101f1dcd#diff-432e492d6a25cce04b77014813aea6861769ff42ce0c362d3f7850fb9e21ab5cL165-R168).
* Haven't yet figured out if pointers to pointers to C structures (such as `CslStringList`) can be effectively managed by this library.
* The mutability semantics around `RasterBand` are decoupled from `Dataset`. Even though `Dataset` owns every `RasterBand` and defines its lifetime, you can mutate a `RasterBand` without having a mutable reference to `Dataset`, which seems wrong. I _think_ [this is what we want to do](https://github.com/georust/gdal/commit/877c8ea1b55074a612d5d6e95f1489b2101f1dcd#diff-b0b3f516e4a8d589b9f4d18fa849158e60b50f7ed974d69533b58bbe082ddc2bR502-R507): Under `foreign_types`, `Dataset::rasterband(&self, band_index: isize)` would return `Result<&mut Rasterband>`. Or maybe there's a new method `Dataset::rasterband_mut(&mut self, band_index: isize)`. See _Remaining Tasks_ above.


fixtures/tinymarble.tif.aux.xml
```diff
@@ -1,3 +1,12 @@
...
   <SRS dataAxisToSRSAxisMapping="2,1">GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AXIS["Latitude",NORTH],AXIS["Longitude",EAST],AUTHORITY["EPSG","4326"]]</SRS>
+  <PAMRasterBand band="1">
+    <Metadata>
+      <MDI key="STATISTICS_MAXIMUM">255</MDI>
+      <MDI key="STATISTICS_MEAN">68.4716</MDI>
+      <MDI key="STATISTICS_MINIMUM">0</MDI>
+      <MDI key="STATISTICS_STDDEV">83.68444773935</MDI>
+      <MDI key="STATISTICS_VALID_PERCENT">100</MDI>
+    </Metadata>
+  </PAMRasterBand>
```
metasim, line 11: I'll revert this.

src/vector/sql.rs
```diff
@@ -153,7 +154,7 @@ impl Dataset {
 
         Ok(Some(sql::ResultSet {
             layer,
-            dataset: c_dataset,
+            dataset: c_dataset, // TODO: We're ending up with a shared reference here...
```
metasim, line 157: Will need a second pair of eyes on this one.

metasim:
While I think the approaches herein are ones we should adopt, I think it's too late (with volunteer labor) to apply a unified approach across the whole library. While the porting `raster` bindings was pretty straight forward, the `vector` and `mdarray` bindings are complicated enough to require the original author of those to do the work, or explain how  ownership/exclusive access should be handled to someone else to take on the work.

PR #447, opened 2023-10-01
Maintenance cleanup on `dataset.rs`

metasim:
---

Changes: 
* Moved ancillary types from `dataset.rs` into their own or related files.
* Moved non-core Dataset methods into respective `raster` or `vector` modules. Aims to simplify maintainability via "separation of concerns".
* Removed `unsafe` from pointer accessor methods.

_Possibly_ breaking changes:

* Moved `LayerIterator`, `LayerOptions` and `Transaction` to `crate::vector`.

---

Aside: The impetus for this refactor is to focus the changes necessary for adding `Dataset` to https://github.com/georust/gdal/pull/445. But even if that PR takes a different route, this change will benefit long-term maintenance.


fixtures/tinymarble.tif.aux.xml
```diff
@@ -1,3 +1,12 @@
 <PAMDataset>
   <SRS dataAxisToSRSAxisMapping="2,1">GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AXIS["Latitude",NORTH],AXIS["Longitude",EAST],AUTHORITY["EPSG","4326"]]</SRS>
+  <PAMRasterBand band="1">
```
lnicola, line 3: Should we add this to `.gitignore`? I'd have expected the statistics test to fail here, because it checks that they're not available unless `bForce`, but CI seems to pass.

metasim: @lnicola It used to be in `.gitignore`, but you removed it in https://github.com/georust/gdal/commit/779b7cfa1b160fa736bf870652488c6cf0905312 , and I wasn't sure why.

😄

metasim: I'll add it back, and undo this change.

metasim: Not sure why `.gitignore` isn't sufficient, but I had to run `git update-index --assume-unchanged fixtures/tinymarble.tif.aux.xml` to get the test-caused changes to not get detected. Not sure if this propagates upstream. I may try to figure out what test is generating it, and have the test run in a `TempFixture`.

metasim:
bors r=lnicola

PR #455, opened 2023-10-23
Fix/454

### LINKED ISSUE #454: `CslStringList` does not propertly support non-key-value entries.
## TL;DR: 

The design of `CslStringList::iter` is incomplete, and arguably wrong altogether.

## Details

`CslStringList` Is used extensively in GDAL to pass stringly represented arguments/parameters to a number of GDAL subsystems. It's a glorified wrapper around a C-array of C-strings (`*mut *mut c_char`). The GDAL API provides a number of convenience methods for manipulating this memory in a slightly more safe way than working with the C pointers directly. The API has grown over time, and has some eccentricities about it that aren't obvious at first glance.

One of these nuances is you can insert two forms of entries:

| Name | Example | Insert Method | Fetch Method | 
|--------|--------|--------|
| _key-value_ | `"KEY=VALUE"` | `CslStringList::set_name_value` | `CslStringList::fetch_name_value` |
| _flag_ | `"FLAG"` | `CslStringList::add_string` |  Unimplemented |

The current Rust binding design did not fully consider the _flag_ form of entries, at least not from the read/fetch perspective (you can insert them, but never retrieve or view them). 

Furthermore, the implementation of `CslStringList::iter` assumes the _key-value_ form (returning `(String, String)`, and skips over entries of the _flag_ form, as illustrated by this test:

```rust
#[test]
fn iter_over_all() -> Result<()>{
    let mut l = CslStringList::new();
    l.set_name_value("ONE", "1")?;
    l.set_name_value("TWO", "2")?;
    l.set_name_value("THREE", "3")?;
    l.add_string("SOME_FLAG")?;

    assert_eq!(l.len(), 4);
    assert_eq!(l.iter().count(), 4); // <-- 'assertion failed: `(left == right)`

    Ok(())
}
```

As a consequence, _Flag_ form entries are also excluded from `Debug` formatting, which makes it hard to inspect code having to use `CslStringList`.

## Considerations

In addressing these inadequacies, the following decisions need to be made:

1. Do we have a single `Iterator` returning both _key-value_ **and** _flag_ forms, or do we have two different iterators?
2. If it's a single `Iterator`, what's the return type?
3. To flesh out support for _flag_ form entries, should we also expose the following capabilities:
  - `CSLFindString`
  - `CSLFindStringCaseSensitive `
  - `CSLPartialFindString`

metasim:
---

Closes #454

### DIFF
--- a/src/cpl.rs
+++ b/src/cpl.rs
@@ -5,18 +5,43 @@
 
 use std::ffi::CString;
 use std::fmt::{Debug, Display, Formatter};
+use std::mem::ManuallyDrop;
 use std::ops::Deref;
 use std::ptr;
+use std::str::FromStr;
 
-use gdal_sys::{CSLAddString, CSLCount, CSLDestroy, CSLDuplicate, CSLFetchNameValue, CSLFindString, CSLFindStringCaseSensitive, CSLGetField, CSLPartialFindString, CSLSetNameValue};
+use gdal_sys::{
+    CSLAddString, CSLCount, CSLDestroy, CSLDuplicate, CSLFetchNameValue, CSLFindString,
+    CSLFindStringCaseSensitive, CSLGetField, CSLPartialFindString, CSLSetNameValue,
+    CSLTokenizeString2,
+};
 use libc::{c_char, c_int};
 
 use crate::errors::{GdalError, Result};
 use crate::utils::_string;
 
-/// Wraps a [`gdal_sys::CSLConstList`]  (a.k.a. `char **papszStrList`). This data structure
-/// (a null-terminated array of null-terminated strings) is used throughout GDAL to pass
-/// `KEY=VALUE`-formatted options to various functions.
+/// Wraps a [`gdal_sys::CSLConstList`]  (a.k.a. `char **papszStrList`).
+///
+/// This data structure (a null-terminated array of null-terminated strings) is used throughout
+/// GDAL to pass `KEY=VALUE`-formatted options to various functions.
+///
+/// There are a number of ways to populate a [`CslStringList`]. See below for examples.
+///
+/// # Example
+///
+/// ```rust
+/// use gdal::cpl::CslStringList;
+///
+/// let mut sl1 = CslStringList::new();
+/// sl1.set_name_value("NUM_THREADS", "ALL_CPUS").unwrap();
+/// sl1.set_name_value("COMPRESS", "LZW").unwrap();
+///
+/// let sl2: CslStringList = "NUM_THREADS=ALL_CPUS COMPRESS=LZW".parse().unwrap();
+/// let sl3: CslStringList = (&[("NUM_THREADS", "ALL_CPUS"), ("COMPRESS", "LZW")]).try_into().unwrap();
+///
+/// assert_eq!(sl1, sl2);
+/// assert_eq!(sl2, sl3);
+/// ```
 ///
 /// See the [`CSL*` GDAL functions](https://gdal.org/api/cpl.html#cpl-string-h) for more details.
 pub struct CslStringList {
@@ -122,7 +147,7 @@ impl CslStringList {
     ///
     /// Returns `Some(usize)` of value index position, or `None` if not found.
     ///
-    /// See:: [`CSLPartialFindString`](https://gdal.org/api/cpl.html#_CPPv420CSLPartialFindString12CSLConstListPKc)
+    /// See: [`CSLPartialFindString`](https://gdal.org/api/cpl.html#_CPPv420CSLPartialFindString12CSLConstListPKc)
     /// for details.
     pub fn partial_find_string(&self, fragment: &str) -> Option<usize> {
         let fragment = CString::new(fragment).ok()?;
@@ -143,21 +168,20 @@ impl CslStringList {
         // See: https://github.com/OSGeo/gdal/blob/fada29feb681e97f0fc4e8861e07f86b16855681/port/cpl_string.cpp#L181-L182
         let field = unsafe { CSLGetField(self.as_ptr(), index as c_int) };
         if field.is_null() {
-            return None
+            return None;
         }
 
         let field = _string(field);
         if field.is_empty() {
             None
-        }
-        else {
+        } else {
             Some(field.deref().into())
         }
     }
 
     /// Determine the number of entries in the list.
     ///
-    /// See [`CSLCount`](https://gdal.org/api/cpl.html#_CPPv48CSLCount12CSLConstList) for details.
+    /// See: [`CSLCount`](https://gdal.org/api/cpl.html#_CPPv48CSLCount12CSLConstList) for details.
     pub fn len(&self) -> usize {
         (unsafe { CSLCount(self.as_ptr()) }) as usize
     }
@@ -168,13 +192,20 @@ impl CslStringList {
     }
 
     /// Get an iterator over the `name=value` elements of the list.
-    pub fn iter<'a>(&'a self) -> impl Iterator<Item=CslStringListEntry> + 'a {
+    pub fn iter(&self) -> impl Iterator<Item = CslStringListEntry> + '_ {
         CslStringListIterator::new(self)
     }
 
     /// Get the raw pointer to the underlying data.
     pub fn as_ptr(&self) -> gdal_sys::CSLConstList {
         self.list_ptr
+    }
+
+    /// Get the raw pointer to the underlying data, passing ownership
+    /// (and responsibility for freeing) to the the caller.
+    pub fn into_ptr(self) -> gdal_sys::CSLConstList {
+        let s = ManuallyDrop::new(self);
+        s.list_ptr
     }
 }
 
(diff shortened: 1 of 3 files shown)


lnicola:
Oh, this is great. I'll take a look tomorrow.

metasim:
Sanitizer results: https://github.com/metasim/gdal/actions/runs/6618106265/job/17975832725

src/cpl.rs
```diff
@@ -125,7 +228,121 @@ impl Clone for CslStringList {
...
+        for i in 0..len {
+            if let Some(field) = self.get_field(i) {
+                f.write_fmt(format_args!("{field}\n"))?;
+            }
+        }
+        Ok(())
+    }
+}
+
+impl PartialEq for CslStringList {
```
lnicola, line 256: Maybe we shouldn't implement this? Users might have different expectations here, like lists with a different order of elements being equal.

src/cpl.rs
```diff
@@ -59,9 +83,10 @@ impl CslStringList {
         Ok(())
     }
 
-    /// Adds the string `value` to the list.
+    /// Adds a copy of the string slice `value` to the list.
     ///
-    /// Returns `Ok<()>` on success, `Err<GdalError>` if `value` cannot be converted to a C string.
+    /// Returns `Ok<()>` on success, `Err<GdalError>` if `value` cannot be converted to a C string,
```
lnicola, line 88: I know these are old, but we should probably take the chance to fix them to say `Ok(())` and `Err(GdalError::Whatever)`.

src/cpl.rs
```diff
@@ -125,7 +228,121 @@ impl Clone for CslStringList {
...
+    fn try_from(pairs: &[(&str, &str); N]) -> Result<Self> {
+        let mut result = Self::default();
+        for (k, v) in pairs {
+            result.set_name_value(k, v)?;
+        }
+        Ok(result)
+    }
+}
+
+/// Represents an entry in a [CslStringList], which is ether a single token, or a key/value assignment.
```
lnicola, line 309:
```suggestion
/// Represents an entry in a [`CslStringList`], which is ether a single token, or a key/value assignment.
```

src/cpl.rs
```diff
@@ -125,7 +228,121 @@ impl Clone for CslStringList {
...
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum CslStringListEntry {
```
lnicola, line 311: I wonder if `Flag` and `Pair` would be clearer.

src/cpl.rs
```diff
@@ -125,7 +228,121 @@ impl Clone for CslStringList {
...
+        match self {
+            CslStringListEntry::Arg(s) => f.write_str(s),
+            CslStringListEntry::Assign { key, value } => f.write_fmt(format_args!("{key}={value}")),
+        }
+    }
+}
+
 /// State for iterator over [`CslStringList`] entries.
+///
+/// Note: Does not include values inserted with [CslStringList::add_string]
```
lnicola, line 345:
```suggestion
/// Note: Does not include values inserted with [`CslStringList::add_string`]
```

src/cpl.rs
```diff
@@ -74,18 +99,89 @@ impl CslStringList {
...
+        let fragment = CString::new(fragment).ok()?;
+        let idx = unsafe { CSLPartialFindString(self.as_ptr(), fragment.as_ptr()) };
+        if idx < 0 {
+            None
+        } else {
+            Some(idx as usize)
+        }
+    }
+
+    /// Fetch the [CslStringListEntry] for the entry at the given index.
```
lnicola, line 162:
```suggestion
    /// Fetch the [`CslStringListEntry`] for the entry at the given index.
```

src/cpl.rs
```diff
@@ -74,18 +99,89 @@ impl CslStringList {
...
+    ///
+    /// Returns `None` if index is out of bounds
+    pub fn get_field(&self, index: usize) -> Option<CslStringListEntry> {
+        // In the C++ implementation, an index-out-of-bounds returns an empty string, not an error.
```
lnicola, line 166: We should probably panic or return `None` if `index > c_int::MAX`.

src/cpl.rs
```diff
@@ -95,15 +191,22 @@ impl CslStringList {
         self.len() == 0
     }
 
-    /// Get an iterator over the name/value elements of the list.
-    pub fn iter(&self) -> CslStringListIterator {
+    /// Get an iterator over the `name=value` elements of the list.
+    pub fn iter(&self) -> impl Iterator<Item = CslStringListEntry> + '_ {
```
lnicola, line 195: It might be better to return `CslStringListIterator` here, just in case the user wants to name that type.

Although, with generators coming "soon", this expectation might change (because they're can't be named anyway).

src/cpl.rs
```diff
@@ -125,7 +228,121 @@ impl Clone for CslStringList {
...
+            b.field(&e.to_string());
+        }
+
+        b.finish()
+    }
+}
+
+impl Display for CslStringList {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        // CSLPrint would be preferable here, but it can only write to a file descriptor.
```
lnicola, line 245: Should we use `iter` instead?

src/cpl.rs
```diff
@@ -59,9 +83,10 @@ impl CslStringList {
         Ok(())
     }
 
-    /// Adds the string `value` to the list.
+    /// Adds a copy of the string slice `value` to the list.
```
lnicola, line 86: Can you also add `CSLAddNameValue` for symmetry?

src/cpl.rs
```diff
@@ -125,7 +228,121 @@ impl Clone for CslStringList {
...
+                cstr.as_ptr(),
+                DELIM.as_ptr() as *const c_char,
+                CSLT_HONOURSTRINGS | CSLT_PRESERVEESCAPES,
+            )
+        };
+        Ok(Self { list_ptr: c_list })
+    }
+}
+
+/// Convenience for creating a [`CslStringList`] from a slice of _key_/_value_ tuples.
```
lnicola, line 287: It's a little annoying that this is quadratic. Can you add a comment mentioning that?

lnicola (approved):
Looks good overall, I left some nits (mostly).

src/cpl.rs
metasim: I did it for the tests and doc-tests. So, for example, without this we'd have to change `assert_eq!(sl1, sl2);` to `assert_eq!(sl1.to_string(), sl2.to_string());`.  Think that's ok?

src/cpl.rs
```diff
@@ -125,7 +228,121 @@ impl Clone for CslStringList {
...
+///
+/// See [`CSLTokenizeString`](https://gdal.org/api/cpl.html#_CPPv417CSLTokenizeStringPKc) for details
+impl FromStr for CslStringList {
+    type Err = GdalError;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        // See: https://github.com/OSGeo/gdal/blob/cd2a054b0d7b881534baece69a8f52ddb69a53d9/port/cpl_string.h#L86C1-L97
+        static CSLT_HONOURSTRINGS: c_int = 0x0001;
+        static CSLT_PRESERVEESCAPES: c_int = 0x0008;
+        static DELIM: &[u8; 4] = b" \n\t\0";
```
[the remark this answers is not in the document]
metasim: No kidding!! Went down a rabbit hole on that one!

src/cpl.rs
lnicola: But users will use it!

src/cpl.rs
metasim: Thanks. 🤦  I reimplemented that 3 times, reevaluating my assumptions, and lost the forest for the trees.

src/cpl.rs
metasim: I was wondering if we should get rid of it all together... Adding `FromStr` makes it much less necessary.

Can you think of any way to _not_ make it quadratic? Seems like the whole API is ironically inefficient.

lnicola: Nope, I don't think we can. I like kinda' like the API more than `FromStr`, though.

metasim: Unless you have objections, I'm going to remove it.

lnicola: It's useful when you don't know the values at compile-time. You'd need to format a string otherwise.

metasim: Another problem with it is that I couldn't get it to work with generic `&[(&str, &str)]` slices. IOW, it _does_ only work if you know the length at compile time:

```rust
impl<const N: usize> TryFrom<&[(&str, &str); N]> for CslStringList
```

If you remove the `const` parameter, you get the following compiler error:

<details><summary>Details</summary>
<pre>
error[E0277]: the trait bound `cpl::CslStringList: From<&[(&str, &str); 2]>` is not satisfied
   --> src/cpl.rs:518:41
    |
518 |         let l = CslStringList::try_from(&[("ONE", "1"), ("TWO", "2")])?;
    |                 ----------------------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<&[(&str, &str); 2]>` is not implemented for `cpl::CslStringList`
    |                 |
    |                 required by a bound introduced by this call
    |
    = help: the following other types implement trait `TryFrom<T>`:
              <cpl::CslStringList as TryFrom<RasterizeOptions>>
              <cpl::CslStringList as TryFrom<&[(&str, &str)]>>
    = note: required for `&[(&str, &str); 2]` to implement `Into<cpl::CslStringList>`
    = note: required for `cpl::CslStringList` to implement `TryFrom<&[(&str, &str); 2]>`
</pre>
</details>

metasim: I'm going to decline adding a comment right now. Not quite sure how to phrase it given it's a problem with the C API and not the Rust bindings, and also I'm not happy with the current implementation at all.

lnicola: I think that was just me trying to be cute about this, so `try_from(["ONE", "1"])` would work without the `&`. But looks like it doesn't work anyway.

We can replace that impl with `impl TryFrom<&[(&str, &str)]> for CslStringList`, which should work in your example.

lnicola: As for the comment, we can just say "Due to the GDAL API, this is quadratic and might be slow for large number of entries. Consider using `FromStr` instead.".

metasim: @lnicola I'll revisit this in my next PR. I have to add a `merge` method anyway.

lnicola: Ah, that might need an `as_slice` on the caller side. Anyway, we should probably implement `FromIterator` and `Extend`? I'm not sure.

metasim: `FromIterator` is a great idea. I tried parameterizing on `AsRef<[(&str, &str)]>` but got into orphan rule issues.

PR #456, opened 2023-10-24
GDAL DEM Processing Routines

metasim:
---
image

- [x] Add `hillshade`
- [x] Add `color-relief`
- [x] Wait for and merge https://github.com/georust/gdal/pull/457
- [x] Remove `DemCommonOptionsOwner` and inline the fields into the various structs.
- [x] Remove `DemCommonOptions`
- [x] Remove `Option` from around `DemCommonOptions::compute_edges` and `additional_options`.
- [x] ~Convert `DemProcessing` methods to inherent methods on `Dataset`.~
- [x] Convert `DemProcessing` methods to free functions.


src/raster/processing/dem/mod.rs
metasim, line 1: `dem` is a submodules of `processing` because I'm planning on a `raster::processing::warp` submodule as well.

src/test_utils.rs
```diff
@@ -107,3 +107,54 @@ pub fn open_gpkg_for_update(path: &Path) -> (TempPath, Dataset) {
...
+            "field {}: |{} - {}| = {} is greater than epsilon {:.4e}",
+            $field,
+            $left,
+            $right,
+            ($left - $right).abs(),
+            $ep
+        )
+    };
+    // Pseudo-specialization
+    (StatisticsAll, $left:expr, $right:expr, epsilon = $ep:expr) => {
```
lnicola, line 149: Nit: not really a fan of this arm and the one above it, they're more complex than calling `assert_near!` four times.

metasim: > And it's a public API.

Is it? It's only part of`cfg(test)`...

src/raster/processing/dem/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Execute the processor on the given [`Dataset`].
+    fn eval(&self, ds: &Dataset) -> Result<Dataset> {
+        let opts = self.to_options();
+        let popts = GdalDEMProcessingOptions::new(&opts);
+        let mode = CString::new(self.dem_algorithm().to_string())?;
+        let dest = self
+            .common_options()
+            .destination()
+            .clone()
+            .unwrap_or_else(|| MemFilePath::new().as_ref().to_path_buf());
```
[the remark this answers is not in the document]
metasim: > This is not going to be deleted once you close the result Dataset, is it?

Good question; I'll research. My assumption was that closing the DS released the memory.

src/test_utils.rs
lnicola: Ah, sorry. Didn't see that in the diff.

src/raster/processing/dem/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+use libc::c_int;
+use std::ffi::CString;
+use std::fmt::{Display, Formatter};
+use std::marker::PhantomData;
+use std::num::NonZeroUsize;
+use std::path::{Path, PathBuf};
+use std::ptr;
+
+#[derive(Debug, Clone, Default)]
+pub struct DemCommonOptions {
```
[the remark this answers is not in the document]
metasim: Got it. I'll try it that way and see how the ergonomics feel.

metasim:
@lnicola Thank you for the early feedback. Course corrections earlier rather than laters are definitely helpful.

metasim:
@lnicola Two questions:

1. Do you think this set of features should be behind a `dem` feature flag?
2. Are you adverse to exposing the capability like this:

### In `raster/processing/dem/slope.rs`
```rust

impl Dataset {
    fn slope(&self, options: &SlopeOptions) -> Result<Dataset> { ... }
    ...
}
```

### In `raster/processing/dem/aspect.rs`
```rust

impl Dataset {
    fn aspect(&self, options: &AspectOptions) -> Result<Dataset> { ... }
    ...
}
```

etc.

I like this approach as it makes it more discoverable in IDE/rust-analyzer.

lnicola:
No to both

src/raster/processing/dem/mod.rs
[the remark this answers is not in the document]
lnicola: I didn't include it, but the destination should probably be a separate parameter instead of an options field.

lnicola:
Sorry, I was on my phone earlier and didn't pay enough attention to the code snippets.

1. I don't think a feature is necessary, since it only gates some pretty thin bindings to the GDAL functionality, not the whole implementation
2. I'm not sure :grimacing:. I disliked it initially, but perhaps it's not such a bad idea. Maybe we should ask for another opinion.

src/raster/processing/dem/mod.rs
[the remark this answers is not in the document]
metasim: I'm going to make that change in the next commit, but it's going to add a bit more friction to the examples and tests.

src/raster/processing/dem/mod.rs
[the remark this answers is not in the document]
metasim: @lnicola You are right. I was getting `/vsimem/` and the **Memory** driver confused. The latter does get released when the Dataset is closed, but the former does not (requires an explicit `unlink`) We do have the path in `MemFilePath` to `unlink` it, but it's starting to feel like feature creep around something that's not key to the feature at hand. 

* https://gdal.org/user/virtual_file_systems.html#vsimem-in-memory-files
* https://gdal.org/drivers/vector/memory.html#memory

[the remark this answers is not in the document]
lnicola: Yup, feels like we should just let the user decide (and maybe offer `MemFilePath` or something like that).

metasim:
> 2. I'm not sure 😬. I disliked it initially, but perhaps it's not such a bad idea. Maybe we should ask for another opinion.

FWIW: We do have precedence for it (and I like the code-organization benefits):

```
❯ rg 'impl Dataset'
src/dataset.rs
37:impl Dataset {

src/vector/sql.rs
60:impl Dataset {

src/vector/layer.rs
593:impl Dataset {

src/gcp.rs
109:impl Dataset {

src/vector/transaction.rs
100:impl Dataset {

src/raster/rasterband.rs
23:impl Dataset {

src/raster/mdarray.rs
800:impl Dataset {

src/raster/processing/dem/mod.rs
54:impl Dataset {
```

lnicola:
The organization is fine, which is why I initially said I had nothing about it.

What I'm not sure I like is exposing the algorithms as inherent methods on the datasets. Most of the stuff we have there maps directly to `GDALDatasetXXX`. The DEM and warper are separate in the GDAL API.

metasim:
> What I'm not sure I like is exposing the algorithms as inherent methods on the datasets. Most of the stuff we have there maps directly to GDALDatasetXXX. The DEM and warper are separate in the GDAL API.

Aaaaah. Got it. A nuance I didn't catch. It's hard for me to tell what in the C API design is well-considered, and what's accidental.

metasim:
I flip-flopped between `impl Dataset` in each sub-module, vs in the `dem` module. Settled on the latter for coherence sake in the generated docs.

https://github.com/georust/gdal/pull/456/files#diff-c3e410b73d29819318373ad9149f478156e2060050ec127f3a234f245f862843R53-R305

src/raster/processing/mod.rs
```diff
@@ -0,0 +1,15 @@
...
+
+// TODO: Remove after https://github.com/georust/gdal/pull/457 is merged.
+pub(crate) fn merge(left: &mut CslStringList, right: &CslStringList) {
+    for opt in right.iter() {
+        match opt {
+            CslStringListEntry::Flag(a) => left.add_string(&a).unwrap(),
+            CslStringListEntry::Pair { name, value } => left.set_name_value(&name, &value).unwrap(),
+        }
+    }
+}
```
metasim, line 15: To be replaced by `impl Extend<...> for CslStringList`.

src/vsi.rs
metasim, line 1: These changes are really just cleanup of maintenance annoyances. I was in this file when I thought I'd try supporting in-memory generations of DEM results, and couldn't resist moving monomorphization optimizations into inner functions to cut down on the visual clutter. If the reviewer objects, I can undo.

tests/driver-no-auto-register.rs
```diff
@@ -1,6 +1,12 @@
...
-
 use gdal::{Dataset, DriverManager};
+use std::path::{Path, PathBuf};
+
+/// Returns the fully qualified path to `filename` in `${CARGO_MANIFEST_DIR}/fixtures`.
+fn fixture(filename: &str) -> PathBuf {
+    Path::new(env!("CARGO_MANIFEST_DIR"))
+        .join("fixtures")
+        .join(filename)
+}
```
metasim, line 9: Been considering a non-published `gdal-testkit` module so that we can have shared testing support for unit and integration tests...

tests/utils.rs
metasim, line 1: Misc. cleanup after making fixture lookup consistent in `tests`.

lnicola:
This looks pretty nice, except:

 - I'm not a fan of the `DemCommonOptionsOwner` thing. I would inline the four fields into the structs; this causes some code duplication, but feels much easier to follow, and might even more robust in the long term (e.g. maybe one of the options gets dropped by one of the apps, so it's no longer "common"). Is `DemCommonOptions` something that applications will actually care about?
 - `DemCommonOptions::compute_edges` should probably be `bool`, not `Option<bool>`; same for `additional_options`
 - `DemProcessing` being a trait is not too bad, but they might as well be inherent methods. It seems like extra complexity for no good reason ([cf.](https://twitter.com/mycoliza/status/1708933310548935148)).

metasim:
> DemCommonOptions::compute_edges should probably be bool, not Option<bool>; same for additional_options

The reason for `compute_edges` being `Option` is so we don't have to keep in sync with what the GDAL defaults are, should they change. IOW, If the caller doesn't explicitly set it, we do whatever the GDAl default is. It probably won't change, so I'll just set it to the documented default, but wanted to point out there was some rationale to it.

lnicola:
> so we don't have to keep in sync with what the GDAL defaults are

That's what I suspected, but as a user, it makes for a pretty confusing API.

> `DemProcessing` trait

I'm really sorry for the back and forth on this, but I now think they should be free functions. Some of the reasons are:

 - matching the C++ / Python / Java API; if you've used GDAL before, this makes you feel at home with the bindings
 - not crowding the `Dataset` API
 - putting them on the `Dataset` feels very arbitrary -- why not on `Rasterband`?

metasim:
> putting them on the Dataset feels very arbitrary -- why not on Rasterband?

Because the primary input to the all the DEM functions is the Dataset.


I'll convert them to free functions, but to me it seems just as arbitrary as any other decisions to make something inherent on dataset. From where I sit, in the C API, if the first parameter is `Dataset`, it might as well be treated as the receiver.

lnicola:
The interface I see for `Dataset` is basically https://gdal.org/doxygen/classGDALDataset.

metasim:
@lnicola I got rid of all the `DemCommonOptions` constructs. I somewhat understand your reasoning (but not really, as the user experience is the same), but need to be on the record that I'm not happy with it. Doing so added > 2000 lines of code, which seems like we're violating DRY principles in a big way. Without using traits or macros, I don't know to get rid of all the copy-and-pasting.

https://github.com/georust/gdal/pull/456/commits/ff8a2b614a0c64a2085eedf504116d8df03e44db

Do I need to remove all the situations where the options structs have `Option` fields, and explicitly default to what the GDAL documentation says are the internal defaults?

metasim:
@lnicola I believe I've applied all the changes requested [here](https://github.com/georust/gdal/pull/456#issuecomment-1801962818).

src/raster/processing/dem/options.rs
```diff
@@ -0,0 +1,172 @@
...
+use gdal_sys::{
+    GDALDEMProcessingOptions, GDALDEMProcessingOptionsFree, GDALDEMProcessingOptionsNew,
+};
+
+use crate::cpl::CslStringList;
+use crate::errors;
+use crate::utils::_last_null_pointer_err;
+
+/// Payload for [`GDALDEMProcessing`]. Intended for internal use only.
+pub struct GdalDEMProcessingOptions<'opts>(
```
metasim, line 15: I didn't mark this as `pub(crate)` in case someone else needed it for a missed binding on a new feature, but maybe it should be?

lnicola: I think it's fine to make it public, but are you sure it needs the lifetime? Looking at the [GDAL code](https://github.com/OSGeo/gdal/blob/59eaaed3168f49e8a7a3821730277aff68a86d16/apps/gdaldem_lib.cpp#L3770), it seems to copy what it needs out of `papszArgv`. So we can probably drop the `PhantomData` and `'opts`.

src/raster/processing/dem/options.rs
```diff
@@ -0,0 +1,172 @@
...
+            self
+        }
+
+        /// Fetch additional options.
+        pub fn additional_options(&self) -> &CslStringList {
+            &self.additional_options
+        }
+
+        /// Private utility to convert common options into [`CslStringList`] options.
+        fn store_common_options_to(&self, opts: &mut CslStringList) {
```
lnicola, line 150: We can return a `Result<()>` from this one and `expect` on the caller side instead of doing 5 `unwrap`s.

src/raster/processing/dem/options.rs
```diff
@@ -0,0 +1,172 @@
...
+        /// ...
+        /// ```
+        ///
+        pub fn with_output_format(&mut self, format: &str) -> &mut Self {
+            self.output_format = Some(format.to_owned());
+            self
+        }
+
+        /// Fetch the specified output format driver identifier.
+        pub fn output_format(&self) -> Option<String> {
```
lnicola, line 120: I think it's fine to return `&str` for these, the caller probably won't care that much.

src/raster/processing/dem/options.rs
```diff
@@ -0,0 +1,172 @@
...
+    PhantomData<&'opts CslStringList>,
+);
+
+impl<'opts> GdalDEMProcessingOptions<'opts> {
+    pub fn new(opts: &'opts CslStringList) -> errors::Result<Self> {
+        let popts = unsafe { GDALDEMProcessingOptionsNew(opts.as_ptr(), ptr::null_mut()) };
+        if popts.is_null() {
+            return Err(_last_null_pointer_err("GDALDEMProcessingOptionsNew"));
+        }
+        Ok(Self(unsafe { NonNull::new_unchecked(popts) }, PhantomData))
```
lnicola, line 26: Let's use `NonNull::new` and match on that, it's not uglier than checking `popts`. We could even use https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html, but only after a MSRV bump (we're on 1.58? what happened to "`stable` + 3" or whatever policy we had?).

src/raster/processing/dem/aspect.rs
```diff
@@ -0,0 +1,125 @@
...
+
+/// Configuration options for [`aspect()`][super::aspect()].
+#[derive(Debug, Clone, Default)]
+pub struct AspectOptions {
+    input_band: Option<NonZeroUsize>,
+    compute_edges: bool,
+    output_format: Option<String>,
+    additional_options: CslStringList,
+    algorithm: Option<DemSlopeAlg>,
+    zero_for_flat: Option<bool>,
```
lnicola, line 14: I'm inclined to say we should avoid `Option` for most, if not all of these. I understand that the default might change, but if `-zero_for_flat` ever becomes the default, there will be a new `-non_zero_for_flat` and we'll have to update the code anyway. That is,

```rust
        if self.zero_for_flat == Some(true) {
            opts.add_string("-zero_for_flat").unwrap();
        }
```

already assumes that the default is `false`.

But we can probably change them later, you must be sick of this PR by now.

src/raster/processing/dem/aspect.rs
```diff
@@ -0,0 +1,125 @@
...
+impl AspectOptions {
+    /// Create a DEM-aspect options set.
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    common_dem_options!();
+
+    /// Specify the slope computation algorithm.
+    pub fn with_algorithm(&mut self, algorithm: DemSlopeAlg) -> &mut Self {
```
lnicola, line 27: We don't have getters for these, but the common options have them.

src/raster/processing/dem/aspect.rs
```diff
@@ -0,0 +1,125 @@
...
+
+    /// Return trigonometric angle instead of azimuth. Thus 0° means East, 90° North, 180° West, 270° South.
+    pub fn with_trigonometric_angles(&mut self, state: bool) -> &mut Self {
+        self.trigonometric = Some(state);
+        self
+    }
+
+    /// Render relevant common options into [`CslStringList`] values, as compatible with
+    /// [`gdal_sys::GDALDEMProcessing`].
+    pub fn to_options_list(&self) -> CslStringList {
```
lnicola, line 48: Same as for `store_common_options_to`, but more user-facing we could return `Result<CslStringList>` here. These are unlikely to fail but might, for example on OOM.

src/raster/processing/dem/options.rs
```diff
@@ -0,0 +1,172 @@
...
+    Tpi,
+    Tri,
+}
+
+impl Display for DemAlg {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::ColorRelief => f.write_str("color-relief"),
+            _ => {
+                let s = format!("{self:?}").to_lowercase();
```
lnicola, line 57: Man page says `TPI` and `TRI`, if this works, then I suspect GDAL isn't that case-sensitive here.

src/raster/processing/dem/options.rs
lnicola: https://github.com/metasim/gdal/pull/1

src/raster/processing/dem/options.rs
lnicola: https://github.com/metasim/gdal/pull/1

src/raster/processing/dem/options.rs
lnicola: https://github.com/metasim/gdal/pull/1

src/raster/processing/dem/aspect.rs
lnicola: https://github.com/metasim/gdal/pull/1

src/raster/processing/dem/options.rs
lnicola: https://github.com/metasim/gdal/pull/1

src/raster/processing/dem/options.rs
lnicola: https://github.com/metasim/gdal/pull/1/commits/7d2e1dd6a8a0fc95e3497c521379daffedcd1ec4

lnicola:
And one final question is where to put these. They're currently under `raster::processing::dem`, but we also have `programs::raster`.

Another aspect is that I think the Rust guidelines encourage tall/deep module hierarchies, but with re-exports so that the user only sees shallow ones.

metasim:
> And one final question is where to put these. They're currently under `raster::processing::dem`, but we also have `programs::raster`. I think I like `programs` more, since it's a little weird to have `raster::processing::translate` and `vector::processing::vector_translate`.

The rationale for this is two fold:

1. I plan to work on the Warp subsystem, and slotted it to live in `raster::processing:warp`. But that's separate from the export structure. See below.
2. I don't think `programs` makes any sense in a library. If I'm looking at a crate structure, I'd expect `programs` be the same as `bins`: a place to put source for executable targets. From the outside I wouldn't know how to semantically interpret `programs`. Heck, even from the inside I don't know how to interpret `programs`. Long term, I think the contents of the `programs` module should be migrated to something more logical.

On your point about `raster::processing` vs `programs::raster`, I'd argue the latter breaks the convention in the main library by reversing scope. We train users to think "domain first, capability second", but with `programs::raster` it's "if I don't find it in `raster::`, I should also look in `programs::raster`. But because I find `programs` confusing, perhaps I'm missing a larger point.

I can see arguments against `raster::processing`, but to fix its weaknesses I think we need to think more holistically about module structure and what heuristics we use to determine the proper destination for a capability. If the answer is "do exactly what is done in C++" (which would be valid) then we have even more of a gap to address. But I think we can do better than that. To me, `raster::processing` is a place for advanced capabilities beyond opening and querying `Dataset`s.

> Another aspect is that I think the Rust guidelines encourage tall/deep module hierarchies, but with re-exports so that the user only sees shallow ones.

My intent was the use would `use gdal::raster::processing::dem::*`, since DEM stuff is kinda niche. But I see an argument that this is too deep. I don't really think it should be exported at the `gdal::raster::*` level. Again, `programs::raster` doesn't makes sense to me as "what's a 'program' here?".

How about this: we close the PR as it is (or tweak the `pub use` structure), and open a new ticket to discuss a design around the organization of all constituent parts?

src/raster/processing/dem/options.rs
metasim: > it seems to copy what it needs out of papszArgv.

I was attempting to avoid a copy, since it's possible (at least in my use case) to construct the options and rn them over a bunch of files. But I guess a copy is in the performance noise.

src/raster/processing/dem/options.rs
metasim: I thought of this more like using `unsafe`: As a human with a static set of invariants, I'm guaranteeing those calls will never fail.

src/raster/processing/dem/options.rs
metasim: Nice!

src/raster/processing/dem/aspect.rs
metasim: It was oversight, but we're removing getters, right?

src/raster/processing/dem/options.rs
lnicola: Not sure I understand. Even after https://github.com/metasim/gdal/pull/1/commits/f39e73d83c3b084e633e9f86f434c2b073e0c31a, it's still possible to reuse a `GdalDEMProcessingOptions`. The copy is made by `GDALDEMProcessingOptionsNew`, and I'm saying that we don't need to keep a borrow to the `CslStringList`.

lnicola:
> I don't think programs makes any sense in a library. If I'm looking at a crate structure, I'd expect programs be the same as bins: a place to put source for executable targets. From the outside I wouldn't know how to semantically interpret programs. Heck, even from the inside I don't know how to interpret programs. Long term, I think the contents of the programs module should be migrated to something more logical.

I agree it's unusual, but many of the GDAL ["apps"](https://github.com/OSGeo/gdal/tree/master/apps) are already exposed as a library (everything that ends up taking an `argv` as an input, including `GDALDEMProcessing`. I'm not that familiar with them, but I think they're split between `gdal_alg.h` and `gdal_utils.h` in the GDAL headers.

> But because I find programs confusing, perhaps I'm missing a larger point.

My thinking is "I want `gdalwarp`, that's a "program". I want (to give a random example) `ExecuteSQL`, that's not a program". Even today, `raster` is a bit of a kitchen sink: https://docs.rs/gdal/latest/gdal/raster/index.html.

> How about this: we close the PR as it is (or tweak the pub use structure), and open a new ticket to discuss a design around the organization of all constituent parts?

That will never get anywhere, so please no :stuck_out_tongue:.

src/raster/processing/dem/options.rs
lnicola: Yeah, same thing as `dataset.rasterband(0)`. But I think it looks cleaner like this, and it will never matter in practice.

src/raster/processing/dem/options.rs
metasim: Yes, you're correct. Sorry, I missed the implications of where the copy was happening in the C++ code.

metasim:
> "I want `gdalwarp`, that's a "program".

`gdalwarp` barely exposes the potential of the Warp subsystem. After reading through a lot of the GDAL Warp code over the last year, I think it's potential is barely tapped by the `gdalwarp` executable, and limited by the lack of examples and complex options model. It may be wishful/idealistic thinking on my part, but with Rust, we could expose a lot of the processing potential in a much less intimidating way than it it is in C++. Rust enabled pixel functions alone could be very powerful. That's my dream anyway.

src/raster/processing/dem/mod.rs
```diff
@@ -22,6 +22,8 @@
 //! * [`topographic_position_index()`]
 //!
 
+#![deny(missing_docs)]
```
metasim, line 25: 👍

metasim:
bors r=lincola

PR #457, opened 2023-10-26
Added `Extend` and various `FromIterator` implementations to `CslStringList`.

metasim:
---

Addresses some leftover feedback from [here](https://github.com/georust/gdal/pull/455#discussion_r1371744038).

The docs summarize what this PR enables:

image


metasim:
@lnicola @jdroenner r+?

lnicola:
Sorry, I was away for the past couple of days. I'll take a look tomorrow.

fixtures/tinymarble.tif.aux.xml
```diff
@@ -1,3 +1,12 @@
 <PAMDataset>
   <SRS dataAxisToSRSAxisMapping="2,1">GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AXIS["Latitude",NORTH],AXIS["Longitude",EAST],AUTHORITY["EPSG","4326"]]</SRS>
+  <PAMRasterBand band="1">
```
lnicola, line 3: I don't remember if we want this or not, but perhaps it's harmless.

src/options.rs
```diff
@@ -2,7 +2,7 @@ use bitflags::bitflags;
 use gdal_sys::GDALAccess;
 use libc::c_uint;
 
-/// Open options for [Dataset]
+/// Open options for [crate::Dataset]
```
lnicola, line 5:
```suggestion
/// Open options for [`crate::Dataset`]
```

src/cpl.rs
```diff
@@ -324,30 +358,46 @@ impl FromStr for CslStringList {
...
+
+impl FromIterator<String> for CslStringList {
+    fn from_iter<T: IntoIterator<Item = String>>(iter: T) -> Self {
+        iter.into_iter()
+            .map(Into::<CslStringListEntry>::into)
+            .collect()
+    }
+}
+
+impl Extend<CslStringListEntry> for CslStringList {
```
lnicola, line 387: Too bad we can't use `CSLMerge` here.

src/cpl.rs
```diff
@@ -324,30 +358,46 @@ impl FromStr for CslStringList {
...
+    fn extend<T: IntoIterator<Item = CslStringListEntry>>(&mut self, iter: T) {
+        for e in iter {
+            self.add_entry(&e).unwrap_or_default();
```
lnicola, line 390: Oof :pensive:.

src/cpl.rs
```diff
@@ -358,22 +408,31 @@ pub enum CslStringListEntry {
...
+impl From<String> for CslStringListEntry {
+    fn from(value: String) -> Self {
+        match value.split_once('=') {
+            Some(kv) => kv.into(),
+            None => Self::Flag(value),
         }
     }
 }
 
+impl From<(String, String)> for CslStringListEntry {
```
lnicola, line 430: `CslStringListEntry::pair(key, value)` might be nice to have.

lnicola (approved):
The error handling makes this a bit icky, but I don't have any good suggestions.

I think it's workable, if not ideal, let's see how jdroenner, feels about it.

src/cpl.rs
metasim: I actually started to implement it (and still can) but the [C++ implementation](https://github.com/OSGeo/gdal/blob/27e2cf443dea54b5fcb60fc2900480b7ef866448/port/cpl_string.cpp#L273) scared me as I saw no evidence that it would extend the length of the original array. That said, my C++ is 20 years rusty.

lnicola: That function just appends one element at a time.

src/cpl.rs
metasim: @lnicola Here's my rational as to why this might be OK.
Most of the time you're going to be extending one `CslStringList` with another. The only way that call can fail is if your insertion `String` has `\r`, `\n`, or most importantly `\0`. If `iter` comes from a `CslStringList` then that validation has already happened. 

However, if `iter` comes from `[CslStringListEntry]`, then you could presumably circumvent that check. I added [this comment](https://github.com/georust/gdal/pull/457/files#diff-735b4e8f56e7d6ad6abe40c99808df64a2b2e05b2b13f976303cefe2dca42affR399-R400) for that reason.

The obvious alternative means no `Extend`, and `TryFrom`s to cover what is probably a rare edge case. On balance, I'd rather have `Extend`.

lnicola: Exactly. OOM might be another case where it fails, but it's pretty unlikely.

src/cpl.rs
metasim: Couldn't find a `{re|m}alloc` in there or in `CSLSetNameValue` either.

src/cpl.rs
metasim: FWIW, Added additional documentation to `CslStringListEntry::new_{pair|flag}`.

src/cpl.rs
lnicola: I think it ends up in https://github.com/OSGeo/gdal/blob/27e2cf443dea54b5fcb60fc2900480b7ef866448/port/cpl_string.cpp#L111.

metasim:
@jdroenner r+?

src/cpl.rs
```diff
@@ -324,30 +358,46 @@ impl FromStr for CslStringList {
...
     }
 }
 
-/// Represents an entry in a [`CslStringList`], which is ether a single token (`Flag`),
-/// or a `name=value` assignment (`Pair`).
+/// Represents an entry in a [`CslStringList`]
+///
+/// An Entry is ether a single token (`Flag`), or a `name=value` assignment (`Pair`).
+///
+/// Note: When constructed directly, assumes string values do not contain newline characters nor
```
lnicola, line 399: We should probably spell out more clearly that those values will be ignored. And not just here, but also on `CslStringList`.

lnicola (approved):

src/cpl.rs
metasim: `CslStringList` actually propagates the error.
https://github.com/georust/gdal/blob/16f2aefd16e7d01cf27e2fce680e7eed93a7a5f2/src/cpl.rs#L116-L117

metasim:
bors r=lnicola

metasim:
bors try

PR #468, opened 2023-11-19
Add histogram calculation for RasterBand

spadarian:
---

This PR exposes 2 GDAL methods to calculate band histograms (default and user-defined).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -50,6 +50,8 @@
 
   - <https://github.com/georust/gdal/pull/439>
 
+- Added Histogram calculation
+
 ## 0.16
 
 - **Breaking**: `Dataset::close` now consumes `self`
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -52,6 +52,8 @@
 
 - Added Histogram calculation
 
+  - <https://github.com/georust/gdal/pull/468>
+
 ## 0.16
 
 - **Breaking**: `Dataset::close` now consumes `self`
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -50,7 +50,7 @@
 
   - <https://github.com/georust/gdal/pull/439>
 
-- Added Histogram calculation
+- Added histogram calculation
 
   - <https://github.com/georust/gdal/pull/468>
 
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -5,9 +5,10 @@ use crate::raster::{GdalDataType, GdalType};
 use crate::utils::{_last_cpl_err, _last_null_pointer_err, _string};
 use gdal_sys::{
     self, CPLErr, GDALColorEntry, GDALColorInterp, GDALColorTableH, GDALComputeRasterMinMax,
-    GDALCreateColorRamp, GDALCreateColorTable, GDALDestroyColorTable, GDALGetPaletteInterpretation,
-    GDALGetRasterStatistics, GDALMajorObjectH, GDALPaletteInterp, GDALRIOResampleAlg, GDALRWFlag,
-    GDALRasterBandH, GDALRasterIOExtraArg, GDALSetColorEntry, GDALSetRasterColorTable,
+    GDALCreateColorRamp, GDALCreateColorTable, GDALDestroyColorTable, GDALGetDefaultHistogram,
+    GDALGetPaletteInterpretation, GDALGetRasterStatistics, GDALMajorObjectH, GDALPaletteInterp,
+    GDALRIOResampleAlg, GDALRWFlag, GDALRasterBandH, GDALRasterIOExtraArg, GDALSetColorEntry,
+    GDALSetRasterColorTable,
 };
 use libc::c_int;
 use std::ffi::CString;
@@ -847,6 +848,96 @@ impl<'a> RasterBand<'a> {
             max: min_max[1],
         })
     }
+
+    /// Fetch default raster histogram.
+    ///
+    /// # Arguments
+    ///
+    /// * `force` - If `true`, force the computation. If `false` and no default histogram is available, the method will return None
+    pub fn get_default_histogram(&self, force: bool) -> Result<Option<Histogram>> {
+        let mut hist = Histogram {
+            min: 0.0,
+            max: 0.0,
+            n_buckets: 0,
+            histogram: std::ptr::null_mut(),
+            histogram_vec: None,
+        };
+
+        let rv = unsafe {
+            GDALGetDefaultHistogram(
+                self.c_rasterband,
+                &mut hist.min,
+                &mut hist.max,
+                &mut hist.n_buckets,
+                &mut hist.histogram as *mut *mut i32,
+                libc::c_int::from(force),
+                None,
+                std::ptr::null_mut(),
+            )
+        };
+
+        match CplErrType::from(rv) {
+            CplErrType::None => {
+                // if everything when ok, can this still be null?
+                if hist.histogram.is_null() {
+                    Ok(None)
+                } else {
+                    Ok(Some(hist))
+                }
+            }
+            CplErrType::Warning => Ok(None),
+            _ => Err(_last_cpl_err(rv)),
+        }
+    }
+
+    /// Compute raster histogram.
+    ///
+    /// # Arguments
+    ///
+    /// * `min` - Histogram lower bound
+    /// * `max` - Histogram upper bound
+    /// * `n_buckets` - Number of buckets in the histogram
+    /// * `include_out_of_range` - if `true`, values below the histogram range will be mapped into the first bucket, and values above will be mapped into the last one. If `false`, out of range values are discarded
+    /// * `is_approx_ok` - If an approximate, or incomplete histogram is OK
+    pub fn get_histogram(
+        &self,
+        min: f64,
+        max: f64,
+        n_buckets: i32,
+        include_out_of_range: bool,
+        is_approx_ok: bool,
+    ) -> Result<Option<Histogram>> {
+        let mut hist = Histogram {
+            min,
+            max,
+            n_buckets,
+            histogram_vec: Some(vec![0; n_buckets as usize]),
+            histogram: std::ptr::null_mut(),
+        };
+
+        let rv = unsafe {
+            gdal_sys::GDALGetRasterHistogram(
+                self.c_rasterband,
+                min,
+                max,
+                n_buckets,
+                hist.histogram_vec.as_mut().unwrap().as_mut_ptr(),
+                libc::c_int::from(include_out_of_range),
+                libc::c_int::from(is_approx_ok),
+                None,
+                std::ptr::null_mut(),
+            )
+        };
+
+        match CplErrType::from(rv) {
+            CplErrType::None => {
+                hist.histogram = hist.histogram_vec.as_mut().unwrap().as_mut_ptr();
+                Ok(Some(hist))
+            }
+            CplErrType::Warning => Ok(None),
+            _ => Err(_last_cpl_err(rv)),
+        }
+    }
 }
 
 #[derive(Debug, PartialEq)]
@@ -861,6 +952,50 @@ pub struct StatisticsAll {
     pub max: f64,
     pub mean: f64,
     pub std_dev: f64,
+}
+
+#[derive(Debug)]
+pub struct Histogram {
+    min: f64,
+    max: f64,
+    n_buckets: i32,
+    histogram: *mut i32,
+    histogram_vec: Option<Vec<i32>>,
+}
+
+impl Histogram {
+    /// Histogram lower bound
+    pub fn min(&self) -> f64 {
+        self.min
+    }
+
+    /// Histogram upper bound
+    pub fn max(&self) -> f64 {
+        self.max
+    }
+    /// Histogram values for each bucket
+    pub fn histogram(&self) -> &[i32] {
+        if let Some(hist) = self.histogram_vec.as_ref() {
+            return hist.as_slice();
+        }
+
+        unsafe { std::slice::from_raw_parts(self.histogram, self.n_buckets as usize) }
+    }
+
+    /// Histogram bucket size
+    pub fn bucket_size(&self) -> f64 {
+        (self.max - self.min) / self.histogram().len() as f64
+    }
+}
+
+impl Drop for Histogram {
+    fn drop(&mut self) {
+        if self.histogram_vec.is_none() {
+            unsafe {
+                gdal_sys::VSIFree(self.histogram as *mut libc::c_void);
+            }
+        }
+    }
 }
 
 impl<'a> MajorObject for RasterBand<'a> {
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -5,10 +5,10 @@ use crate::raster::{GdalDataType, GdalType};
 use crate::utils::{_last_cpl_err, _last_null_pointer_err, _string};
 use gdal_sys::{
     self, CPLErr, GDALColorEntry, GDALColorInterp, GDALColorTableH, GDALComputeRasterMinMax,
-    GDALCreateColorRamp, GDALCreateColorTable, GDALDestroyColorTable, GDALGetDefaultHistogram,
-    GDALGetPaletteInterpretation, GDALGetRasterStatistics, GDALMajorObjectH, GDALPaletteInterp,
-    GDALRIOResampleAlg, GDALRWFlag, GDALRasterBandH, GDALRasterIOExtraArg, GDALSetColorEntry,
-    GDALSetRasterColorTable,
+    GDALCreateColorRamp, GDALCreateColorTable, GDALDestroyColorTable, GDALGetDefaultHistogramEx,
+    GDALGetPaletteInterpretation, GDALGetRasterHistogramEx, GDALGetRasterStatistics,
+    GDALMajorObjectH, GDALPaletteInterp, GDALRIOResampleAlg, GDALRWFlag, GDALRasterBandH,
+    GDALRasterIOExtraArg, GDALSetColorEntry, GDALSetRasterColorTable,
 };
 use libc::c_int;
 use std::ffi::CString;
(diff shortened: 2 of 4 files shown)


src/raster/rasterband.rs
```diff
@@ -847,6 +848,96 @@ impl<'a> RasterBand<'a> {
...
+    /// Compute raster histogram.
+    ///
+    /// # Arguments
+    ///
+    /// * `min` - Histogram lower bound
+    /// * `max` - Histogram upper bound
+    /// * `n_buckets` - Number of buckets in the histogram
+    /// * `include_out_of_range` - if `true`, values below the histogram range will be mapped into the first bucket, and values above will be mapped into the last one. If `false`, out of range values are discarded
+    /// * `is_approx_ok` - If an approximate, or incomplete histogram is OK
+    pub fn get_histogram(
```
metasim, line 902: I'd drop the `get_` prefix to be more idiomatic.

src/raster/rasterband.rs
```diff
@@ -863,6 +954,50 @@ pub struct StatisticsAll {
...
+    pub fn min(&self) -> f64 {
+        self.min
+    }
+
+    /// Histogram upper bound
+    pub fn max(&self) -> f64 {
+        self.max
+    }
+    /// Histogram values for each bucket
+    pub fn histogram(&self) -> &[i32] {
```
metasim, line 977: Consider naming `buckets` or `counts`? With this name, you'll end up with things like this:

```rust
let rb: Rasterband<'_> = ...;
let histogram = rb.get_histogram(...);

for count in histogram.histogram() {
...
}
```

src/raster/rasterband.rs
```diff
@@ -863,6 +954,50 @@ pub struct StatisticsAll {
     pub std_dev: f64,
 }
 
+#[derive(Debug)]
+pub struct Histogram {
+    min: f64,
+    max: f64,
+    n_buckets: i32,
```
metasim, line 961: Why isn't this `usize`? Sometimes GDAL uses signed values when it doesn't make semantic sense, and there's [precedence](https://github.com/georust/gdal/blob/f180427c099f0093f7b8b10111b556517ee7cba9/src/raster/rasterband.rs#L83-L85) for rectifying this in Rust.

src/raster/rasterband.rs
```diff
@@ -863,6 +954,50 @@ pub struct StatisticsAll {
...
+    histogram: *mut i32,
+    histogram_vec: Option<Vec<i32>>,
```
metasim, line 963: Is this storing the counts twice? Who owns `histogram`?

src/raster/rasterband.rs
```diff
@@ -847,6 +848,96 @@ impl<'a> RasterBand<'a> {
...
+            histogram: std::ptr::null_mut(),
+        };
+
+        let rv = unsafe {
+            gdal_sys::GDALGetRasterHistogram(
+                self.c_rasterband,
+                min,
+                max,
+                n_buckets,
+                hist.histogram_vec.as_mut().unwrap().as_mut_ptr(),
```
metasim, line 924: Instead of `unwrap` can you use `?`?

src/raster/rasterband.rs
```diff
@@ -847,6 +848,96 @@ impl<'a> RasterBand<'a> {
             max: min_max[1],
         })
     }
+
+    /// Fetch default raster histogram.
+    ///
+    /// # Arguments
+    ///
+    /// * `force` - If `true`, force the computation. If `false` and no default histogram is available, the method will return None
+    pub fn get_default_histogram(&self, force: bool) -> Result<Option<Histogram>> {
```
metasim, line 857: For coherence with other methods/functions, I'd suggest collapsing these two methods into a single one accepting a new `HistogramOptions` with the options, and implement `Default` for it.

src/raster/rasterband.rs
```diff
@@ -847,6 +848,96 @@ impl<'a> RasterBand<'a> {
...
+        };
+
+        let rv = unsafe {
+            GDALGetDefaultHistogram(
+                self.c_rasterband,
+                &mut hist.min,
+                &mut hist.max,
+                &mut hist.n_buckets,
+                &mut hist.histogram as *mut *mut i32,
+                libc::c_int::from(force),
```
metasim, line 873: Nice. I didn't realize `libc::c_int::from` existed.

src/raster/rasterband.rs
```diff
@@ -863,6 +954,50 @@ pub struct StatisticsAll {
...
+    pub fn bucket_size(&self) -> f64 {
+        (self.max - self.min) / self.histogram().len() as f64
+    }
+}
+
+impl Drop for Histogram {
+    fn drop(&mut self) {
+        if self.histogram_vec.is_none() {
+            unsafe {
+                gdal_sys::VSIFree(self.histogram as *mut libc::c_void);
```
metasim, line 995: If it turns out that double storage mechanisms are indeed necessary, please document the reason and any associated assumptions. I'd personally go further and create an `enum` capturing the two storage cases to enforce the logic with types, e.g.:

```
enum HistCounts {
   UnsafePointer(*mut libc::c_void),
   OwnedVec(Vec<i32>)
}
```

But I think unifying it is preferable.

src/raster/rasterband.rs
metasim, line 963: I personally think the counts should be `usize` or `u32`, regardless of what GDAL C returns, but not sure that's a consensus opinion. @lnicola @jdroenner you opinion?

CHANGES.md
```diff
@@ -45,6 +45,8 @@
 
   - <https://github.com/georust/gdal/pull/439>
 
+- Added Histogram calculation
+
```
metasim, line 49: Pealse add PR link here, as other entries do.

src/raster/rasterband.rs
spadarian: I also found it weird but I was just being consistent with GDAL naming. Happy to change it if we are not aiming for that.

src/raster/rasterband.rs
spadarian: I haven't contributed for a while and from memory we were using `isize`. I do agree that we could rectify that.

How far do we want to take it? I could also use a `NonZeroUsize`, which could be also useful when selecting a band from a dataset... (@lnicola)

src/raster/rasterband.rs
spadarian: Yes, sorry. I should document that. I used that double storage mechanisms because with `GDALGetDefaultHistogram` C owns histogram and with `GDALGetRasterHistogram` Rust does.

src/raster/rasterband.rs
metasim: Fair point. Let's hear what some of the other contributors think. I'm on the extreme of "let's not leak C-language workarounds" and "let's leverage Rust to improve ergonomics". Others, I think, are more in your camp, so let's let someone else make the call. We've had related debates [here](https://github.com/georust/gdal/pull/456).

src/raster/rasterband.rs
metasim: Lol...  yeah, I've argued for `NonZeroUsize`, but creating them is really annoying/verbose.

When I see `isize` or `i32` I immediately ask myself "What does a negative value mean?". If it has meaning, we should document it. If it doesn't, then I'd at least try in the bindings to coerce it into an unsigned type.

src/raster/rasterband.rs
spadarian: No worries mate. I'm pushing those changes now.

spadarian:
I've incorporated some of the changes suggested by @metasim, including:
- Double storage mechanism for Borrowed and Owned histogram counts (makes sense and it is hidden from the user)
- Changing isize to usize since we recently changed that for [raster indexing](https://github.com/georust/gdal/commit/072773a36b96e90a384e3d4a79c3f4ca9df89710)

I did not merge the calls for `GDALGetDefaultHistogram` and `GDALGetRasterHistogram` since I think having a API more or less consistent with the C code is desirable. Open to change it if that is where the project is going @lnicola @jdroenner

src/raster/rasterband.rs
```diff
@@ -863,6 +945,55 @@ pub struct StatisticsAll {
...
+    Borrowed(*mut i32),
+    Owned(Vec<i32>),
+}
+
+impl Drop for HistCounts {
+    fn drop(&mut self) {
+        match self {
+            HistCounts::Borrowed(p) => unsafe {
+                gdal_sys::VSIFree(*p as *mut libc::c_void);
+            },
```
metasim, line 991: @spadarian This is where I'm still confused.... if it were truly borrowed, it means it's owned by the C runtime, and we shouldn't be freeing it? That's what I meant earlier about not being sure if I had the owned vs borrowed cases correct.

spadarian: I don't know... not an expert in these things. The C documentation says: "@param ppanHistogram pointer to array into which the histogram totals are placed. To be freed with VSIFree" and we expose the `VSIFree` function, so I assumed that it is a way of saying to the C runtime that we don't need that anymore, which is what happen when we drop it.

If it is not needed, I can remove it.

metasim: Great, that helps. It's needed, but the names `Borrowed` and `Owned` need to be something else. I'll try to make time today to come up with an alternative proposal. Need to spend some time with the docs for `GDALGetRasterHistogram` and `GDALGetDefaultHistogram` first. I think you're close here.

src/raster/rasterband.rs
```diff
@@ -847,6 +848,87 @@ impl<'a> RasterBand<'a> {
             max: min_max[1],
         })
     }
+
+    /// Fetch default raster histogram.
+    ///
+    /// # Arguments
+    ///
+    /// * `force` - If `true`, force the computation. If `false` and no default histogram is available, the method will return None
+    pub fn default_histogram(&self, force: bool) -> Result<Option<Histogram>> {
```
metasim, line 857: Can we document what returning `Ok(None)` means?

src/raster/rasterband.rs
```diff
@@ -847,6 +848,87 @@ impl<'a> RasterBand<'a> {
...
+    /// * `include_out_of_range` - if `true`, values below the histogram range will be mapped into the first bucket, and values above will be mapped into the last one. If `false`, out of range values are discarded
+    /// * `is_approx_ok` - If an approximate, or incomplete histogram is OK
+    pub fn histogram(
+        &self,
+        min: f64,
+        max: f64,
+        n_buckets: usize,
+        include_out_of_range: bool,
+        is_approx_ok: bool,
+    ) -> Result<Option<Histogram>> {
```
metasim, line 904: Can we document what returning `Ok(None)` means?

src/raster/rasterband.rs
lnicola: Yeah, I wish constructing a `NonZeroUsize` from a constant/literal was possible without `unsafe` or bloating the binary. :pensive:

src/raster/rasterband.rs
```diff
@@ -847,6 +848,87 @@ impl<'a> RasterBand<'a> {
...
+    ///
+    /// * `force` - If `true`, force the computation. If `false` and no default histogram is available, the method will return None
+    pub fn default_histogram(&self, force: bool) -> Result<Option<Histogram>> {
+        let mut counts = std::ptr::null_mut();
+        let mut min = 0.0;
+        let mut max = 0.0;
+        let mut n_buckets = 0i32;
+
+        let rv = unsafe {
+            GDALGetDefaultHistogram(
```
lnicola, line 864: I think we should use `GDALGetDefaultHistogramEx` instead? It's safer, and I don't think the memory usage will be that much of an issue.

src/raster/rasterband.rs
```diff
@@ -847,6 +848,87 @@ impl<'a> RasterBand<'a> {
...
+        min: f64,
+        max: f64,
+        n_buckets: usize,
+        include_out_of_range: bool,
+        is_approx_ok: bool,
+    ) -> Result<Option<Histogram>> {
+        let mut counts = vec![0; n_buckets];
+
+        let rv = unsafe {
+            gdal_sys::GDALGetRasterHistogram(
```
lnicola, line 908: Same for `GDALGetRasterHistogramEx`.

CHANGES.md
```diff
@@ -45,6 +45,10 @@
 
   - <https://github.com/georust/gdal/pull/439>
 
+- Added Histogram calculation
```
lnicola, line 48: Nit:

```suggestion
- Added histogram calculation
```

src/raster/rasterband.rs
lnicola: Yeah, https://github.com/OSGeo/gdal/blob/59eaaed3168f49e8a7a3821730277aff68a86d16/gcore/gdal.h#L1561.

src/raster/rasterband.rs
lnicola: I really hate how these `bForce` parameters work. If I'm reading the code right, even with `bForce` you can still get a saved histogram from the PAM dataset (`.aux.xml`). I wish `!bForce` meant "return cached or compute it" and `bForce` meant "always recompute".

src/raster/rasterband.rs
```diff
@@ -863,6 +945,55 @@ pub struct StatisticsAll {
...
+    /// Histogram lower bound
+    pub fn min(&self) -> f64 {
+        self.min
+    }
+
+    /// Histogram upper bound
+    pub fn max(&self) -> f64 {
+        self.max
+    }
+    /// Histogram values for each bucket
```
lnicola, line 966: Nit: newline

src/raster/rasterband.rs
```diff
@@ -863,6 +945,55 @@ pub struct StatisticsAll {
...
+    }
+
+    /// Histogram bucket size
+    pub fn bucket_size(&self) -> f64 {
+        (self.max - self.min) / self.counts().len() as f64
+    }
+}
+
+#[derive(Debug)]
+enum HistCounts {
```
lnicola, line 981: Nit: my slight preference is to call this `HistogramCounts`.

src/raster/rasterband.rs
```diff
@@ -863,6 +945,55 @@ pub struct StatisticsAll {
...
+        }
+    }
+
+    /// Histogram bucket size
+    pub fn bucket_size(&self) -> f64 {
+        (self.max - self.min) / self.counts().len() as f64
+    }
+}
+
+#[derive(Debug)]
```
lnicola, line 980: We probably want a manually-written `Debug` to handle the `Borrowed` case.

lnicola: Also, I wonder if we should turn this into a more generic thing. We have something similar -- `GcpRef`, but it looks quite differently. Probably badly-named, too.

spadarian: What do you want exactly? At the moment it shows that it is a pointer (wrapped on GdalAllocated). Do you want it to show a Vec?

lnicola: As a slice, you can delegate to that. It's much more useful than showing a pointer.

src/raster/rasterband.rs
spadarian: In terms of the documentation, I have this in the docs for the `force` parameter:

```rust
/// * `force` - If `true`, force the computation. If `false` and no default histogram is available, the method will return `Ok(None)`
```

Is that enough or you prefer a Note or something similar?

Also, I just realised that histogram should not return None (`GDALGetRasterHistogramEx` does not emit a `CplErrType::Warning`).

src/raster/rasterband.rs
spadarian: @lnicola: I guess we could force it if we get None but that would diverge from the C API.

lnicola: No, let's follow the C API. I just like to complain.

spadarian:
I addressed all the comments... but CI failed on a unrelated test ( raster::processing::dem::slope::tests::test_slope)

lnicola:
> Error: CplError { class: 3, number: 1, msg: "/__w/gdal/gdal/target/dem-hills-slope.tiff, band 1: IReadBlock failed at X offset 0, Y offset 0: TIFFReadEncodedStrip() failed." }

:exclamation:

lnicola:
Ah, there's a race between `tpi` and `slope`. I can try to fix it tomorrow.

metasim:
@spadarian Will you accept this? https://github.com/spadarian/gdal/pull/12

lnicola:
@metasim isn't that already merged?

metasim:
> @metasim isn't that already merged?

Oops! Yes. Sorry. Was looking at a non-refreshed page.

rouault:
little robustness fix in https://github.com/spadarian/gdal/pull/13

spadarian:
Rebased @lnicola

src/raster/rasterband.rs
```diff
@@ -863,6 +948,82 @@ pub struct StatisticsAll {
...
+
+/// Union type over histogram storage mechanisms.
+///
+/// This private enum exists to normalize over the two different ways
+/// [`GDALGetRasterHistogram`] and [`GDALGetDefaultHistogram`] return data:
+/// * `GDALGetRasterHistogram`: requires a pre-allocated array, stored in `HistogramCounts::RustAllocated`.
+/// * `GDALGetDefaultHistogram`: returns a pointer (via an 'out' parameter) to a GDAL allocated array,
+///   stored in `HistogramCounts::GdalAllocated`, to be deallocated with [`VSIFree`][gdal_sys::VSIFree].
+enum HistogramCounts {
+    /// Pointer to GDAL allocated array and lenght of histogram counts.
```
lnicola, line 996:
```suggestion
    /// Pointer to GDAL allocated array and length of histogram counts.
```

src/raster/rasterband.rs
```diff
@@ -863,6 +948,82 @@ pub struct StatisticsAll {
...
+    ///
+    /// Requires freeing with [`VSIFree`][gdal_sys::VSIFree].
+    GdalAllocated(*mut u64, usize),
```
lnicola, line 999: @metasim this seems to use `u64` consistently for the counts, did you have some concerns about that?

src/raster/rasterband.rs
```diff
@@ -863,6 +948,82 @@ pub struct StatisticsAll {
...
+    }
+
+    /// Histogram upper bound
+    pub fn max(&self) -> f64 {
+        self.max
+    }
+
+    /// Histogram values for each bucket
+    pub fn counts(&self) -> &[u64] {
+        match &self.counts {
```
lnicola, line 971: Let's move this to `HistogramCounts::as_slice` and reuse it in the `Debug` implementation.

lnicola (approved):

PR #469, opened 2023-11-30
Fix tpi test output name and use `/vsimem/` for DEM tests

lnicola:
---

CC https://github.com/georust/gdal/pull/468#issuecomment-1832636108.


spadarian:
LGTM

lnicola:
Merging early so @spadarian can rebase over this.

PR #472, opened 2023-11-30
.github/workflows/ci.yml: run 'cargo valgrind test --lib'

rouault:
---

### DIFF
--- a/src/raster/mdarray.rs
+++ b/src/raster/mdarray.rs
@@ -163,14 +163,14 @@ impl<'a> MDArray<'a> {
         let n_dst_buffer_alloc_size = 0;
 
         let rv = unsafe {
-            let data_type = GDALExtendedDataTypeCreate(T::gdal_ordinal());
-
             if !self.datatype().class().is_numeric() {
                 return Err(GdalError::UnsupportedMdDataType {
                     data_type: self.datatype().class(),
                     method_name: "GDALMDArrayRead",
                 });
             }
+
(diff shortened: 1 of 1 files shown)


lnicola:
I expected this to be slower, but 30 seconds isn't bad at all.

lnicola (approved):

lnicola:
I'm thinking it might be a good idea to run this in the 3.8 container, since LTS might not have some of the APIs (like MD) we've implemented.

lnicola:
Great, we can disable it for now.

https://github.com/georust/gdal/tree/master/gdal-sys#build might help with linking.

lnicola:
Not sure if that, but we might be leaking the data type in the error case at https://github.com/georust/gdal/blob/master/src/raster/mdarray.rs#L166.

rouault:
ok, the leak in read_into_slice() is fixed now, but there's a "system leak" related to threads & TLS, that is IMHO a false positive. But I can't find an option to specify a Valgrind suppression file to cargo valgrind. I guess we need to skip 3.8.1 testing for now

lnicola:
Yeah, let's disable it. I'll merge this in the morning.

rouault:
> Yeah, let's disable it. I'll merge this in the morning.

I've dropped the 3.8.1 related commit

lnicola:
Funny, I don't get the TLS leak on my system (Gedora 39, GDAL 3.7.3).

I think we can pass in a suppressions file with `VALGRINDFLAGS="--suppressions=gdal.supp" cargo valgrind test`.

PR #481, opened 2023-12-01
MDArray::read_as_string_array(): make it robust to null string pointer being returned by GDAL

rouault:
---

I can't really come with a unit test for that given what exists in the bindings, but this is definitely something that could happen for example with the MEM driver where a user could write an array with null strings in it.

### DIFF
--- a/src/raster/mdarray.rs
+++ b/src/raster/mdarray.rs
@@ -330,11 +330,15 @@ impl<'a> MDArray<'a> {
             let strings = string_pointers
                 .into_iter()
                 .map(|string_ptr| {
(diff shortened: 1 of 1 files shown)


src/raster/mdarray.rs
```diff
@@ -330,11 +330,15 @@ impl<'a> MDArray<'a> {
             let strings = string_pointers
                 .into_iter()
                 .map(|string_ptr| {
-                    let string = _string(string_ptr);
+                    if string_ptr == std::ptr::null() {
```
lnicola, line 333:
```suggestion
                    if string_ptr.is_null() {
```

PR #486, opened 2023-12-03
Add method to set default histogram for RasterBand

spadarian:
---

For completeness, this PR adds method to set histogram for `RasterBand`.
Get methods were added by this PR: https://github.com/georust/gdal/pull/468


src/raster/rasterband.rs
```diff
@@ -884,6 +884,32 @@ impl<'a> RasterBand<'a> {
...
+    /// Set default raster histogram.
+    ///
+    /// # Arguments
+    ///
+    /// * `min` - Histogram lower bound
+    /// * `max` - Histogram upper bound
+    /// * `n_buckets` - Number of buckets in the histogram
+    /// * `counts` - Histogram values for each bucket
+    pub fn set_default_histogram(&self, min: f64, max: f64, counts: &mut [u64]) -> Result<()> {
+        let n_buckets = counts.len();
```
lnicola, line 896: Let's add an `assert!(n_buckets <= i32::MAX)` here and in the getter if missing.

(that matches `Dataset::rasterband`, there was a discussion on `Err` vs. `assert!` there)

lnicola (approved):

metasim (approved):
LGTM

PR #487, opened 2023-12-04
Add `Geometry::difference`

### LINKED ISSUE #432: FR: Wrapper for OGR_G_Difference
(Sorry if I'm getting annoying with these feature requests)

I needed to do a 'difference' on two polygon features and realized it wasn't implemented. The following is my helper trait to work around it, but in this case, because of private fields, I couldn't just copy the code in `union`, so I'm unsure if this is safe. (That's related to #306 )

```rust
pub(crate) trait GeometryFix: Sized {
    fn difference(&self, other: &Self) -> Option<Self>;
}

impl GeometryFix for Geometry {
    fn difference(&self, other: &Self) -> Option<Self>  {
        if !self.has_gdal_ptr() {
            return None;
        }
        if !other.has_gdal_ptr() {
            return None;
        }
        unsafe {
            let ogr_geom = gdal_sys::OGR_G_Difference(self.c_geometry(), other.c_geometry());
            if ogr_geom.is_null() {
                return None;
            }
            // TODO: Unfortunately, with_c_geometry is private, so I can't use it.
            let geometry = Self::lazy_feature_geometry();
            geometry.set_c_geometry(ogr_geom);
            // TODO: DANGER!: I can't set owned = true on the thing, there's no way.
            // However, I *think* cloning will take care of that. Because the original
            // value won't dereference the API handle, as it's not owned, but clone
            // will set it to owned.
            Some(geometry.clone())
        }
    }
}
```

lnicola:
---

Closes #432

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,9 +2,13 @@
 
 ## Unreleased
 
+- Added `Geometry::difference`.
+
+   - <https://github.com/georust/gdal/pull/487>
+
 - Added support for digital elevation model raster processing: `aspect`, `color_relief`, `hillshade`, `roughness`, `slope`, `terrain_ruggedness_index`, `topographic_position_index`.
 
-   - <https://github.com/georust/gdal/pull/456> 
+   - <https://github.com/georust/gdal/pull/456>
 
 - Added pre-built bindings for GDAL 3.8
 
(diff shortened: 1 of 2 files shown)


metasim (approved):
Wow... surprised we missed that the first time around. Nice.

lnicola:
We missed others too, like `OGR_G_SymDifference`, `OGR_G_UnaryUnion` and so on.

metasim:
> We missed others too, like OGR_G_SymDifference, OGR_G_UnaryUnion and so on.

Geeze.... PR author was a real slacker! ;-)

PR #490, opened 2023-12-10
Add `RasterBand::write_block`

lnicola:
---

### DIFF
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -530,7 +530,30 @@ impl<'a> RasterBand<'a> {
     /// * `block_index` - the block index
     ///
     /// # Notes
-    /// The Matrix shape is (rows, cols) and raster shape is (cols in x-axis, rows in y-axis).
+    /// Blocks indexes start from 0 and are of form (x, y), where x grows in the horizontal direction.
+    ///
+    /// The matrix shape is (rows, cols) and raster shape is (cols in x-axis, rows in y-axis).
+    ///
+    /// The block size of the band can be determined using [`RasterBand::block_size`].
+    /// The last blocks in both directions can be smaller.
+    /// [`RasterBand::actual_block_size`] will report the correct dimensions of a block.
+    ///
+    /// # Errors
+    /// If the block index is not valid, GDAL will return an error.
+    ///
+    /// # Example
+    ///
+    /// ```rust, no_run
+    /// # fn main() -> gdal::errors::Result<()> {
+    /// use gdal::Dataset;
+    ///
+    /// let dataset = Dataset::open("fixtures/m_3607824_se_17_1_20160620_sub.tif")?;
+    /// let band1 = dataset.rasterband(1)?;
+    /// let arr = band1.read_block::<u8>((0, 0))?;
+    /// assert_eq!(arr.shape(), &[300, 6]);
+    /// # Ok(())
+    /// # }
+    /// ```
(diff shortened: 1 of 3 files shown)


lnicola:
Will rebase on top of #489.

r? @metasim

metasim:
Should we also support `{read|write}_block` without `ndarray`?

src/raster/rasterband.rs
```diff
@@ -556,6 +562,41 @@ impl<'a> RasterBand<'a> {
         Array2::from_shape_vec((size.1, size.0), data).map_err(Into::into)
     }
 
+    #[cfg(feature = "ndarray")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "array")))]
+    /// Write a [`Array2<T>`] from a [`Dataset`] block, where `T` implements [`GdalType`].
+    ///
+    /// # Arguments
```
metasim, line 569: Can we add a reference to the method for getting the block layout (so one can know how many blocks there are in each dimension)?

src/raster/rasterband.rs
```diff
@@ -556,6 +562,41 @@ impl<'a> RasterBand<'a> {
...
+    /// * `block_index` - the block index
+    ///
+    /// # Notes
+    /// The Matrix shape is (rows, cols) and raster shape is (cols in x-axis, rows in y-axis).
+    pub fn write_block<T: Copy + GdalType>(
+        &mut self,
+        block_index: (usize, usize),
```
metasim, line 576: What happens if the block index is out of bounds? Do we get an error, or `panic` (or worse)?

src/raster/rasterband.rs
metasim: An example would be helpful here.

metasim (changes requested):
I'm torn about having core operations that are only supported with in `ndarray` enabled, mode without a technical reason.

* Pros: easier to submit PRs.
* Cons: forces users to use ndarry when in reality

I'd like to use these features as a `Buffer` user.

But I admittedly have a bit of an axe to grind here. I would have preferred ndarray integration to have happened as a copy-free `Buffer::to_ndarray` method (for example), rather than having dual I/O methods for `Buffer` and for `Array2`.

src/raster/rasterband.rs
lnicola: There's no way it can panic, of course :-). Added a note in the docs.

src/raster/rasterband.rs
lnicola: Added an example and a note to `block_size` and `actual_block_size`.

metasim (approved):

lnicola:
@metasim do you ever use the MD API, and would you expect it to require `ndarray`, or use some form of `Buffer3`, `BufferN` version?

metasim:
> do you ever use the MD API, and would you expect it to require ndarray, or use some form of Buffer3, BufferN version?

@lnicola I do not. That's a good point (I didn't realize MD required it). For that API it seems reasonable to require `ndarray`, and would be silly to implement a bespoke multidimensional array capability just to avoid a dependency.

I wouldn't be adverse to `ndarray` being required for everything. But because we advertise it as an "optional" dependency, I don't think not using it should preclude access to basic functionality (i.e. block I/O).

lnicola:
> I didn't realize MD required it

It doesn't, I should have checked before asking :pensive:.

metasim:
PS: I think `ndarray` is fantastic. My position is more about it's optionality, and treating `Buffer` as a first-class citizen if face of that optionality.

PR #492, opened 2023-12-15
Fixes for static pkg-config Windows build

PaulWagener:
---

Defers the `gdal_i.lib` missing message until after the pkg-config check and outputs pkg-config metadata in case of a static build.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


lnicola:
Bah, this script is so hard to follow.

Can you please add a changelog entry?

PaulWagener:
Done

lnicola:
I can't really verify this, but it seems reasonable.

PR #494, opened 2023-12-17
Lift `ndarray` compatibility into `Buffer<T>`

metasim:
---

- Added ability to convert between `Buffer<T>` and `ndarray::Array2<T>`. 
- Implemented `IntoIterator`, `Index` and `IndexMut` for `Buffer<T>`. 
- **Breaking**: `Buffer<T>::size` is now private and accessed via `Buffer<T>::shape().
- **Breaking**: `Buffer<T>::data` is now private and accessed via `Buffer<T>::data(). 
- **Breaking**: Removed `Rasterband::read_as_array`, changed signature of `Rasterband::read_block` to return a `Buffer<T>`. 
- **Breaking**: `Rasterband::write` and `Rasterband::write_block` now require a `&mut Buffer<T>` to handle possible case of drivers temporarily mutating input buffer.

cc: @mwielocha

### DIFF
--- a/src/raster/buffer.rs
+++ b/src/raster/buffer.rs
@@ -24,8 +24,10 @@ impl<T: GdalType> Buffer<T> {
         assert_eq!(
             shape.0 * shape.1,
             data.len(),
-            "size {:?} does not match length {}",
-            shape,
+            "shape {}*{}={} does not match length {}",
+            shape.0,
+            shape.1,
+            shape.0 * shape.1,
             data.len()
         );
         Buffer { shape, data }
--- a/src/raster/buffer.rs
+++ b/src/raster/buffer.rs
@@ -37,7 +37,7 @@ use ndarray::Array2;
 ///
 /// // Gather basic statistics:
 /// fn min_max(b: &Buffer<f64>) -> (f64, f64) {
-///     b.into_iter().fold((f64::MAX, f64::MIN), | a, v | (v.min(a.0), v.max(a.1)))
+///     b.into_iter().fold((f64::MAX, f64::MIN), |a, v| (v.min(a.0), v.max(a.1)))
 /// }
 /// let (min, max) = min_max(&buf);
 /// assert_eq!(min, -999999.0); // <--- data has a sentinel value
@@ -129,9 +129,7 @@ impl<T: GdalType> Buffer<T> {
     #[inline(never)]
     #[track_caller]
     fn panic_bad_index(shape: (usize, usize), coord: (usize, usize)) -> ! {
-        panic!(
-        "index out of bounds: buffer has shape `{shape:?}` but coordinate `{coord:?}` was requested",
-    );
+        panic!("index out of bounds: buffer has shape `{shape:?}` but coordinate `{coord:?}` was requested");
     }
 
     #[inline]
--- a/src/raster/buffer.rs
+++ b/src/raster/buffer.rs
@@ -125,19 +125,20 @@ impl<T: GdalType> Buffer<T> {
         )?)
     }
 
+    #[cold]
+    #[inline(never)]
+    #[track_caller]
+    fn panic_bad_index(shape: (usize, usize), coord: (usize, usize)) -> ! {
+        panic!(
+        "index out of bounds: buffer has shape `{shape:?}` but coordinate `{coord:?}` was requested",
+    );
+    }
+
     #[inline]
+    #[track_caller]
     fn vec_index_for(&self, coord: (usize, usize)) -> usize {
-        if coord.0 >= self.shape.0 {
-            panic!(
-                "index out of bounds: buffer has {} columns but row {} was requested",
-                self.shape.0, coord.0
-            );
-        }
-        if coord.1 >= self.shape.1 {
-            panic!(
-                "index out of bounds: buffer has {} rows but row {} was requested",
-                self.shape.1, coord.1
-            );
+        if coord.0 >= self.shape.0 || coord.1 >= self.shape.1 {
+            Self::panic_bad_index(self.shape, coord);
         }
         coord.0 * self.shape.0 + coord.1
     }
--- a/src/raster/buffer.rs
+++ b/src/raster/buffer.rs
@@ -1,0 +1,86 @@
+use crate::raster::GdalType;
+
+#[cfg(feature = "ndarray")]
+use ndarray::Array2;
+
+/// A 2-D array backed by it's `size` (cols, rows) and a row-major `Vec<T>` and it's dimensions.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Buffer<T> {
+    pub size: (usize, usize),
+    pub data: Vec<T>,
+}
+
+impl<T: GdalType> Buffer<T> {
+    /// Construct a new buffer from `size` (`(cols, rows)`) and `Vec<T>`.
+    ///
+    /// # Notes
+    /// Assumes `size.0 * size.1 == data.len()`.
+    pub fn new(size: (usize, usize), data: Vec<T>) -> Buffer<T> {
+        debug_assert!(
+            size.0 * size.1 == data.len(),
+            "size {:?} does not match length {}",
+            size,
+            data.len()
+        );
+        Buffer { size, data }
+    }
+
+    #[cfg(feature = "ndarray")]
+    /// Convert `self` into an [`ndarray::Array2`].
+    pub fn to_array(self) -> crate::errors::Result<Array2<T>> {
+        // Array2 shape is (rows, cols) and Buffer shape is (cols in x-axis, rows in y-axis)
+        Ok(Array2::from_shape_vec(
+            (self.size.1, self.size.0),
+            self.data,
+        )?)
+    }
+}
+
+pub type ByteBuffer = Buffer<u8>;
+
+#[cfg(feature = "ndarray")]
+impl<T: GdalType> TryFrom<Buffer<T>> for Array2<T> {
+    type Error = crate::errors::GdalError;
+
+    fn try_from(value: Buffer<T>) -> Result<Self, Self::Error> {
+        value.to_array()
+    }
+}
+
+#[cfg(feature = "ndarray")]
+impl<T: GdalType + Copy> From<Array2<T>> for Buffer<T> {
+    fn from(value: Array2<T>) -> Self {
+        // Array2 shape is (rows, cols) and Buffer shape is (cols in x-axis, rows in y-axis)
+        let shape = value.shape().to_vec();
+        let data: Vec<T> = if value.is_standard_layout() {
+            value.into_raw_vec()
+        } else {
+            value.iter().copied().collect()
+        };
+
+        Buffer::new((shape[1], shape[0]), data)
+    }
+}
+
+#[cfg(feature = "ndarray")]
+#[cfg(test)]
+mod tests {
+    use crate::raster::Buffer;
+    use ndarray::Array2;
+
+    #[test]
+    fn convert_to() {
+        let b = Buffer::new((5, 10), (0..5 * 10).collect());
+        let a = b.clone().to_array().unwrap();
+        let b2: Buffer<_> = a.into();
+        assert_eq!(b, b2);
+    }
+
+    #[test]
+    fn convert_from() {
+        let a = Array2::from_shape_fn((10, 5), |(y, x)| y as i32 * 10 + x as i32);
+        let b: Buffer<_> = a.clone().into();
+        let a2 = b.to_array().unwrap();
+        assert_eq!(a, a2);
+    }
+}
--- a/src/raster/buffer.rs
+++ b/src/raster/buffer.rs
@@ -1,4 +1,6 @@
 use crate::raster::GdalType;
+use std::ops::{Index, IndexMut};
+use std::slice::Iter;
 
 #[cfg(feature = "ndarray")]
 use ndarray::Array2;
@@ -6,24 +8,56 @@ use ndarray::Array2;
 /// A 2-D array backed by it's `size` (cols, rows) and a row-major `Vec<T>` and it's dimensions.
 #[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq)]
 pub struct Buffer<T> {
-    pub size: (usize, usize),
-    pub data: Vec<T>,
+    shape: (usize, usize),
+    data: Vec<T>,
 }
 
 impl<T: GdalType> Buffer<T> {
     /// Construct a new buffer from `size` (`(cols, rows)`) and `Vec<T>`.
     ///
-    /// # Panic
+    /// # Panics
     /// Will panic if `size.0 * size.1 != data.len()`.
-    pub fn new(size: (usize, usize), data: Vec<T>) -> Self {
+    pub fn new(shape: (usize, usize), data: Vec<T>) -> Self {
         assert_eq!(
-            size.0 * size.1,
+            shape.0 * shape.1,
             data.len(),
             "size {:?} does not match length {}",
-            size,
+            shape,
             data.len()
         );
-        Buffer { size, data }
+        Buffer { shape, data }
+    }
+
+    /// Destructures `self` into constituent parts.
+    pub fn into_shape_and_vec(self) -> ((usize, usize), Vec<T>) {
+        (self.shape, self.data)
+    }
+
+    /// Gets the 2-d shape of the buffer.
+    ///
+    /// Returns `(cols, rows)`
+    pub fn shape(&self) -> (usize, usize) {
+        self.shape
+    }
+
+    /// Get a slice over the buffer contents.
+    pub fn data(&self) -> &[T] {
+        self.data.as_slice()
+    }
+
+    /// Get a mutable slice over the buffer contents.
+    pub fn data_mut(&mut self) -> &mut [T] {
+        self.data.as_mut_slice()
+    }
+
+    /// Get the number of elements in the buffer
+    pub fn len(&self) -> usize {
+        self.data.len()
+    }
+
+    /// Determine if the buffer has no elements.
+    pub fn is_empty(&self) -> bool {
+        self.data.is_empty()
     }
 
     #[cfg(feature = "ndarray")]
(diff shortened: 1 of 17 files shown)


lnicola:
I'll have to think a bit more about this one, but I'm not sure I'm a fan. We should probably enable conversion from `Buffer` to `Array2` and back instead of having two separate read and write functions.

But more importantly, we should have a way to pass in a buffer instead of allocating a new one each time.

metasim:
@lnicola Let me restate the use case just to make sure it's clear what I'm trying to do here.

I have an application where I do not want to include/use the `ndarray` library, and I want to be able to read an image block by block. Before this PR I could not do that. Currently, block reading is only available if you are an `ndarray` user.

The renaming of the existing one is clearly not required, so I can figure out some other naming if that's objectionable.

metasim:
I just re-read your comment. If what you're arguing is that we should normalize the I/O around the `Buffer<T>` type (or, perhaps even better, `Vec<T>`!), then I'm even more of a fan of that, but wondered how controversial that would be.

lnicola:
Yes, I think we should use `Buffer` with easy conversions to `ndarray`. It's better than `Vec` because it has the shape information.

metasim:
I'll prototype something.

src/raster/rasterband.rs
```diff
@@ -636,29 +600,32 @@ impl<'a> RasterBand<'a> {
...
             gdal_sys::GDALWriteBlock(
                 self.c_rasterband,
                 block_index.0 as c_int,
                 block_index.1 as c_int,
-                data.as_mut_ptr() as GDALRasterBandH,
+                // This parameter is marked as `* mut c_void` because the C/C++ API for some reason
+                // doesn't mark it as `const void *`. From code inspection starting at the link
+                // below, it appears to be a read-only array.
+                // https://github.com/OSGeo/gdal/blob/b5d004fb9e3fb576b3ccf5f9740531b0bfa87ef4/gcore/gdalrasterband.cpp#L688
+                block.data.as_ptr() as *mut c_void,
```
metasim, line 628: @lnicola I have a pending question to the GDAL devs on this:

https://lists.osgeo.org/pipermail/gdal-dev/2023-December/058130.html

I would have expected a "write" method to only need a shared reference/pointer to the data.

src/raster/rasterband.rs
```diff
@@ -689,7 +656,7 @@ impl<'a> RasterBand<'a> {
                 window.1 as c_int,
                 window_size.0 as c_int,
                 window_size.1 as c_int,
-                buffer.data.as_ptr() as GDALRasterBandH,
+                buffer.data.as_ptr() as *mut c_void,
```
metasim, line 659: The `as GDARasterBandH` was semantically incorrect.

src/raster/rasterband.rs
```diff
@@ -1175,19 +1142,6 @@ impl<'a> MajorObject for RasterBand<'a> {
...
-}
-
-impl<T: GdalType> Buffer<T> {
-    pub fn new(size: (usize, usize), data: Vec<T>) -> Buffer<T> {
-        Buffer { size, data }
-    }
-}
-
-pub type ByteBuffer = Buffer<u8>;
-
```
metasim, line 1190: Moved to `buffer.rs`

src/raster/buffer.rs
```diff
@@ -0,0 +1,86 @@
...
+#[cfg(feature = "ndarray")]
+impl<T: GdalType + Copy> From<Array2<T>> for Buffer<T> {
+    fn from(value: Array2<T>) -> Self {
+        // Array2 shape is (rows, cols) and Buffer shape is (cols in x-axis, rows in y-axis)
+        let shape = value.shape().to_vec();
+        let data: Vec<T> = if value.is_standard_layout() {
+            value.into_raw_vec()
+        } else {
+            value.iter().copied().collect()
+        };
```
metasim, line 59: I'm attempting here to efficiently convert back to a `Vec<T>` while protecting against receiving a column-major Array.

The old implementation just assumed the storage ordering was always row-major: 


https://github.com/georust/gdal/pull/494/files#diff-432e492d6a25cce04b77014813aea6861769ff42ce0c362d3f7850fb9e21ab5cL655

That said, I'm not sure how to test this properly.

src/raster/rasterband.rs
lnicola: :scream_cat: I didn't even notice that.

src/raster/buffer.rs
lnicola: It might be easier to use [`as_standard_layout`](https://docs.rs/ndarray/latest/ndarray/struct.ArrayBase.html#method.as_standard_layout).

To test, I think we can use e.g. `Shape::from((2, 2)).set_f(true)` (from the `ShapeBuilder` trait).

src/raster/buffer.rs
```diff
@@ -0,0 +1,86 @@
...
+    fn try_from(value: Buffer<T>) -> Result<Self, Self::Error> {
+        value.to_array()
+    }
+}
+
+#[cfg(feature = "ndarray")]
+impl<T: GdalType + Copy> From<Array2<T>> for Buffer<T> {
+    fn from(value: Array2<T>) -> Self {
+        // Array2 shape is (rows, cols) and Buffer shape is (cols in x-axis, rows in y-axis)
+        let shape = value.shape().to_vec();
```
lnicola, line 54: No need to `to_vec()` here, `shape` appears to be a `&[usize]`.

src/raster/buffer.rs
```diff
@@ -0,0 +1,86 @@
+use crate::raster::GdalType;
+
+#[cfg(feature = "ndarray")]
+use ndarray::Array2;
+
+/// A 2-D array backed by it's `size` (cols, rows) and a row-major `Vec<T>` and it's dimensions.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Buffer<T> {
```
lnicola, line 8: I wonder if we should call it `Buffer2`.

src/raster/buffer.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub size: (usize, usize),
+    pub data: Vec<T>,
+}
+
+impl<T: GdalType> Buffer<T> {
+    /// Construct a new buffer from `size` (`(cols, rows)`) and `Vec<T>`.
+    ///
+    /// # Notes
+    /// Assumes `size.0 * size.1 == data.len()`.
+    pub fn new(size: (usize, usize), data: Vec<T>) -> Buffer<T> {
```
lnicola, line 18: Nit:

```suggestion
    pub fn new(size: (usize, usize), data: Vec<T>) -> Self {
```

src/raster/buffer.rs
```diff
@@ -0,0 +1,86 @@
...
+        debug_assert!(
```
lnicola, line 19:
```suggestion
        assert!(
```

I don't think anything good can come out of passing in the wrong shape.

src/raster/buffer.rs
```diff
@@ -0,0 +1,86 @@
+use crate::raster::GdalType;
+
+#[cfg(feature = "ndarray")]
+use ndarray::Array2;
+
+/// A 2-D array backed by it's `size` (cols, rows) and a row-major `Vec<T>` and it's dimensions.
+#[derive(Debug, Clone, PartialEq)]
```
lnicola, line 7: Should we also derive `Eq`, `ParialOrd` and `Ord`?

src/raster/buffer.rs
```diff
@@ -0,0 +1,86 @@
...
+pub struct Buffer<T> {
+    pub size: (usize, usize),
+    pub data: Vec<T>,
+}
+
+impl<T: GdalType> Buffer<T> {
+    /// Construct a new buffer from `size` (`(cols, rows)`) and `Vec<T>`.
+    ///
+    /// # Notes
+    /// Assumes `size.0 * size.1 == data.len()`.
```
lnicola, line 17:
```suggestion
```

src/raster/buffer.rs
metasim: I got an ownership error without that.

src/raster/buffer.rs
metasim: `Eq` and `Ord` get really messy with `Buffer<f32|f64>`.

src/raster/buffer.rs
lnicola: I mean, just in case we want to do something fancy for the MD API (which I didn't look at too closely).

src/raster/buffer.rs
lnicola: A `let (cols, rows) = (shape[1], shape[0]);` should fix that.

src/raster/buffer.rs
lnicola: Just a little. `#[derive]` [automatically adds](https://github.com/rust-lang/rust/issues/26925) a `T: Eq` bound, which works in our favour this time.

src/raster/buffer.rs
lnicola: Well, I don't know if it's really a good idea, just thought it might make sense.

src/raster/rasterband.rs
metasim: @lnicola Based on this response, do you think the above is OK? Seems like our only options are:

1. Make `block` `&mut`
2. Create a copy of `block` before passing the data along
3. Accept the ambiguity and hope no-one runs into problems :/ 

https://lists.osgeo.org/pipermail/gdal-dev/2023-December/058134.html

lnicola: `&mut` with a note in the comments seems perfectly reasonable.

src/raster/buffer.rs
metasim: This is what it looks like now, which I would guess forces a copy regardless of storage ordering, but I'm not sure. At least in the case for `value.into_raw_vec()`, you're given ownership of the underlying buffer:

```rust
        let shape = value.shape();
        let (rows, cols) = (shape[0], shape[1]);
        let data = value
            .as_standard_layout()
            .iter()
            .copied()
            .collect::<Vec<T>>();
        Buffer::new((cols, rows), data)
```
https://github.com/georust/gdal/pull/494/files#diff-92db5ec20d7f8881c279099dfd10ceb3e36c1cec359f7d463d4f5a35497837f1R55-R62

lnicola: Yeah, I think the old version was better. Maybe use `value.into_raw_vec()` as before, and `.to_vec()` instead of `collect`?

src/raster/buffer.rs
```diff
@@ -0,0 +1,86 @@
+use crate::raster::GdalType;
+
+#[cfg(feature = "ndarray")]
+use ndarray::Array2;
+
+/// A 2-D array backed by it's `size` (cols, rows) and a row-major `Vec<T>` and it's dimensions.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Buffer<T> {
+    pub size: (usize, usize),
```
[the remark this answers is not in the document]
metasim: > get the Vec back, e.g. `fn into_shape_and_vec(self)`.

Scala had an `unapply` method for this sort of thing, which I kinda miss in Rust.

src/raster/buffer.rs
metasim: `to_vec` only exists on `Array1`.

https://docs.rs/ndarray/latest/ndarray/struct.ArrayBase.html#methods-for-1-d-arrays

src/raster/buffer.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+impl<'a, T: GdalType> IntoIterator for &'a Buffer<T> {
+    type Item = &'a T;
+    type IntoIter = Iter<'a, T>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.data.iter()
+    }
+}
```
metasim, line 112: @lnicola These look sane to you?

metasim: If you want linear indexing, you can do `buf.data()[idx]`.

src/raster/buffer.rs
metasim: FWIW, still testing/working on this section; inspecting in debugger to follow what's happening.

metasim: The tests:

https://github.com/georust/gdal/pull/494/files#diff-92db5ec20d7f8881c279099dfd10ceb3e36c1cec359f7d463d4f5a35497837f1R174-R191


Confirmed in debugger that with `set_f(true)` the execution path goes through the `! is_standard_layout` path.

src/raster/buffer.rs
```diff
@@ -142,24 +148,48 @@ impl<T: GdalType + Copy> From<Array2<T>> for Buffer<T> {
...
+        let expected = Array2::from_shape_fn((10, 5), |(row, col)| row as i32 * 5 + col as i32);
+        assert_eq!(a, expected);
         let b2: Buffer<_> = a.into();
         assert_eq!(b, b2);
     }
 
     #[test]
     fn convert_from() {
-        let a = Array2::from_shape_fn((10, 5), |(y, x)| y as i32 * 10 + x as i32);
+        let a = Array2::from_shape_fn((10, 5), |(row, col)| row as i32 * 5 + col as i32);
```
lnicola, line 167: `shapes` uses `(y, x)`, do we pick one?

src/raster/buffer.rs
```diff
@@ -28,14 +31,17 @@ impl<T: GdalType> Buffer<T> {
         Buffer { shape, data }
     }
 
-    /// Destructures `self` into constituent parts.
+    /// De-structures `self` into constituent parts.
```
lnicola, line 36: I see where you're coming from, but I really hate this.

src/raster/rasterband.rs
```diff
@@ -467,8 +467,8 @@ impl<'a> RasterBand<'a> {
                 window_size.0 as c_int,
                 window_size.1 as c_int,
                 data.as_mut_ptr() as *mut c_void,
-                size.0 as c_int,
-                size.1 as c_int,
+                shape.0 as c_int,
```
lnicola, line 469: Let's use `try_into` or `try_from` here (unless you've already changed it in another commit). Same for `window` and `window_size`.

src/raster/rasterband.rs
```diff
@@ -662,9 +662,9 @@ impl<'a> RasterBand<'a> {
                 window.1 as c_int,
                 window_size.0 as c_int,
                 window_size.1 as c_int,
-                buffer.data.as_mut_ptr() as *mut c_void,
-                buffer.size.0 as c_int,
-                buffer.size.1 as c_int,
+                buffer.data_mut().as_mut_ptr() as *mut c_void,
+                shape.0 as c_int,
```
lnicola, line 669: Ditto

src/raster/buffer.rs
```diff
@@ -1,40 +1,114 @@
...
+}
+
+impl<T: GdalType> IndexMut<(usize, usize)> for Buffer<T> {
+    fn index_mut(&mut self, index: (usize, usize)) -> &mut Self::Output {
+        let idx = self.vec_index_for(index);
+        &mut self.data[idx]
+    }
+}
+
+impl<'a, T: GdalType> IntoIterator for &'a Buffer<T> {
```
lnicola, line 113: We want three implementations, for `Buffer<T>`, for `&Buffer<T>` and for `&mut Buffer<T>`.

src/raster/buffer.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub fn to_array(self) -> crate::errors::Result<Array2<T>> {
+        // Array2 shape is (rows, cols) and Buffer shape is (cols in x-axis, rows in y-axis)
+        Ok(Array2::from_shape_vec(
+            (self.shape.1, self.shape.0),
+            self.data,
+        )?)
+    }
+
+    #[inline]
+    pub(self) fn vec_index_for(&self, coord: (usize, usize)) -> usize {
```
lnicola, line 82: I don't think `pub(self)` does anything.

src/raster/buffer.rs
metasim: WDYM? The respelling of the word, or the method all together? 

If the former, I'll change it
If the latter, I thought we needed something to pull it apart. Other options?

src/raster/buffer.rs
metasim: I only changed the binding names to check my sanity. Want me to change it back?

src/raster/buffer.rs
```diff
@@ -0,0 +1,224 @@
...
+        value.to_array()
+    }
+}
+
+#[cfg(feature = "ndarray")]
+impl<T: GdalType + Copy> From<Array2<T>> for Buffer<T> {
+    fn from(value: Array2<T>) -> Self {
+        // Array2 shape is (rows, cols) and Buffer shape is (cols in x-axis, rows in y-axis)
+        let shape = value.shape();
+        let (rows, cols) = (shape[0], shape[1]);
```
lnicola, line 138: Nit:

```suggestion
        let (cols, rows) = (shape[1], shape[0]);
```

(don't ask :sweat_smile:)

src/raster/buffer.rs
lnicola: The spelling with a hyphen.

src/raster/rasterband.rs
lnicola: Oh, sorry, I thought you'd already merged `master`.

src/raster/buffer.rs
lnicola: Just thinking we should be consistent about them.

src/raster/buffer.rs
lnicola: For reference, it avoids a bounds check for `shape[0]` and a panic message in the binary:

image

src/raster/rasterband.rs
metasim: Actually I had, but hadn't addressed these cases yet. Will do.

src/raster/rasterband.rs
metasim: Maybe we should have made them `u16`!??‽? 😮

lnicola: Sorry, what do you mean?

src/raster/mdarray.rs
```diff
@@ -248,7 +248,7 @@ impl<'a> MDArray<'a> {
 
     #[cfg(feature = "ndarray")]
     #[cfg_attr(docsrs, doc(cfg(feature = "array")))]
-    /// Read a 'Array2<T>' from this band. T implements 'GdalType'.
+    /// Read a [`ArrayD<T>`] from this band. T implements [`GdalType`].
```
lnicola, line 251:
```suggestion
    /// Read an [`ArrayD<T>`] from this band. T implements [`GdalType`].
```

src/utils.rs
```diff
@@ -80,3 +80,12 @@ pub fn _path_to_c_string(path: &Path) -> Result<CString> {
     let path_str = path.to_string_lossy();
     CString::new(path_str.as_ref()).map_err(Into::into)
 }
+
+#[inline]
+pub(crate) fn coerce_size_tuple<T>(s: (T, T)) -> Result<(libc::c_int, libc::c_int)>
```
lnicola, line 85: 😵‍💫

src/raster/buffer.rs
```diff
@@ -0,0 +1,305 @@
...
+impl<'a, T: GdalType> IntoIterator for &'a mut Buffer<T> {
+    type Item = &'a mut T;
+    type IntoIter = IterMut<'a, T>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.data.iter_mut()
+    }
+}
+
+pub type ByteBuffer = Buffer<u8>;
```
lnicola, line 187: Why do we even have this? :confused:

src/raster/buffer.rs
```diff
@@ -0,0 +1,305 @@
...
+                self.shape.0, coord.0
+            );
+        }
+        if coord.1 >= self.shape.1 {
+            panic!(
+                "index out of bounds: buffer has {} rows but row {} was requested",
+                self.shape.1, coord.1
+            );
+        }
+        coord.0 * self.shape.0 + coord.1
```
lnicola, line 142: I think we should move the `panic!`s into another function with `#[cold] #[inline(never)]`. It's all right if we check the coordinates twice (both here and in that function).

lnicola: Actually, I think a single `panic!` that prints the index and shape should be enough. The user can figure it out.

src/raster/buffer.rs
metasim: no clue. Chesterton's Fence.

metasim: I almost removed it (only in one test) until I saw it was `pub`.

lnicola (approved):

PR #497, opened 2023-12-19
Changed a number of APIs using `isize` when `usize` is semantically more appropriate

### LINKED ISSUE #495: Change raster and block sizes to `usize`
While negative offsets have a purpose, negative sizes do not.

Create methods also use `isize` for band counts.

metasim:
---

Scope:

* `Driver::create.*`
* `Rasterband::overview`
* `Dataset::{layer|into_layer|layer_count}`

Closes #495

### DIFF
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -835,7 +835,9 @@ impl<'a> RasterBand<'a> {
         unsafe { Ok(gdal_sys::GDALGetOverviewCount(self.c_rasterband)) }
     }
 
-    pub fn overview(&self, overview_index: isize) -> Result<RasterBand<'a>> {
+    pub fn overview(&self, overview_index: usize) -> Result<RasterBand<'a>> {
+        let overview_index = i32::try_from(overview_index)?;
+
         unsafe {
             let c_band = self.c_rasterband;
             let overview = gdal_sys::GDALGetOverview(c_band, overview_index as libc::c_int);
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -836,11 +836,11 @@ impl<'a> RasterBand<'a> {
     }
 
     pub fn overview(&self, overview_index: usize) -> Result<RasterBand<'a>> {
-        let overview_index = i32::try_from(overview_index)?;
+        let overview_index = libc::c_int::try_from(overview_index)?;
 
         unsafe {
             let c_band = self.c_rasterband;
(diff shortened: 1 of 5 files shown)


src/vector/layer.rs
```diff
@@ -620,7 +619,8 @@ impl Dataset {
     ///
     /// Applies to vector datasets, and fetches by the given
     /// _0-based_ index.
-    pub fn into_layer(self, idx: isize) -> Result<OwnedLayer> {
+    pub fn into_layer(self, idx: usize) -> Result<OwnedLayer> {
+        let idx = i32::try_from(idx)?;
```
lnicola, line 623: I didn't check, but does `c_int::try_from` work? It might be more portable, and we cast to `c_int` afterwards anyway.

Ditto everywhere else.

lnicola (approved):
LGTM and I think it's a great idea, but if we go down this route, should we also change `rasterband` and the histogram APIs to return an error?

src/vector/layer.rs
metasim: `libc::c_int::try_from` does work! Good idea.

metasim:
@lnicola 
> should we also change rasterband and the histogram APIs to return an error?

My apologies, but I'm not following your question... 🤔

lnicola:
We panic there if the index is too large.

lnicola:
I meant https://github.com/georust/gdal/blob/master/src/raster/rasterband.rs#L39 and https://github.com/georust/gdal/blob/master/src/raster/rasterband.rs#L1010 etc.

metasim:
ooooh, yeah... let me fix that. Given the option, I think passing the error up is better.

metasim:
@lnicola Do I still have your approval after this additional change?

https://github.com/georust/gdal/pull/497/commits/75b4c122fbd2e6c487c6647f0bc8fe65ba08f321

src/raster/rasterband.rs
```diff
@@ -36,10 +36,10 @@ impl Dataset {
     /// # Panics
     /// Panics if the band index is greater than `c_int::MAX`.
     pub fn rasterband(&self, band_index: usize) -> Result<RasterBand> {
-        assert!(band_index <= c_int::MAX as usize);
+        let band_index = libc::c_int::try_from(band_index)?;
```
lnicola, line 39: The panic note needs to be dropped.

metasim: 🙏 Good catch.

PR #501, opened 2023-12-20
Implemented `Feature::set_field_null`

### LINKED ISSUE #427: FR: add wrapper for OGR_F_SetFieldNull
All of the "get" field functions on Feature return options, so I can know if the value is already null. But I don't see anything in the rust API for setting a field to None that shouldn't have a value, and `OGR_F_UnsetField` is only referenced in the pre-built bindings on a sourcecode search here.

My work-around right now is to use `feature.set_field_double(<field_name>,f64::NAN)` but I've only had to do this for double fields so far, so I don't know if this will work for integer, or if it will work in other drivers than geopackage.

metasim:
---

Closes #427 
Credit to @nms-scribe

### DIFF
--- a/src/vector/feature.rs
+++ b/src/vector/feature.rs
@@ -943,14 +943,12 @@ mod tests {
     fn test_field_set_null() {
         let ds = Dataset::open(fixture("roads.geojson")).unwrap();
 
-        for mut layer in ds.layers() {
-            for feature in layer.features() {
-                for (field_name, field_value) in feature.fields() {
-                    if let Some(_value) = field_value {
-                        feature.set_field_null(&field_name).unwrap();
-                        assert!(feature.field(&field_name).unwrap().is_none());
-                    }
-                }
+        let mut layer = ds.layers().next().expect("layer");
+        let feature = layer.features().next().expect("feature");
+        for (field_name, field_value) in feature.fields() {
+            if let Some(_value) = field_value {
+                feature.set_field_null(&field_name).unwrap();
+                assert!(feature.field(&field_name).unwrap().is_none());
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/vector/feature.rs
```diff
@@ -911,9 +926,32 @@ pub fn field_type_to_name(ty: OGRFieldType::Type) -> String {
...
+        assert_eq!(field_type_to_name(OGRFieldType::OFTReal), "Real");
+        // We don't care what it returns when passed an invalid value, just that it doesn't crash.
+        field_type_to_name(4372521);
+    }
+
+    #[test]
+    fn test_field_set_null() {
+        let ds = Dataset::open(fixture("roads.geojson")).unwrap();
+
+        for mut layer in ds.layers() {
```
lnicola, line 946: I think we can limit this to a single layer and feature.

lnicola:
I was too afraid to check in #427, but still have no idea. Does anyone know the difference between `OGR_F_SetFieldNull` and `OGR_F_UnsetField`?

Does the latter actually remove the field value in formats like GeoJSON?

metasim:
[Docs](https://gdal.org/api/vector_c_api.html): 

* `OGR_F_SetFieldNull`: Clear a field, marking it as null.
* `OGR_F_UnsetField`: Clear a field, marking it as unset.

@lnicola Yeah, I don't understand the semantics either.

Maybe we just mark this as "draft" and come back to it later?

lnicola:
No, I would have merged it, but unsetting all fields of all features in all layers seems a bit excessive. But it's only 21 features, so feel free to merge :slightly_smiling_face:.

metasim:
I simplified the test.

Issue #499, opened 2023-12-20
Mutators in `Feature` don't require mutable receiver


metasim:
`gdal::vector::Feature` has several methods which mutate state, but don't require `&mut self`:

* `set_field_string`
* `set_field_string_list`
* `set_field_double`
* `set_field_double_list`
* `set_field_integer`
* `set_field_integer_list`
* `set_field_integer64`
* `set_field_integer64_list`
* `set_field_datetime`
* `set_field`

Incidentally, `set_geometry` _does_ require `&mut self`.

PR #503, opened 2023-12-22
Add Feature::unset_field

lnicola:
---

### DIFF
--- a/src/vector/feature.rs
+++ b/src/vector/feature.rs
@@ -648,9 +648,25 @@ impl<'a> Feature<'a> {
         }
     }
 
+    /// Clear a field, marking it as null.
+    ///
+    /// See: [`OGRFeature::SetFieldNull`][SetFieldNull]
+    ///
+    /// [SetFieldNull]: https://gdal.org/api/ogrfeature_cpp.html#_CPPv4N10OGRFeature12SetFieldNullEi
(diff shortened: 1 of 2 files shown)


src/vector/feature.rs
```diff
@@ -648,12 +648,28 @@ impl<'a> Feature<'a> {
...
         unsafe { gdal_sys::OGR_F_SetFieldNull(self.c_feature(), idx) };
         Ok(())
     }
 
+    /// Clear a field, marking it as unset.
+    ///
+    /// See: [`OGRFeature::UnsetField`][UnsetField]
+    ///
+    /// [UnsetField]: https://gdal.org/api/ogrfeature_cpp.html#_CPPv4N10OGRFeature10UnsetFieldEi
+    pub fn unset_field(&self, field_name: &str) -> Result<()> {
```
rouault, line 667: I guess that should be (same for set_field_null())
```suggestion
    pub fn unset_field(&mut self, field_name: &str) -> Result<()> {
```

lnicola: It should, but all of these currently take `&self` for some reason and @metasim ran into some issues when trying to change them to `&mut self`.

lnicola: That's https://github.com/georust/gdal/issues/499. I think it should work, but didn't try it yet.

rouault (approved):

PR #508, opened 2023-12-28
Expansion of Warp API bindings

metasim:
---

- [x] Extract replicated methods in `*Options` types into macros.
- [ ] Figure out why `warp::tests::test_create_reproject` fails on GDAL < 3.4 due to statistics seemingly not respecting no-data values.


CHANGES.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- **Breaking**: Moved `raster::processing::dem` to `raster::dem`
+- **Breaking**: Moved `raster::reproject` to `raster::warp::reproject_image` and added more extensive options
+- Added `raster::warp::create_and_reproject_image`.
+
```
lnicola, line 10: Missing PR number.

src/raster/warp/create_and_reproject.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// Returns `None` if unset.
+    pub fn max_error(&self) -> Option<f64> {
+        self.max_error
+    }
+
+    /// Specify the source projection.
+    ///
+    /// If unset, the source projection is read from the source dataset.
+    /// If set, the source projection os overridden.
+    pub fn with_src_projection(&mut self, srs: &SpatialRef) -> &mut Self {
```
lnicola: - we should take `srs` by value instead of doing `srs.clone()`
 - the comment talks about "unset", but there's no `Option` around it
 - I think we use `projection` for strings and `spatial_ref` or `spatial_reference` for `SpatialRef`s.

src/raster/warp/create_and_reproject.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    /// Fetch the specified source projection, if any.
+    pub fn src_projection(&self) -> Option<&SpatialRef> {
+        self.src_srs.as_ref()
+    }
+
+    /// Specify the source no-data value.
+    ///
+    /// Overrides any no-data value specified in the source dataset.
+    pub fn with_src_nodata(&mut self, nodata_value: f64) -> &mut Self {
```
lnicola: Missing getter?

src/raster/warp/create_and_reproject.rs
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /// Specify the destination no-data value.
+    ///
+    /// # Notes
+    ///
+    /// While this setting this value ensures the given no-data value is used when no-data needs
+    /// to be written to the output, due to [this issue](https://lists.osgeo.org/pipermail/gdal-dev/2023-November/057887.html)
+    /// the resulting dataset will not have `NO_DATA` set in the metadata when used with `GDALCreateAndReprojectImage`.
+    pub fn with_dst_nodata(&mut self, nodata_value: f64) -> &mut Self {
```
lnicola: Ditto

src/raster/warp/create_and_reproject.rs
```diff
@@ -0,0 +1,250 @@
...
+        self.output_format.clone()
+    }
+
+    /// Fetch an immutable reference to the general Warp options.
+    pub fn warp_options(&self) -> &GdalWarpOptions {
+        &self.warp_options
+    }
+
+    /// Fetch a mutable reference to the general Warp options.
+    pub fn warp_options_mut(&mut self) -> &mut GdalWarpOptions {
```
lnicola: I think a plain setter would be useful here.

src/raster/warp/create_and_reproject.rs
```diff
@@ -0,0 +1,250 @@
...
+        // See: https://github.com/OSGeo/gdal/blob/7b6c3fe71d61699abe66ea372bcd110701e38ff3/alg/gdalwarper.cpp#L235
+        gdal_sys::GDALCreateAndReprojectImage(
+            src.c_dataset(),
+            src_wkt_ptr,
+            dest.as_ptr(),
+            dst_wkt.as_ptr(),
+            driver.c_driver(),
+            ptr::null_mut(),
+            warp_options.resampling_alg().to_gdal(),
+            warp_options.memory_limit() as f64,
```
lnicola: I wonder why this is `double` :smile:.

src/raster/warp/create_and_reproject.rs
```diff
@@ -0,0 +1,250 @@
...
+            warp_options.as_ptr_mut(),
+        )
+    };
+
+    if rv != CPLErr::CE_None {
+        return Err(_last_cpl_err(rv));
+    }
+
+    // See https://lists.osgeo.org/pipermail/gdal-dev/2023-November/057887.html for
+    // why this is required. To get around it We should rewrite this function to use the lower
```
lnicola: "we", and you probably meant "the lower-level `GDALWarp` API"

src/raster/warp/resample.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// Min (selects minimum of all non-NODATA contributing pixels)
+    Min = GDALResampleAlg::GRA_Min,
+    /// Med (selects median of all non-NODATA contributing pixels)
+    Med = GDALResampleAlg::GRA_Med,
+    /// Q1 (selects first quartile of all non-NODATA contributing pixels)
+    Q1 = GDALResampleAlg::GRA_Q1,
+    /// Q3 (selects third quartile of all non-NODATA contributing pixels)
+    Q3 = GDALResampleAlg::GRA_Q3,
+    /// Sum (weighed sum of all non-NODATA contributing pixels). Added in GDAL 3.1
+    Sum = GDALResampleAlg::GRA_Sum,
```
lnicola, line 36: Needs a `cfg`

src/raster/warp/resample.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// RMS (weighted root mean square (quadratic mean) of all non-NODATA contributing pixels)
+    RMS = GDALResampleAlg::GRA_RMS,
```
lnicola, line 39: Ditto, added in 3.3

src/raster/warp/resample.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+impl WarpResampleAlg {
+    pub fn to_gdal(self) -> GDALResampleAlg::Type {
+        self as GDALResampleAlg::Type
+    }
+    pub fn from_gdal(alg: GDALResampleAlg::Type) -> Result<Self> {
```
lnicola, line 46: Nit: let's add a newline.

src/raster/warp/warp_options.rs
```diff
@@ -0,0 +1,300 @@
...
+    pub fn resampling_alg(&self) -> WarpResampleAlg {
+        // `unwrap` below is ok because `with_resampling_alg` is the only way it got set,
+        // aside from the GDAL default, which is `GRA_NearestNeighbour`.
+        WarpResampleAlg::from_gdal(unsafe { (*self.as_ptr()).eResampleAlg }).unwrap_or_default()
+    }
+
+    /// Set the datatype used during processing.
+    ///
+    /// If unset, the algorithm picks the datatype.
+    pub fn with_working_datatype(&mut self, dt: GdalDataType) -> &mut Self {
```
lnicola, line 55: Nit: do we use `datatype` or `data_type`?

src/raster/warp/warp_options.rs
```diff
@@ -0,0 +1,300 @@
...
+        self
+    }
+
+    /// Fetch working datatype, if specified.
+    pub fn working_datatype(&self) -> Option<GdalDataType> {
+        let c_dt = unsafe { (*self.as_ptr()).eWorkingDataType };
+        let dt: GdalDataType = c_dt.try_into().ok()?;
+
+        // Default is `Unknown`, so we consider that "unspecified".
+        if dt == GdalDataType::Unknown {
```
lnicola, line 66: This might be a little confusing, should we drop the `Option`?

src/raster/warp/warp_options.rs
```diff
@@ -0,0 +1,300 @@
...
+    ///
+    /// `0` indicates all bands.
+    pub fn band_count(&mut self) -> usize {
+        let cnt = unsafe { (*self.as_ptr()).nBandCount };
+        cnt as usize
+    }
+
+    /// Sets the source Dataset no-data value. Internal use only.
+    ///
+    /// Specifying a no-data value for GDAL Warp requires it be specified for every band.
```
lnicola, line 113: This doesn't read very well.

What happens if `self.band_count()` is 0?

src/raster/warp/warp_options.rs
```diff
@@ -0,0 +1,300 @@
...
+        }
+
+        // The GDALWarpOptions destructor frees this. See:
+        // https://github.com/OSGeo/gdal/blob/a9635785a2db8f575328326f2b1833e743ec8828/alg/gdalwarper.cpp#L1295
+        unsafe { GDALWarpInitDstNoDataReal(self.as_ptr_mut(), no_data_value) };
+
+        // This ensures the destination cells are initialized with no-data
+        // See: https://gdal.org/api/gdalwarp_cpp.html#_CPPv4N15GDALWarpOptions16papszWarpOptionsE
+        self.extra_options_mut()
+            .set_name_value("INIT_DEST", "NO_DATA")?;
```
lnicola, line 150: This might be too opinionated for the users?

src/raster/warp/warp_options.rs
```diff
@@ -0,0 +1,300 @@
...
+    }
+}
+
+impl Default for GdalWarpOptions {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Display for GdalWarpOptions {
```
lnicola, line 219: Let's not implement this.

src/xml.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn drop(&mut self) {
+        unsafe { CPLDestroyXMLNode(self.0.as_mut()) };
+    }
+}
+
+impl FromStr for GdalXmlNode {
+    type Err = crate::errors::GdalError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let s2 = CString::new(s)?;
```
lnicola, line 50: Nit: shadowing (`s`) is probably more idiomatic.

src/xml.rs
```diff
@@ -0,0 +1,89 @@
...
+        let c_xml = unsafe { CPLParseXMLString(s2.as_ptr()) };
+        if c_xml.is_null() {
+            Err(_last_cpl_err(CPLErr::CE_Failure))
+        } else {
+            Ok(unsafe { GdalXmlNode::from_ptr(c_xml) })
+        }
+    }
+}
+
+impl Display for GdalXmlNode {
```
lnicola, line 60: Should this also implement `ToString`?

lnicola:
Looks pretty good, but I only checked the first two commits.

lnicola:
I do wonder, though, if we shouldn't go straight for `GDALWarp`. It's not that much harder to use, and I think that mailing list reply suggested that it's the most powerful API.

metasim:
> I do wonder, though, if we shouldn't go straight for `GDALWarp`. It's not that much harder to use, and I think that mailing list reply suggested that it's the most powerful API.

Good point. I'll research. At first glance it looked like it required a bunch more work, but maybe I jumped to conclusions there.

src/raster/warp/create_and_reproject.rs
metasim: 🤷

lnicola:
It's a lot of work to write the bindings for, it's similar to gdaldem.

src/raster/warp/create_and_reproject.rs
metasim: Fixed in subsequent commit.

src/raster/warp/create_and_reproject.rs
metasim: Also addressed in later commit.

src/raster/warp/warp_options.rs
metasim: We're not consistent by any stretch:

```
❯ rg --count-matches --ignore-case datatype --glob '*.rs' src
src/driver.rs:12
src/errors.rs:2
src/raster/tests.rs:6
src/raster/types.rs:177
src/raster/mod.rs:4
src/raster/rasterize.rs:1
src/raster/rasterband.rs:9
src/raster/mdarray.rs:76
```

```
❯ rg --count-matches --ignore-case 'data[-_]type' --glob '*.rs' src
src/raster/mdarray.rs:25
src/errors.rs:2
src/driver.rs:2
src/raster/types.rs:3
```

src/xml.rs
metasim: I've tended to conflate `Display` and `ToString` (since `Display` is `ToString`). Any advice on how to decide?

src/raster/warp/resample.rs
metasim: Shame on `rustfmt` for not catching that! :neckbeard:

src/raster/warp/warp_options.rs
metasim: I have a bias toward the one word "datatype" form. 

Since Wiktionary is our authoritative source for justifying our own spelling biases, I think I'm justified! 🤣 

https://en.wiktionary.org/wiki/datatype

src/raster/warp/warp_options.rs
metasim: If we did, would we change this

```rust
let dt: GdalDataType = c_dt.try_into().ok()?;
```

to 

```rust
let dt: GdalDataType = c_dt.try_into().unwrap_or(GdalDataType::Unknown);
```
?

src/xml.rs
metasim: Interesting... Docs say this:

image

sevberg:
Recently I've been experimenting with using gdal-sys to make calls to GDALWarp, which I've finally gotten running to my liking (besides the pesky MEM format..). But of course, only afterwards did I stumble across this PR

Anyway, getting GDALWarp implemented into the gdal crate would be a big win, IMO. So, if it helps, I would be happy to push this topic forward a bit with my learnings from the past few days? Please let me know :)

sevberg:
Oh good, looks like my [MEM issue](https://gis.stackexchange.com/questions/484666/calling-gdalwarp-with-a-mem-formatted-raster-is-ineffective) is resolved, as well. Maybe something to take into account here?

To summarize...

The issue: 
The behavior of GDALWarp seems to differ slightly depending on the format of the source/destination Datasets. For something like GTIFF -> GTIFF, things behave intuitively and the code you've written [here](https://github.com/metasim/gdal/blob/feature/warp/src/raster/warp/mod.rs#L166-L219) should work well (besides of course the `todo!`) without any options specified. However if either side of that equation is MEM-formatted, then the GDALWarp utility does not take the provided Dataset's SRS into account, and ends up warping into some other geospatial domain; which will almost certainly "miss" the extent of the destination dataset. However GDALWarp does not complain. It will return successfully, and the destination dataset will remain unchanged.

The solution:
The solution suggested to me on my SO question was to explicitly set the source/destination SRS in the GDALWarpAppOptions, which I can confirm works. 

My take:
TBH I'm not super happy about this solution, since it then requires a (what should be) unnecessary step of dumping the SRS into a string, and then passing them in again via the GDALWarpAppOptions. At least in my code, I'll just simply add a routine to check the input format and automatically add the SRS options when needed. However I'm not sure if doing the same here is in the spirit this library, since it changes the fundamental behavior of the wrapped codebase (unintuitive as it may be 🙈). I'm also not sure if this issue affects other formats besides MEM. So ultimately I'm happy to leave that decision up to you.

In any case, I just wanted to bring this to your attention.

metasim:
@sevberg I'm happy for you to take over this branch. I've had to change jobs to something outside of the geospatial field, and sadly no longer have time to focus on this (awesome) project.  Let me know if you have any questions.

metasim:
Abandoned.

PR #510, opened 2023-12-29
Add `DriverManager::get_drivers_for_filename` for the ability to auto detect compatible `Driver`s for writing data

### LINKED ISSUE #506: [Feature Request] Guess the Driver Type based on the file extension
Currently, you can open a dataset with just `Dataset::open("filename.ext")` without giving the driver, but we cannot create a file without driver. If we can, please let me know.

My attempt to do `Dataset::open` on a new file has failed, even when using `Dataset::open_ex` like below didn't work:
```
Dataset::open_ex(
                filename,
                DatasetOptions {
                    open_flags: gdal::GdalOpenFlags::GDAL_OF_UPDATE
                        .union(gdal::GdalOpenFlags::GDAL_OF_VECTOR),
                    ..Default::default()
                },
            )
```
I want to be able to output to any gis file format. GPKG, shp, json, and so on.

So far I only know how to make a new file like this:
```
let driver = DriverManager::get_driver_by_name(driver)?;
let mut out_data = driver.create_vector_only(filename)?;
```
Since we need to pass filename as well as the driver, it feels redundant, or extra work for user, where they might make mistakes.

So is there a way to add `DriverManager::get_driver_by_extension`, that can guess the driver based on the file exntension? 

Considering the `Dataset::open` doesn't need driver I thought maybe there is already a function, but it seems to call C functions and take some valid drivers list, so I wasn't able to find how to replicate that for writing new files.

Atreyagaurav:
---

Fixes #506

I have the tests at the end of `src/driver.rs`, let me know if I should remove it. 

I've only tested for getting the drivers here, but I tried using that function to see if the code I wrote that wrote GPKG files can write SHP now it works. Though some driver specific functions seem to make it not work if we're not careful (like `out_data.start_transaction()?` works for GPKG but not SHP).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Added `DriverManager::get_drivers_for_filename` for the ability to auto detect compatible `Driver`s for writing data.
+
 - Added `Feature::unset_field`
   - <https://github.com/georust/gdal/pull/503>
 
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -11,7 +11,7 @@ use crate::metadata::Metadata;
 use crate::raster::{GdalDataType, GdalType, RasterCreationOption};
 use crate::utils::{_last_cpl_err, _last_null_pointer_err, _path_to_c_string, _string};
 
-use crate::errors::*;
+use crate::{errors::*, GdalOpenFlags};
 
 static START: Once = Once::new();
 
@@ -410,6 +410,73 @@ impl DriverManager {
         Ok(Driver { c_driver })
     }
 
+    /// Get the Driver based on the file extension
+    ///
+    /// Searches for the available extensions in the registered
+    /// drivers and returns the first match
+    // it's not a good implementation but should work for now
+    pub fn get_drivers_for_filename(filename: &str, options: &GdalOpenFlags) -> Vec<Driver> {
+        let ext = {
+            let filename = filename.to_ascii_lowercase();
+            if let Some(mut ext) = Path::new(&filename).extension().and_then(|s| s.to_str()) {
+                if ext == "zip" {
+                    // doing .ends_with on the &Path filename doesn't work
+                    if filename.ends_with(".shp.zip") {
+                        ext = "shp.zip";
+                    } else if filename.ends_with(".gpkg.zip") {
+                        ext = "gpkg.zip";
+                    }
+                }
+                ext.to_string()
+            } else {
+                "".to_string()
+            }
+        };
+
+        let mut drivers: Vec<Driver> = Vec::new();
+        for i in 0..DriverManager::count() {
+            let d = DriverManager::get_driver(i).expect("Index for this loop should be valid");
+            let mut supports = false;
+            if (d.metadata_item("DCAP_CREATE", "").is_some()
+                || d.metadata_item("DCAP_CREATECOPY", "").is_some())
+                && ((options.contains(GdalOpenFlags::GDAL_OF_VECTOR)
+                    && d.metadata_item("DCAP_VECTOR", "").is_some())
+                    || (options.contains(GdalOpenFlags::GDAL_OF_RASTER)
+                        && d.metadata_item("DCAP_RASTER", "").is_some()))
+            {
+                supports = true;
+            } else if options.contains(GdalOpenFlags::GDAL_OF_VECTOR)
+                && d.metadata_item("DCAP_VECTOR_TRANSLATE_FROM", "").is_some()
+            {
+                supports = true;
+            }
+            if !supports {
+                continue;
+            }
+
+            if let Some(e) = &d.metadata_item("DMD_EXTENSION", "") {
+                if *e == ext {
+                    drivers.push(d);
+                    continue;
+                }
+            }
+            if let Some(e) = d.metadata_item("DMD_EXTENSIONS", "") {
+                if e.split(" ").collect::<Vec<&str>>().contains(&ext.as_str()) {
+                    drivers.push(d);
+                    continue;
+                }
+            }
+
+            if let Some(pre) = d.metadata_item("DMD_CONNECTION_PREFIX", "") {
+                if filename.starts_with(&pre) {
+                    drivers.push(d);
+                }
+            }
+        }
+
+        return drivers;
+    }
+
     /// Register a driver for use.
     ///
     /// Wraps [`GDALRegisterDriver()`](https://gdal.org/api/raster_c_api.html#_CPPv418GDALRegisterDriver11GDALDriverH)
@@ -465,5 +532,29 @@ mod tests {
 
         assert!(DriverManager::count() > 0);
         assert!(DriverManager::get_driver(0).is_ok());
+    }
+
+    #[test]
+    fn test_driver_extension() {
+        // convert the driver into short_name for testing purposes
+        let drivers = |filename, isvec| {
+            DriverManager::get_drivers_for_filename(
+                filename,
+                if isvec {
+                    &GdalOpenFlags::GDAL_OF_VECTOR
+                } else {
+                    &GdalOpenFlags::GDAL_OF_RASTER
+                },
+            )
+            .iter()
+            .map(|d| d.short_name())
+            .collect::<Vec<String>>()
+        };
+
+        assert_eq!(drivers("test.gpkg", true), vec!["GPKG"]);
+        assert_eq!(drivers("test.gpkg.zip", true), vec!["GPKG"]);
+        assert_eq!(drivers("test.tiff", false), vec!["GTiff", "COG"]);
+        assert_eq!(drivers("test.nc", false), vec!["netCDF"]);
+        assert_eq!(drivers("ES:test", true), vec!["Elasticsearch"]);
     }
 }
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -410,27 +410,50 @@ impl DriverManager {
         Ok(Driver { c_driver })
     }
 
-    /// Get the Driver based on the file extension
+    /// Get the Driver based on the file extension from filename
     ///
     /// Searches for the available extensions in the registered
-    /// drivers and returns the first match
-    // it's not a good implementation but should work for now
+    /// drivers and returns the matches, the `options:&GdalOpenFlags`
+    /// is only used for determining vector or raster type.
+    ///
+    /// See also: [`get_driver_by_name`](Self::get_driver_by_name)
+    ///
+    /// # Example
+    ///
+    /// ```rust, no_run
+    /// use gdal::{DriverManager,GdalOpenFlags};
+    /// # fn main() -> gdal::errors::Result<()> {
+    /// let compatible_drivers =
+    ///     DriverManager::get_drivers_for_filename("test.gpkg", &GdalOpenFlags::GDAL_OF_VECTOR)
+    ///         .iter()
+    ///         .map(|d| d.short_name())
+    ///         .collect::<Vec<String>>();
+    /// println!("{:?}", compatible_drivers);
+    /// # Ok(())
+    /// # }
+    /// ```
+    /// ```text
+    /// ["GPKG"]
+    /// ```
     pub fn get_drivers_for_filename(filename: &str, options: &GdalOpenFlags) -> Vec<Driver> {
         let ext = {
             let filename = filename.to_ascii_lowercase();
-            if let Some(mut ext) = Path::new(&filename).extension().and_then(|s| s.to_str()) {
-                if ext == "zip" {
-                    // doing .ends_with on the &Path filename doesn't work
-                    if filename.ends_with(".shp.zip") {
-                        ext = "shp.zip";
-                    } else if filename.ends_with(".gpkg.zip") {
-                        ext = "gpkg.zip";
+            let e = match filename.rsplit_once(".") {
+                Some(("", _)) => "", // hidden file no ext
+                Some((f, "zip")) => {
+                    // zip files could be zipped shp or gpkg
+                    if f.ends_with(".shp") {
+                        "shp.zip"
+                    } else if f.ends_with(".gpkg") {
+                        "gpkg.zip"
+                    } else {
+                        "zip"
                     }
                 }
-                ext.to_string()
-            } else {
-                "".to_string()
-            }
+                Some((_, e)) => e, // normal file with ext
+                None => "",
+            };
+            e.to_string()
         };
 
         let mut drivers: Vec<Driver> = Vec::new();
@@ -550,7 +573,6 @@ mod tests {
             .map(|d| d.short_name())
             .collect::<Vec<String>>()
         };
-
         assert_eq!(drivers("test.gpkg", true), vec!["GPKG"]);
         assert_eq!(drivers("test.gpkg.zip", true), vec!["GPKG"]);
         assert_eq!(drivers("test.tiff", false), vec!["GTiff", "COG"]);


Atreyagaurav:
Tested locally for: 
```
  cargo fmt --all --check
  cargo clippy --all-targets -- -D warnings
  cargo clippy --all-targets --all-features -- -D warnings
  cargo build
  cargo test
  cargo build --all-features
  cargo test --all-features -- --nocapture
```

Note: The original code suggested in #506 had the tests for writing capabilities of the driver, and I've kept that as, for reading a file, `Dataset::open` can already read any dataset without us having to determine a driver. I've added that in docstring.

Atreyagaurav:
The tests pass locally but are failing in CI, So I thought it could be because we need to install something for GPKG driver to be present in the ubuntu distribution.

So I've added conditional in the test to only test extension if the driver we're looking for is already there. It should work.

Atreyagaurav:
Found the problem: The previous versions don't support `.gpkg.zip` and only `.gpkg` files.

Relevant link: https://gdal.org/drivers/vector/gpkg.html#compressed-files

```
$ docker run -it --entrypoint bash ghcr.io/osgeo/gdal:ubuntu-full-3.8.1
root@65b7fb97e696:/# python3
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from osgeo import ogr
>>> ogr.GetDriverByName("GPKG").GetMetadataItem('DMD_EXTENSIONS')
'gpkg gpkg.zip'
>>> ogr.GetDriverByName("GPKG").GetMetadataItem('DMD_EXTENSION')
>>> ogr.GetDriverByName("GPKG").GetMetadataItem('DCAP_CREATE')
'YES'
>>> ogr.GetDriverByName("GPKG").GetMetadataItem('DCAP_VECTOR')
'YES'
>>> 
```
But 
```
$ docker run -it --entrypoint bash ghcr.io/osgeo/gdal:ubuntu-full-3.5.3
root@6691affad24c:/# python
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from osgeo import ogr
>>> ogr.GetDriverByName("GPKG").GetMetadataItem('DMD_EXTENSIONS')
'gpkg'
>>> ogr.GetDriverByName("GPKG").GetMetadataItem('DMD_EXTENSION')
'gpkg'
```

So I just made a condition on the `.gpkg.zip` assert in the test for newer versions only. Hopefully it works now

Atreyagaurav:
I checked that all upto gdal 3.1.3 has `shp.zip` support: 
https://gdal.org/drivers/vector/shapefile.html#compressed-files

This link shows that ubuntu-lts has gdal 3.0 https://askubuntu.com/questions/1267844/installing-libgdal-dev-on-ubuntu-20-04

I've made that changes in the test as well.

Atreyagaurav:
Ok, should I remove the ES test? I don't understand it, and I just added it for complete tests. I can't find if something changed in version 3.1 for that. Or should I add the condition of >= 3010000 for it as well.

Atreyagaurav:
```
$ docker run -it --entrypoint bash ubuntu:20.04
root@1015807d8b92:/# apt install libpq-dev gdal-bin libgdal-dev
root@1015807d8b92:/# python3
Python 3.8.10 (default, Nov 22 2023, 10:22:35) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from osgeo import ogr
>>> import osgeo
>>> osgeo.__version__
'3.0.4'
>>> ogr.GetDriverByName("Elasticsearch").GetMetadataItem('DMD_CONNECTION_PREFIX')
'ES:'
>>> ogr.GetDriverByName("Elasticsearch").GetMetadataItem('DCAP_CREATE')
'YES'
>>> ogr.GetDriverByName("Elasticsearch").GetMetadataItem('DCAP_VECTOR')
'YES'
>>>
```

I don't know why it's failing there. the things we're checking seems fine. Should we just put `gdal_version >= 3010000` for the test?

lol, the name was different `ogr.GetDriverByName("Elasticsearch").name` came up as `ElasticSearch` in 3.0.4, while for others it's `Elasticsearch`. Should be fixed now. Hopefully no more problems. 

The logic for the function is correct, tests are troublesome.

src/driver.rs
```diff
@@ -410,6 +410,90 @@ impl DriverManager {
...
+    ///         .map(|d| d.short_name())
+    ///         .collect::<Vec<String>>();
+    /// println!("{:?}", compatible_drivers);
+    /// # Ok(())
+    /// # }
+    /// ```
+    /// ```text
+    /// ["GPKG"]
+    /// ```
+    pub fn get_drivers_for_filename(filename: &str, is_vector: bool) -> Vec<Driver> {
```
metasim, line 440: Why is `is_vector` necessary? Is it possible to have multiple drivers that read the same  file extension but differentiate on vector vs. raster data? If so, is this something we need to impose on every user (IOW, could we let the caller deal with that if it's a rare edge case?)

I guess what I'm wondering here is if we should only be evaluating against `DMD_EXTENSION` and `DMD_EXTENSIONS`, and defer capability filtering to a different mechanism (e.g. extension method on `Driver`). The method name implies we're only looking at matching drivers against filename (extension).

src/driver.rs
```diff
@@ -410,6 +410,90 @@ impl DriverManager {
...
+
+        let mut drivers: Vec<Driver> = Vec::new();
+        for i in 0..DriverManager::count() {
+            let d = DriverManager::get_driver(i).expect("Index for this loop should be valid");
+            let can_create = d.metadata_item("DCAP_CREATE", "").is_some()
+                || d.metadata_item("DCAP_CREATECOPY", "").is_some();
+            let check_vector = is_vector && d.metadata_item("DCAP_VECTOR", "").is_some();
+            let check_raster = !is_vector && d.metadata_item("DCAP_RASTER", "").is_some();
+            let check_vector_translate =
+                is_vector && d.metadata_item("DCAP_VECTOR_TRANSLATE_FROM", "").is_some();
```
metasim, line 469: See comment above... i.e. should we consider selection and filtering two different operations?

src/driver.rs
```diff
@@ -410,6 +410,90 @@ impl DriverManager {
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    /// ```text
+    /// ["GPKG"]
+    /// ```
+    pub fn get_drivers_for_filename(filename: &str, is_vector: bool) -> Vec<Driver> {
+        let ext = {
+            let filename = filename.to_ascii_lowercase();
+            let e = match filename.rsplit_once('.') {
```
metasim, line 443: If we want to ensure true logical parity with GDAL C behavior, we'd use `CPLGetExtension`. 50/50 on whether that's good idea or not.

src/driver.rs
metasim, line 469: See this description of what GDAL does when `Dataset::open` is called. 
https://github.com/OSGeo/gdal/blob/master/gcore/gdaldataset.cpp#L3630-L3645

From my 5 minute reading it looks like some of the logic is delegated to `GDALIdentifyDriverEx`:

https://github.com/OSGeo/gdal/blob/master/gcore/gdaldriver.cpp#L2649C25-L2649C45

My concern here is not wanting to diverge from the accepted GDAL logic.

src/driver.rs
```diff
@@ -466,4 +552,46 @@ mod tests {
...
+        }
+        if DriverManager::get_driver_by_name("Elasticsearch").is_ok() {
+            // Elasticsearch short name was ElasticSearch in older versions
+            if gdal_version >= 3010000 {
+                assert!(drivers("ES:test", true).contains("Elasticsearch"));
+            } else {
+                assert!(drivers("ES:test", true).contains("ElasticSearch"));
+            }
+        }
+    }
```
metasim, line 596: Can you add tests for files like `this.is.my.test.tiff` and `my.other.test.shp.zip`?

src/driver.rs
```diff
@@ -410,6 +410,90 @@ impl DriverManager {
         Ok(Driver { c_driver })
     }
 
+    /// Get the Driver based on the file extension from filename
+    ///
+    /// Searches for the available extensions in the registered
+    /// drivers and returns the matches. The determined driver is
+    /// checked for writing capabilities as
+    /// [`Dataset::open`](Dataset::open) can already open datasets
+    /// with just path.
```
metasim, line 419: Not sure I understand why we limit returned drivers to ones with "write" capabilities...  maybe read-only use cases might also want a list of drivers to pick from (e.g. when reading JP2k, there are several options and you might want to inspect the metadata to decide on which to select).

metasim (changes requested):

metasim:
> I have the tests at the end of src/driver.rs, let me know if I should remove it.

That's the appropriate place :)

src/driver.rs
lnicola: `GDALIdentifyDriverEx` is used for opening, but that's usually not necessary.

The functions I linked to in https://github.com/georust/gdal/issues/506#issuecomment-1868587093 is used by the GDAL binaries to pick an output format. If we offer this function, we should match the latter, and use a better name (`get_driver(s)_for_writing` or something like that).

src/driver.rs
lnicola: I don't know. In principle, the user might have all sorts of extra drivers installed. Matching the GDAL logic seemed like a safe approach.

We could ask Even, I guess.

Atreyagaurav: The logic I went by is this: there can be multiple drivers for the same extension, and some could support raster some vector and some both. And this function is (related with other comments) made to know the driver in advance while creating a file as if we have a file already, then we can call `Dataset::open` without driver information. But we cannot create and save data without deciding the driver first. And if we're writing a data, we'd want the list of drivers that supports that type of data, hence that bool. 

It was also in the function Inicola sent me as an example, so it's my understanding of that, not my innovation.

src/driver.rs
Atreyagaurav: I did consider using the rust's "extension()" method on the `Path` as well, but considering that it can only do the last "." onwards, which requires us to do the `endswith` on the filename for `.shp.zip` and `.gpkg.shp`, I thought just using them directly might be better. And the online protocol addresses with prefix (like `Elasticsearch` in the test) may not be taken as a valid path, so I just did the string thing.

src/driver.rs
Atreyagaurav: So, initially I was thinking that, but then I realized that when we already have a dataset, we can just call `Dataset::open` and then look at the driver. Having a dataset you can open means you have a lot in your arsenal to detect the driver, like the metadata you mentioned.

The use case we're thinking is to when we need to write a dataset, currently the only way I could find to write to file is to create a file from a driver, which requires you to know the driver in advance. Of course, you can make the user pass the driver name along with the filename and call it a day (that's what I was doing on one of my program), but all users are not likely to know about the driver names. And if we can more or less guess based on the file extension, then we make it convenient and remove any user error (could write a Tiff file in .shp extension file and it'll write it). And since GIS programs let you write to any file, making the program only dependent on the filename makes it easier to integrate with other programs as well.

I think renaming the function to mention it's for writing should help with it.

Atreyagaurav:
> That's the appropriate place :)

Thank you. That was more of a comment due to worry that the tests might fail on the CI or on other people's computer because some drivers are not available. But that should be fixed now. Thought the tests did give a lot of trouble for other reasons.

lnicola:
Nit: maybe we should use something more common like PG instead of ElasticSearch?

Atreyagaurav:
> Nit: maybe we should use something more common like PG instead of ElasticSearch?

Done.

I also renamed the funtion. Idk if PG will also have some quirks on the tests like ElasticSearch though. Hopefully not

src/driver.rs
Atreyagaurav: done

src/driver.rs
```diff
@@ -410,6 +410,90 @@ impl DriverManager {
...
+                    }
+                }
+                Some((_, e)) => e, // normal file with ext
+                None => "",
+            };
+            e.to_string()
+        };
+
+        let mut drivers: Vec<Driver> = Vec::new();
+        for i in 0..DriverManager::count() {
```
lnicola, line 462: Note for another PR: we should have an iterator over the available drivers.

Atreyagaurav: You mean to say we don't, but we should make one? Then yeah definitely. Can work on that.

If you mean we already have one, then I can edit it to use that.

lnicola: Don't think we do. In the meanwhile you could rewrite it with iterators, so it doesn't need to return a vector, but it's not super important.

src/driver.rs
metasim: > The functions I linked to in https://github.com/georust/gdal/issues/506#issuecomment-1868587093 is used by the GDAL binaries to pick an output format.

Ahhh... I _thought_ there was something already in the code, but didn't realize it was C++ only :/

src/driver.rs
```diff
@@ -410,6 +410,90 @@ impl DriverManager {
...
+    ///         .map(|d| d.short_name())
+    ///         .collect::<Vec<String>>();
+    /// println!("{:?}", compatible_drivers);
+    /// # Ok(())
+    /// # }
+    /// ```
+    /// ```text
+    /// ["GPKG"]
+    /// ```
+    pub fn guess_drivers_for_write(filename: &str, is_vector: bool) -> Vec<Driver> {
```
metasim, line 440: Sorry to bikeshed here, but should we consider:

1. Also providing `guess_drivers_for_read`
2. Have the signature `guess_drivers_for_write<P: AsRef<Path>>(filename: P, ...)` to be more coherent with other methods accepting files. (`&str` is `AsRef<Path>`).

lnicola: Not sure about `guess_drivers_for_read`. There's already an API that works for existing files, and do we really need to find a driver that doesn't exist?

metasim:
LGTM

src/driver.rs
Atreyagaurav: Yeah, it is for binary apps. Not the part of the `gdal` library. Otherwise, I was planning on just using the function from `gdal-sys`.

src/driver.rs
Atreyagaurav: I did consider using `AsRef<Path>`, but I thought the connections like `PG:https://xxxxxx` might not work. And we need to convert it back to string to use the "endswith" or "split" for double extensions. Hence, I went with string. I haven't tested if those connections work as path or not though.

Atreyagaurav:
Last forced commit is the function name change in the `CHANGES.md`, so it should pass all the tests like before.

I can work on the iterator for drivers in a separate pull request.

lnicola:
You don't have to, but that will fix it the PR approval thing and might be less controversial.

Atreyagaurav:
No, I mean for the whole Drivers. Not related to this PR. Just making `DriverIterator` that can be used to iterate though all drivers. This one is fine to return Vec for now. I was just offering I can do that task. 

That way, we can probably return filter of that iterator for this function later. Since we need to make a datatype to implement iterator, I don't want to do that for this function alone. Since there is no `yield` like in python.

metasim:
@Atreyagaurav With #512 merged, does that affect how you want to finish up this implementation?

Atreyagaurav:
We can use the iterator here, but I'm worried that pulling the master branch might make it troublesome when we'll be squashing the commits here...

Should I still pull the master to make the changes?

Atreyagaurav:
Well , it has conflicts, so might have to pull it anyway. It should be done.

Atreyagaurav:
Now it returns an Iterator. Users should be able to do `next()` for the first match without having to go through the whole list. Thank you for the suggestion.

metasim:
> We can use the iterator here, but I'm worried that pulling the master branch might make it troublesome when we'll be squashing the commits here...

This is standard operating procedure. Do that, or rebasing what you have on top of `master`. No matter, any conflicts have to be resolved eventually, so pay the tax now or later! 😆

src/driver.rs
```diff
@@ -410,6 +410,87 @@ impl DriverManager {
...
+                (can_create && (check_vector || check_raster)) || (check_vector_translate)
+            })
+            .filter(move |d| {
+                if let Some(e) = &d.metadata_item("DMD_EXTENSION", "") {
+                    if *e == ext {
+                        return true;
+                    }
+                }
+                if let Some(e) = d.metadata_item("DMD_EXTENSIONS", "") {
+                    if e.split(' ').collect::<Vec<&str>>().contains(&ext.as_str()) {
```
lnicola, line 480: No need to `collect` here, there's `.find().is_some()` and `.any()`.

src/driver.rs
```diff
@@ -496,6 +577,47 @@ mod tests {
...
+                .map(|d| d.short_name())
+                .collect::<HashSet<String>>()
+        };
+        let gdal_version: i64 = crate::version::version_info("VERSION_NUM").parse().unwrap();
+        if DriverManager::get_driver_by_name("ESRI Shapefile").is_ok() {
+            assert!(drivers("test.shp", true).contains("ESRI Shapefile"));
+            assert!(drivers("my.test.shp", true).contains("ESRI Shapefile"));
+            // `shp.zip` only supported from gdal version 3.1
+            // https://gdal.org/drivers/vector/shapefile.html#compressed-files
+            if gdal_version >= 3010000 {
```
lnicola, line 594: Let's do a `if cfg!(all(major_ge_3, minor_ge_1))`, it's a little more easy to search for.

src/driver.rs
```diff
@@ -496,6 +577,47 @@ mod tests {
...
+                assert!(drivers("test.shp.zip", true).contains("ESRI Shapefile"));
+                assert!(drivers("my.test.shp.zip", true).contains("ESRI Shapefile"));
+            }
+        }
+        if DriverManager::get_driver_by_name("GPKG").is_ok() {
+            assert!(drivers("test.gpkg", true).contains("GPKG"));
+            assert!(drivers("my.test.gpkg", true).contains("GPKG"));
+            // `gpkg.zip` only supported from gdal version 3.7
+            // https://gdal.org/drivers/vector/gpkg.html#compressed-files
+            if gdal_version >= 3070000 {
```
lnicola, line 604: Same

src/driver.rs
```diff
@@ -410,6 +410,87 @@ impl DriverManager {
...
+
+        DriverManager::all()
+            .filter(move |d| {
+                let can_create = d.metadata_item("DCAP_CREATE", "").is_some()
+                    || d.metadata_item("DCAP_CREATECOPY", "").is_some();
+                let check_vector = is_vector && d.metadata_item("DCAP_VECTOR", "").is_some();
+                let check_raster = !is_vector && d.metadata_item("DCAP_RASTER", "").is_some();
+                let check_vector_translate =
+                    is_vector && d.metadata_item("DCAP_VECTOR_TRANSLATE_FROM", "").is_some();
+                (can_create && (check_vector || check_raster)) || (check_vector_translate)
```
lnicola, line 471: Nit:

```suggestion
                (can_create && (check_vector || check_raster)) || check_vector_translate
```

src/driver.rs
lnicola: It would be nice to have this work for `Path`s. I think you can use `extension` and then `file_stem` and again `extension` to check for this without that much ceremony.

src/driver.rs
```diff
@@ -410,6 +410,87 @@ impl DriverManager {
...
+                    } else if f.ends_with(".gpkg") {
+                        "gpkg.zip"
+                    } else {
+                        "zip"
+                    }
+                }
+                Some((_, e)) => e, // normal file with ext
+                None => "",
+            };
+            e.to_string()
```
lnicola, line 460: Not needed, but will change anyway if we go with `Path`.

lnicola:
r? @rouault (it's all right if you don't have the time though)

It feels like this functionality should be available in GDAL, do you remember why it's not? And are there pairs of built-in drivers that support the same extension, but one only for vector and one for raster datasets? External ones can do whatever they want, of course.

(that is, `D1` can write vector `.foo` and `D2` can write raster `.foo`)

@Atreyagaurav I think there are a couple more checks in the GDAL code (for GMT/NetCDF and GTiff/COG), can you add those? We might have to drop the iterator idea, though.

Atreyagaurav:
I'll make the changes in the review in a few hours.

The things about GMT/NetCDF, GTiff/COG, should we really do that? Since it returns the iterator, the user of the lib can choose which one they want to go with. That's why I didn't add it. This is a function for library use, which the one we were looking at are for gdal binaries, so we can't make assumptions about preference since we're not making the binaries right?

Also, when I tested it in mine, GTiff came before COG, and GMT didn't come up with (`.nc`) and only NetCDF came up.

src/driver.rs
```diff
@@ -410,6 +410,87 @@ impl DriverManager {
...
+    /// println!("{:?}", compatible_drivers);
+    /// # Ok(())
+    /// # }
+    /// ```
+    /// ```text
+    /// ["GPKG"]
+    /// ```
+    pub fn guess_drivers_for_write(
+        filename: &str,
+        is_vector: bool,
```
lnicola, line 441: An enum would probably be nicer here.

Atreyagaurav: Enum for Vector/Raster? I was thinking if we do that, we should either use the OpenOptions that has those already, but also has other options like read/write flag we can look at. But it has a lot of variants and adding checks for all of them would be complicated.

And if we're making new enum, should we just do vector and raster, or should we add, raster multiband and other raster formats, as well as vector specific subtypes as well. All of those seem complicated. Hence simple bool as we're only looking at vector or raster.

lnicola:
> Since it returns the iterator, the user of the lib can choose which one they want to go with.

The user might not be aware of the subtleties involving GMT and NetCDF. I'm certainly not. Maybe we want two functions, one that returns all drivers and one that returns a single one, with the additional heirustics.

> Also, when I tested it in mine, GTiff came before COG, and GMT didn't come up with (.nc) and only NetCDF came up.

GMT seems to be vector only. And the order might depend on the GDAL version.

src/driver.rs
lnicola: > we should either use the OpenOptions

I'm not sure. `GDAL_OF_XXX_BLOCK_ACCESS` we certainly don't want, the others might be nice to have. A separate enum is certainly more clear and I'm starting to like the idea.

> And if we're making new enum, should we just do vector and raster, or should we add, raster multiband and other raster formats, as well as vector specific subtypes as well. All of those seem complicated. Hence simple bool as we're only looking at vector or raster.

It's fine, we can start with what the function supports. The problem with a bool is that you see `guess_driver("x.nc", true)` and you have no idea what `true` means unless you check. `guess_driver("x.nc", FormatOptions::Raster)` and you immediately know what's up.

rouault:
> It feels like this functionality should be available in GDAL, do you see any reason not to expose it?

Queued for 3.9 per https://github.com/OSGeo/gdal/pull/9041

Atreyagaurav:
Well, now do we wait for the `gdal` release to call that function instead?

Atreyagaurav:
Yup, the order of the drivers seems to be dependent on versions. That could actually mean, we can just have them ordered based on the preference we want in newer versions and just drop it. The people using old versions could be made aware of that. But I've added that in the new function to get a single driver for now.

Also, if we want to wait to use the `gdal-sys` function from the new version, we can wait. Otherwise, we can merge this, and then later replace the implementation by calling that function internally.

metasim:
> Also, if we want to wait to use the gdal-sys function from the new version, we can wait. Otherwise, we can merge this, and then later replace the implementation by calling that function internally.

This is my vote, assuming the semantics are/will be the same. Would be nice to expose the capability for users using older versions of 3.x, but then add a `cfg` block for the new version of gdal, upon release.

src/driver.rs
Atreyagaurav: Did `.any()`, `.find(),is_some()` got flagged by the clippy lol.

src/driver.rs
Atreyagaurav: Done.

src/driver.rs
Atreyagaurav: Done.

src/driver.rs
Atreyagaurav: Fixed along with others.

src/driver.rs
Atreyagaurav: I don't get it. I did some changes here. But idk.

I'm really reluctant to use path, as we could have URLs for PG and other backends. And rust path library is tested for local file paths only (at least from what I can see) URLs could complicate things in the future.

src/driver.rs
Atreyagaurav: I made a simple enum with raster and vector variants for now. We can add more things as we add other stuffs.

lnicola:
Yeah, I think we can add it now, with a note that it's implemented differently before 3.9.

src/driver.rs
lnicola: `Path` works fine with those, especially as they're just valid file names on Linux. `Path::new` is infallible.

I'd like this because otherwise it's pretty annoying to convert a `Path` or `PathBuf` to a string.

Atreyagaurav: That's exactly I don't want to do it lol. 

Like for that: `startswith()` we need to do for "PG:" and similar paths, we'd need to convert the path into string right? And iirc it can't convert it into direct string and has to go through Debug trait and ends up with quoted string or something.

I can try, but unless we can use the `Path` args without having to turn it back to string, it feels like we'll just end up asking for a Path and then internally convert it into string anyway.

lnicola: > And iirc it can't convert it into direct string and has to go through Debug trait and ends up with quoted string or something.

[to_string_lossy](https://doc.rust-lang.org/std/path/struct.Path.html#method.to_string_lossy) will convert it without allocating in the happy case, and without too much ceremony. It's also infallible.

Anyway, I'm not that worried about allocations, just about having a nicer API -- if you have a string, you can get a `Path`, if you have a `Path` it's harder to get a string.

Atreyagaurav: > if you have a Path it's harder to get a string.

That conflicts with the first sentence :D

Users can just call `[to_string_lossy](https://doc.rust-lang.org/std/path/struct.Path.html#method.to_string_lossy)`. If it's easy for us, it's easy for them. 

Anyway, I'm trying to do it with `AsRef<Path>`. I'll make a commit if I am successful.

lnicola: > If it's easy for us, it's easy for them.

It's easy for us in the sense that it's infallible, it requires no unsafe code, and it's almost always efficient. It's harder for them because it's ugly and maybe they won't know about `to_string_lossy` so they might try to debug-format them, as you wanted to do.

---

There's also something to be said about being _lossy_, but I guess we only support UTF-8 file names, while GDAL might, in principle, support arbitrary byte sequences). But we hide it, and we could support non-UTF-8 file names today -- at least on Linux -- without changing our API.

On the other hand, if we make the user call `to_string_lossy` on a non-UTF-8 path, they'll get a different name at the output. It's fine when we do it (and also when the user does it) because it won't affect the driver selection in practice. But as you can see, it's a lot of relatively subtle stuff to think about.

Atreyagaurav: Haha, I'm currently having that same problem. it seems I can't use Cow<'_, str> on pattern. And might have to do, Variants of `Cow::`. At this point doing if statement seems easier than match. It's really a lot of complications when we have to turn it back into string. :(

lnicola: Okay, let's leave it like this, I'll try to clean it up tomorrow.

Atreyagaurav: Because of the abovementioned problem with `Cow`, and not being able to use match properly and so on. I converted it into String after fighting with borrow checking for a while trying to use str.

lnicola:
I pushed a couple of changes, but still want to do another pass before merging, sorry.

Atreyagaurav:
Wouldn't the name `get_drivers_for_path` be better than `name`? As `name` is used as driver name `driver_by_name`, so it might be ambiguous. So `filename` or `path` might be better.

lnicola:
Yeah, probably. I had `dataset_name`, but it seemed a little excessive. `path` is confusing for PG.

Atreyagaurav:
It's still a path in case of PG and others, just not a local one. Should be fine if people check the doc strings, if they don't I thought `name` might make people try the driver name instead.

Atreyagaurav:
This is finalized, right? No more modifications?

CHANGES.md
```diff
@@ -10,6 +10,8 @@
 
   - <https://github.com/georust/gdal/pull/504>
 
+- Added `DriverManager::guess_driver_for_write` and `DriverManager::guess_drivers_for_write` for the ability to auto detect compatible `Driver`(s) for writing data.
+
```
metasim, line 14: Don't forget to add the PR link here.

src/driver.rs
```diff
@@ -461,6 +592,11 @@ impl DriverManager {
     }
 }
 
+pub enum DriverProperties {
```
metasim, line 595: Consider the name `DriverType`, as "Properties" usually implies a collection of settings.

CHANGES.md
Atreyagaurav: Thank you. Done.

src/driver.rs
Atreyagaurav: Thank you, I was initially thinking that properties might be weird. It's renamed now.

metasim:
Looks good to me. @lnicola OK if I merge?

lnicola:
Yes, but I think it still needs a rebase.

I'm a little worried about that enum, since we should probably add a variat (bit) to warn when there are multiple matching drivers (GDAL has a second parameter, but we don't need to match that). That would make the name a little unfortunate, and something like DriverPreference or GetOutputDriverFlags would be nicer.

But let's merge this now, as I'm not going to have much time for it in the near future and I don't want to keep delaying it forever.

metasim:
Is a "Squash and merge" sufficient?

lnicola:
Feel free to try it, I'm seeing:

> This branch cannot be rebased due to conflicts

metasim:
@Atreyagaurav thanks for your hard, patient work on this. We all appreciate it! :)

Atreyagaurav:
Thank you everyone for reviewing and the guidance. And for making gdal available in rust. I'll try to contribute more if I find some simple issues I can fix.

PR #513, opened 2024-01-05
Add `DriverIterator` to iterate though the registered drivers

### LINKED ISSUE #512: [Feature] Add `DriverIterator` to iterate through the available drivers
Following up from the discussions in #510.

I was thinking about it and came across a potential problem. We can add and remove the drivers during runtime, so if we keep track of the index and increase that to get the `Iterator::next()`, then we'll run into:
- Assuming count is same:
  + panic on `Driver::get(i)` if driver(s) is/are removed
  + miss the last driver(s) if driver(s) is/are added
- Assuming we check the count each time:
  + skip driver(s) if driver(s) is/are removed before current count
 
So, to make sure this doesn't happen, we can take a Vec of the Drivers and make an iterator out of it, but that defeats the purpose of the Iterators. But maybe that'll work, as we don't have to call the `get_driver(i)` ourselves in a loop.

Atreyagaurav:
---
Fixes #512 

I've done the simplest implementation for the first commit:

```rust
pub struct DriverIterator {
    current: usize,
}

impl DriverIterator {
    pub fn new() -> Self {
        DriverIterator { current: 0 }
    }
}

impl Iterator for DriverIterator {
    type Item = Driver;

    fn next(&mut self) -> Option<Self::Item> {
        match DriverManager::get_driver(self.current) {
            Ok(d) => {
                self.current += 1;
                Some(d)
            }
            Err(_) => None,
        }
    }
}
```

Which seems too simple, so I'm thinking of adding the ability to filter by metadata.

Which would be something like this:

```rust

pub struct DriverIterator {
    current: usize,
    filters: Vec<String>,
}

impl DriverIterator {
    pub fn all() -> Self {
        DriverIterator {
            current: 0,
            filters: vec![],
        }
    }

    pub fn vectors() -> Self {
        DriverIterator {
            current: 0,
            filters: vec!["DCAP_VECTOR".into()],
        }
    }

    pub fn rasters() -> Self {
        DriverIterator {
            current: 0,
            filters: vec!["DCAP_RASTER".into()],
        }
    }

    pub fn with_metadata(metadata: Vec<String>) -> Self {
        DriverIterator {
            current: 0,
            filters: metadata,
        }
    }
}

impl Iterator for DriverIterator {
    type Item = Driver;

    fn next(&mut self) -> Option<Self::Item> {
        while let Ok(d) = DriverManager::get_driver(self.current) {
            self.current += 1;
            if self
                .filters
                .iter()
                .all(|f| d.metadata_item(f, "").is_some())
            {
                return Some(d);
            }
        }
        None
    }
}
```

If someone needs to filter by more complex conditions, they can use the filter directly in the iterator, but providing some options for vectors and rasters seems like a reasonable thing to do.

What do you think?

### DIFF
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -451,10 +451,45 @@ impl DriverManager {
             gdal_sys::GDALDestroyDriverManager();
         }
     }
+
+    /// Get an `Iterator` for all the drivers.
+    ///
+    /// Warning: Adding or removing drivers while comsuming the
+    /// Iterator can have unexpected results.
+    pub fn all() -> DriverIterator {
+        DriverIterator::new()
+    }
+}
+
+/// Iterator for the registered `Driver`s in `DriverManager`
+pub struct DriverIterator {
+    current: usize,
+}
+
+impl DriverIterator {
+    pub fn new() -> Self {
+        DriverIterator { current: 0 }
+    }
+}
+
+impl Iterator for DriverIterator {
+    type Item = Driver;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match DriverManager::get_driver(self.current) {
+            Ok(d) => {
+                self.current += 1;
+                Some(d)
+            }
+            Err(_) => None,
+        }
+    }
 }
 
 #[cfg(test)]
 mod tests {
+    use std::collections::HashSet;
+
     use super::*;
 
     #[test]
@@ -465,5 +500,16 @@ mod tests {
 
         assert!(DriverManager::count() > 0);
         assert!(DriverManager::get_driver(0).is_ok());
+    }
+
+    #[test]
+    fn test_driver_iterator() {
+        assert_eq!(DriverManager::count(), DriverIterator::new().count());
+        assert_eq!(DriverManager::count(), DriverManager::all().count());
+
+        let drivers: HashSet<String> = DriverManager::all().map(|d| d.short_name()).collect();
+        for i in 0..DriverManager::count() {
+            assert!(drivers.contains(&DriverManager::get_driver(i).unwrap().short_name()))
+        }
     }
 }
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -452,24 +452,18 @@ impl DriverManager {
         }
     }
 
-    /// Get an `Iterator` for all the drivers.
+    /// Get an `Iterator` over for all the loaded drivers.
     ///
-    /// Warning: Adding or removing drivers while comsuming the
-    /// Iterator can have unexpected results.
+    /// Warning: Adding or removing drivers while consuming the
+    /// iterator is safe, but can produce less useful results.
     pub fn all() -> DriverIterator {
-        DriverIterator::new()
-    }
-}
-
-/// Iterator for the registered `Driver`s in `DriverManager`
-pub struct DriverIterator {
-    current: usize,
-}
-
-impl DriverIterator {
-    pub fn new() -> Self {
         DriverIterator { current: 0 }
     }
+}
+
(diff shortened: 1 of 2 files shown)


src/driver.rs
```diff
@@ -451,10 +451,45 @@ impl DriverManager {
             gdal_sys::GDALDestroyDriverManager();
         }
     }
+
+    /// Get an `Iterator` for all the drivers.
```
lnicola, line 455:
```suggestion
    /// Get an over all the loaded drivers.
```

src/driver.rs
```diff
@@ -451,10 +451,45 @@ impl DriverManager {
...
+    ///
+    /// Warning: Adding or removing drivers while comsuming the
```
lnicola, line 457:
```suggestion
    /// Warning: Adding or removing drivers while consuming the
```

src/driver.rs
```diff
@@ -451,10 +451,45 @@ impl DriverManager {
...
+    /// Iterator can have unexpected results.
```
lnicola, line 458:
```suggestion
    /// iterator is safe, but can produce less useful results.
```

src/driver.rs
```diff
@@ -451,10 +451,45 @@ impl DriverManager {
...
+    pub fn all() -> DriverIterator {
+        DriverIterator::new()
+    }
+}
+
+/// Iterator for the registered `Driver`s in `DriverManager`
```
lnicola, line 464:
```suggestion
/// Iterator for the registered [`Driver`]s in [`DriverManager`]
```

src/driver.rs
```diff
@@ -451,10 +451,45 @@ impl DriverManager {
...
+pub struct DriverIterator {
+    current: usize,
+}
+
+impl DriverIterator {
```
lnicola, line 469: Let's drop this method and `impl` block.

src/driver.rs
```diff
@@ -451,10 +451,45 @@ impl DriverManager {
             gdal_sys::GDALDestroyDriverManager();
         }
     }
+
+    /// Get an `Iterator` for all the drivers.
+    ///
+    /// Warning: Adding or removing drivers while comsuming the
+    /// Iterator can have unexpected results.
+    pub fn all() -> DriverIterator {
+        DriverIterator::new()
```
lnicola, line 460:
```suggestion
        DriverIterator { current: 0 }
```

src/driver.rs
```diff
@@ -466,4 +501,15 @@ mod tests {
         assert!(DriverManager::count() > 0);
         assert!(DriverManager::get_driver(0).is_ok());
     }
+
+    #[test]
+    fn test_driver_iterator() {
+        assert_eq!(DriverManager::count(), DriverIterator::new().count());
+        assert_eq!(DriverManager::count(), DriverManager::all().count());
+
+        let drivers: HashSet<String> = DriverManager::all().map(|d| d.short_name()).collect();
```
lnicola, line 510: This feels a little overkill, I'd just keep the `count` test.

lnicola:
I'm personally happy enough with the current version.

src/driver.rs
Atreyagaurav: I was thinking the same in the beginning, but repeating the same driver `DriverManager::count()` times would pass the test. Although that's a very unlikely mistake, this feels like a complete test.

Atreyagaurav:
In that case, it's fine. Someone can always use a filter themselves.

I just thought having named methods might make it easier for the library users instead of having to navigate the metadata for specific driver capabilities.

lnicola:
If we add those, they can use `filter` and return `impl Iterator`.

lnicola (approved):

Issue #484, opened 2023-12-02
`gdal` 0.17 tracker


lnicola:
We should probably put out another release, given the recently-added support for GDAL 3.8 and other improvements.

My wishlist, after skimming the open issues, but by no means definitive:

 - [ ] #483
 - [x] #495
 - [x] #477 - let's settle for Ubuntu LTS-1
 - [x] #432
 - [x] #427
 - [x] #494

Stretch goal:

 - [x] #485 
 - [x] #496

metasim:
@lnicola I'd suggest us not including #483 in this release. We have lots of nice new features already, and that one feels like it needs some design/debate. Thoughts?

metasim:
Maybe we should wrap up https://github.com/georust/gdal/pull/493 and add it to the release?

lnicola:
Seems doable, but I don't really feel like messing around with VS (last time I tried to install the DDK and it was awful).

PR #517, opened 2024-01-26
Provide a bundled flag for `gdal-sys`

### LINKED ISSUE #465: Provide a `bundled` feature
This is mostly a feature request. Many `*-sys` crates provide a `bundled` or `vendored` feature that builds and links static version of the corresponding C dependency as part of the build process. It would be great if `gdal` would provide a similar feature as well.

The main motivation for such a feature is that it would greatly simplify setting up the development/deploy environment as everything would be build as part of the build process.

weiznich:
This commit introduces a new `gdal-src` crate which bundles gdal and builds a minimal version from source via `build.rs`

Fixes #465

---


Cargo.toml
```diff
@@ -36,11 +36,16 @@ tempfile = "3.8"
...
 [package.metadata.docs.rs]
 # include `array` feature in documentation
 features = ["array"]
 # define attribute `docsrs` for feature badges
 rustdoc-args = ["--cfg", "docsrs"]
+
+
+[patch.crates-io]
+proj-sys = { git = "https://github.com/GiGainfosystems/proj", rev = "54716dd8955d4f0561ce9bf8a83610b605e3c007" }
+libsqlite3-sys = { git = "https://github.com/rusqlite/rusqlite/", rev = "42a82fb" }
```
weiznich, line 51: This currently relays on an accepted but unreleased patch to the libsqlite3-sys crate: rusqlite/rusqlite#1441

Cargo.toml
```diff
@@ -36,11 +36,16 @@ tempfile = "3.8"
...
 # docs.rs-specific configuration
 [package.metadata.docs.rs]
 # include `array` feature in documentation
 features = ["array"]
 # define attribute `docsrs` for feature badges
 rustdoc-args = ["--cfg", "docsrs"]
+
+
+[patch.crates-io]
+proj-sys = { git = "https://github.com/GiGainfosystems/proj", rev = "54716dd8955d4f0561ce9bf8a83610b605e3c007" }
```
weiznich, line 50: This currently relays on a non-accepted patch to `proj-sys`: https://github.com/georust/proj/pull/190

gdal-src/build.rs
```diff
@@ -0,0 +1,58 @@
...
+        .define("GDAL_USE_EXTERNAL_LIBS", "OFF")
+        .define("BUILD_SHARED_LIBS", "OFF")
+        .define("BUILD_STATIC_LIBS", "ON")
+        .define("BUILD_APPS", "OFF")
+        .define("BUILD_DOCS", "OFF")
+        .define("BUILD_TESTING", "OFF")
+        .define("BUILD_GMOCK", "OFF")
+        .define("PROJ_INCLUDE_DIR", format!("{proj_root}/include"))
+        .define("PROJ_LIBRARY", format!("{proj_root}/lib/{proj_lib}"))
+        // enable the gpkg driver
```
weiznich, line 27: I require the geopackage driver for my usecase that's why I added these configurations. In the end we already depend on `libsqlite3` via proj, so that should hopefully be fine.

That written: Anyone that tries to actually use that driver to create a geopackage dataset will likely run into: https://github.com/OSGeo/gdal/issues/9135 (which hopefully will be fixed soon upstream, so it might be worth to pull in the fix into the bundled version afterwards.)

weiznich:
Marked as draft as I need to perform a final set of tests on MacOS and Windows.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "gdal-src/source"]
```
lnicola, line 1: Should we point this to a specific tag? I'm not sure we want to bundle `master`.

Cargo.toml
```diff
@@ -36,11 +36,16 @@ tempfile = "3.8"
 arrow2 = "0.18"
 
 [workspace]
-members = ["gdal-sys"]
+members = [ "gdal-src","gdal-sys"]
```
lnicola, line 39: Nit:

```suggestion
members = ["gdal-src", "gdal-sys"]
```

CHANGES.md
```diff
@@ -1,6 +1,7 @@
 # Changes
 
 ## Unreleased
+- Add a `bundled` feature for `gdal-sys` that allows to build and statically link a minimal bundled version of gdal during `cargo build`
```
lnicola, line 4: We should probably mention that it only supports SQLite, GPKG, GeoTIFF, some other less popular formats, and probably no compression. I'm afraid that users will see this and expect it to work for COG, JPEG2000, NetCDF and so on.

Also see https://www.mail-archive.com/gdal-dev@lists.osgeo.org/msg40172.html.

gdal-src/build.rs
```diff
@@ -0,0 +1,58 @@
...
+    let proj_root = PathBuf::from(std::env::var("DEP_PROJ_ROOT").expect("set by proj-sys"));
+    let proj_lib = if proj_root.join("lib").join("proj_d.lib").exists() {
+        "proj_d.lib"
+    } else if proj_root.join("lib").join("proj.lib").exists() {
+        "proj.lib"
+    } else {
+        "libproj.a"
+    };
+
+    let res = cmake::Config::new("source")
```
lnicola, line 14: For anyone wondering, `cmake` [automatically sets](https://docs.rs/cmake/latest/cmake/struct.Config.html#method.profile) `CMAKE_BUILD_TYPE`.

lnicola:
This looks all right, but as I mentioned in the related issue, I'm worried it's only going to be useful to you.

GDAL is pretty large, and its users might have a lot of expectations, but a version that only supports GPKG is going to be very surprising. Even _your_ use case might be better served by https://docs.rs/geozero.

CC @rouault long-term, how do you feel about [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html) (SuperBuild)?

gdal-sys/src/lib.rs
```diff
@@ -4,4 +4,7 @@
 #![allow(clippy::upper_case_acronyms)]
 #![allow(rustdoc::bare_urls)]
 
+#[cfg(feature = "bundled")]
```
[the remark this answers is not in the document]
weiznich: That forces cargo to actually build and link that dependency. Otherwise it's clever and says that it is unused and therefore does not link the dependency.

CHANGES.md
weiznich: The following 19 drivers are supported by this configuration according to `DriverManager::all()`:

```
Virtual Raster
Derived datasets using VRT pixel functions
GeoTIFF
Cloud optimized GeoTIFF generator
Erdas Imagine Images (.img)
In Memory Raster
Geographic Network generic file based model
Geographic Network generic DB based model
ESRI Shapefile
MapInfo File
VRT - Virtual Datasource
Memory
Keyhole Markup Language (KML)
GeoJSON
GeoJSON Sequence
ESRIJSON
TopoJSON
GeoPackage
SQLite / Spatialite
```

weiznich:
> This looks all right, but as I mentioned in the related issue, I'm worried it's only going to be useful to you.
>
> GDAL is pretty large, and its users might have a lot of expectations, but a version that only supports GPKG is going to be very surprising. Even your use case might be better served by https://docs.rs/geozero.

I agree with the point that users might have expectations that are not covered (yet) by the proposed bundling support, but I disagree with with the point that it only supports `GPKG` and therefore is not useful. See [that comment](https://github.com/georust/gdal/pull/517#discussion_r1467398669) for the list of supported drives. That list contains a few more useful drivers like that one for shapefiles, geojson, geotiff or kml. In addition gdal is more than just a tool that provides support for certain file formats. It also exposes other functionality that is helpful in certain situations, like support for spatial projections.

As for other drivers: It is certainly possible to enable other drivers as well. That mostly requires going through the list of supported drivers and classify them if they need external dependencies or not. The later ones can easily be enabled, while for the former ones more work is required. Depending on the required external dependency that might require adding another sys crate as dependency and enabling the bundling support there (like for example for `netcdf`, which already has bundling support in rust) or it might require writing that bundling support for that other sys crate first. I figured out to start with a minimal subset to have something working first. 

> Even your use case might be better served by https://docs.rs/geozero.

Trust me if I say that no geozero won't solve my use-case as it has to much assumptions around which geometry types exist and how they should be handled for my use-case. (It's specifically not about translating from one geometry format to another).

lnicola:
Yes, I think it's all right. We should probably document our features anyway, and that's a good place where we can explain the limitations of the bundled GDAL.

weiznich:
I just checked the documentation. GDAL already provides a list of dependencies for each driver:

https://gdal.org/drivers/vector/index.html
https://gdal.org/drivers/raster/index.html

Anything that lists "Built-in by default" as dependency can probably just be enabled without problems.

I also know that bundling support exists for the following other dependencies:

* hdf5: https://github.com/aldanor/hdf5-rust/blob/master/hdf5-sys/Cargo.toml#L30
* netcdf: https://github.com/georust/netcdf/blob/master/netcdf-sys/Cargo.toml#L33
* libpq: https://github.com/sgrif/pq-sys/blob/master/Cargo.toml#L28
* libcurl: https://github.com/alexcrichton/curl-rust/blob/main/curl-sys/build.rs#L108

So yes, it's probably possible to just enable most of the expected drives, although I personally would prefer having all the native dependencies behind separate feature flags.

lnicola:
If you're not familiar with it, ExternalProject is a CMake component that can download libraries, apply patches, then link your project with them. So it can be useful for people who want a custom build but don't want to build every dependency manually.

If GDAL adds that in the future, we can integrate with it without creating Rust libraries for that bundle the source code of the GDAL dependencies.

If it doesn't, that's fine, people can still use GeoTIFF, GPKG, SHP and a a few other formats with the approach here.

rouault:
ok, I see. That's something I've considered, but I'd be hesitant to go into that business, as it equates to creating yet another packaging system, and for GDAL, with all its transitive dependencies, that can mean ~ 100 libraries for a full build... That said I do recognize that might be a recurring need for people in "unstandard" environments (Android, WASM, etc.) to have to rebuild the whole world. I'm not sure if that would belong to the OSGeo/GDAL repo itself, or if it might be a side repository where people from different teams can collaborate, and be independent of GDAL development itself. Might be worth raising the idea on gdal-dev.

lnicola:
Yeah, that makes sense. You've already got a full plate with GDAL and the other libraries you're maintaining, you don't need to get into a whole new packaging system unless you really want to.

Cargo.toml
```diff
@@ -36,11 +36,17 @@ tempfile = "3.8"
...
 # include `array` feature in documentation
 features = ["array"]
 # define attribute `docsrs` for feature badges
 rustdoc-args = ["--cfg", "docsrs"]
+
+
+[patch.crates-io]
+proj-sys = { git = "https://github.com/GiGainfosystems/proj", rev = "dc67b5afc17bc7d330371e4406dffd65cf9321cd" }
+libsqlite3-sys = { git = "https://github.com/rusqlite/rusqlite/", rev = "42a82fb" }
+pq-sys = { git = "https://github.com/sgrif/pq-sys", rev = "77e6729" }
```
weiznich, line 52: This requires https://github.com/sgrif/pq-sys/pull/50 

(I will likely merge this and cut a `0.1.1` release in the next days. Afterwards I will update this PR)

gdal-src/Cargo.toml
```diff
@@ -0,0 +1,314 @@
...
+netcdf-src = { version = "0.3.2", optional = true }
+pq-src = { version = "0.1", optional = true }
+curl-sys = { version = "0.4.70", features = ["static-curl"], optional = true }
+libz-sys = { version = "1.1.15", features = ["static"], optional = true }
+
+[build-dependencies]
+cmake = "0.1.50"
+
+[features]
+default = []
```
weiznich, line 22: It's up to discussion which drivers should be build by default as cargo feature flags. We cannot easily say build everything and allow users to opt out as cargo feature flags are additive.

weiznich:
I've pushed a new version of the build script that adds feature flags for a lot of the built in drivers. If I enable all of them `DriverManager::all` reports support for the following 162 drivers:

<details>
<summary>List with 162 driver names</summary>

```
Virtual Raster
Derived datasets using VRT pixel functions
GeoTIFF
Cloud optimized GeoTIFF generator
Erdas Imagine Images (.img)
CEOS SAR Image
CEOS Image
JAXA PALSAR Product Reader (Level 1.1/1.5)
Ground-based SAR Applications Testbed File Format (.gff)
ELAS
Arc/Info Binary Grid
Arc/Info ASCII Grid
GRASS ASCII Grid
International Service for the Geoid
SDTS Raster
DTED Elevation Raster
Portable Network Graphics
JPEG JFIF
In Memory Raster
Japanese DEM (.mem)
Graphics Interchange Format (.gif)
Graphics Interchange Format (.gif)
Envisat Image Format
Maptech BSB Nautical Charts
X11 PixMap Format
MS Windows Device Independent Bitmap
SPOT DIMAP
AirSAR Polarimetric Image
RadarSat 2 XML Product
Sentinel-1 SAR SAFE Product
PCIDSK Database File
PCRaster Raster File
ILWIS Raster Map
SGI Image File Format 1.0
SRTMHGT File Format
Leveller heightfield
Terragen heightfield
Network Common Data Format
EarthWatch .TIL
ERMapper .ers Labelled
NOAA Polar Orbiter Level 1b Data Set
FIT Image
GRIdded Binary (.grb, .grb2)
Raster Matrix Format
OGC Web Coverage Service
OGC Web Map Service
EUMETSAT Archive native (.nat)
Idrisi Raster A.1
Golden Software ASCII Grid (.grd)
Golden Software Binary Grid (.grd)
Golden Software 7 Binary Grid (.grd)
COSAR Annotated Binary Matrix (TerraSAR-X)
TerraSAR-X Product
DRDC COASP SAR Processor Raster
R Object Data Store
OziExplorer .MAP
Kml Super Overlay
Planet Labs Mosaics API
CALS (Type 1)
OGC Web Map Tile Service
Sentinel 2
Meta Raster Format
Portable Pixmap Format (netpbm)
USGS DOQ (Old Style)
USGS DOQ (New Style)
PCI .aux Labelled
Vexcel MFF Raster
Vexcel MFF2 (HKV) Raster
GSC Geogrid
EOSAT FAST Format
VTP .bt (Binary Terrain) 1.3 Format
Erdas .LAN/.GIS
Convair PolGASP
NLAPS Data Format
Erdas Imagine Raw
DIPEx
FARSITE v.4 Landscape File (.lcp)
NOAA Vertical Datum .GTX
NADCON .los/.las Datum Grid Shift
NTv2 Datum Grid Shift
CTable2 Datum Grid Shift
ACE2
Snow Data Assimilation System
KOLOR Raw
ROI_PAC raster
R Raster
Natural Resources Canada's Geoid
NOAA GEOCON/NADCON5 .b format
NSIDC Sea Ice Concentrations binary (.bin)
Swedish Grid RIK (.rik)
USGS Optional ASCII DEM (and CDED)
GeoSoft Grid Exchange Format
Bathymetry Attributed Grid
S-102 Bathymetric Surface Product
Hierarchical Data Format Release 5
HDF5 Dataset
Northwood Numeric Grid Format .grd/.tab
Northwood Classified Grid Format .grc/.tab
ARC Digitized Raster Graphics
Standard Raster Product (ASRP/USRP)
Magellan topo (.blx)
SAGA GIS Binary Grid (.sdat, .sg-grd-z)
ASCII Gridded XYZ
HF2/HFZ heightfield raster
OziExplorer Image File
USGS LULC Composite Theme Grid
ZMap Plus Grid
NOAA NGS Geoid Height Grids
IRIS data (.PPI, .CAPPi etc)
Racurs PHOTOMOD PRF
Scaled Integer Gridded DEM .sigdem
TGA/TARGA Image File Format
OGCAPI
Spatio-Temporal Asset Catalog Tiled Assets
Spatio-Temporal Asset Catalog Items
Geographic Network generic file based model
Geographic Network generic DB based model
ESRI Shapefile
MapInfo File
UK .NTF
SDTS
IHO S-57 (ENC)
Microstation DGN
VRT - Virtual Datasource
Memory
Comma Separated Value (.csv)
Keyhole Markup Language (KML)
GeoJSON
GeoJSON Sequence
ESRIJSON
TopoJSON
GMT ASCII Vectors (.gmt)
GeoPackage
SQLite / Spatialite
WAsP .map format
PostgreSQL/PostGIS
ESRI FileGDB
AutoCAD DXF
AutoCAD Driver
FlatGeobuf
Geoconcept
Czech Cadastral Exchange Data Format
PostgreSQL SQL dump
French EDIGEO exchange format
Idrisi Vector (.vct)
Elastic Search
Carto
AmigoCloud
Storage and eXchange Format
Selafin
VDV-451/VDV-452/INTREST Data Format
NextGIS Web
MapML
General Transit Feed Specification
OGC Features and Geometries JSON
U.S. Census TIGER/Line
Arc/Info Binary Coverage
Arc/Info E00 (ASCII) Coverage
Generic Binary (.hdr Labelled)
ENVI .hdr Labelled
ESRI .hdr Labelled
ISCE raster

```
</details>

Notably this is still missing support for anything that depends on `libexpat` or `libmysqlclient` and likely a few other more specific dependencies. For `libexpat` there does not seem to be a crate with up to date rust bindings at all, the `libmysqlclient-sys` crate still misses bundling support. (With my diesel hat on, the last thing is something I  want to address at some point).

It's up to discussion which of these drivers should be enabled by default.

weiznich:
I confirmed that this works on linux (x86_86), windows (msvc, x86_64) and macos (aarch64).

weiznich:
Points that should be likely discussed (and addressed) before merging:

- [ ] Merge all the upstream PR's
- [ ] Select a default set of drivers
- [ ] Write some documentation (where?)
- [ ] Add some CI configuration

Especially the later 3 points are things that I believe are project specific so I would like get some guidance from the maintainers on that.

.gitmodules
```diff
@@ -0,0 +1,4 @@
...
+	path = gdal-src/source
+	url = https://github.com/OSGeo/gdal
+	rev = 654f4907abbbf6bf4226d58a8c067d134eaf3ce9
```
ChristianBeilschmidt, line 4: gdal has tags for versions, e.g., https://github.com/OSGeo/gdal/tree/v3.8.3

wouldn't it be better to link to a fixed version instead of a cryptic revision number (I have no idea if this is a version or just a snapshot from GDAL's master)?

Cargo.toml
ChristianBeilschmidt, line 52: also here, would it be possible to link to some fixed version rather then some revisions in the GitHub repos?

Cargo.toml
```diff
@@ -36,11 +36,17 @@ tempfile = "3.8"
...
 
 # docs.rs-specific configuration
 [package.metadata.docs.rs]
 # include `array` feature in documentation
 features = ["array"]
 # define attribute `docsrs` for feature badges
 rustdoc-args = ["--cfg", "docsrs"]
+
+
+[patch.crates-io]
```
ChristianBeilschmidt, line 49: is it necessary to patch it here or in `gdal-src`?

gdal-src/Cargo.toml
```diff
@@ -0,0 +1,314 @@
+[package]
+name = "gdal-src"
+version = "0.1.0"
```
ChristianBeilschmidt, line 3: Will this be published separately?
If not, should it take the version from the workspace?

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "gdal-src/source"]
```
ChristianBeilschmidt, line 1: Is this common practice to have gdal as a git submodule or would it be better to clone it when it is needed?

ChristianBeilschmidt (changes requested):
I did not look into the actual bundling logic, I just have some questions about the surroundings.

.gitmodules
weiznich: It seems like that this is not possible. Even the `rev` [specifier is not relevant](https://git-scm.com/docs/gitmodules) according to the documentation. That [stackoverflow comment](https://devops.stackexchange.com/questions/5371/where-is-the-git-submodule-revision-actually-stored) says that git stores the rev in the object tree itself.

So I'm not sure how you want to handle that. I've gone with a submodule as that makes it easier to update the code + pull in other changes/patches.

(The commit currently refers to the 3.8.3 release: https://github.com/OSGeo/gdal/releases/tag/v3.8.3, see the commit mentioned there)

Cargo.toml
weiznich: That's certainly possible as soon as those patches are merged in these repositories and fixed versions are released. For `libsqlite3-sys` that's already the case, for `pq-sys` I need to cut an release, for `proj-sys` I need to wait for the maintainers there to decide what to do.

Cargo.toml
weiznich: For cargo workspaces that section needs to be in the workspace `Cargo.toml`:

> Cargo only looks at the patch settings in the Cargo.toml manifest at the root of the workspace. Patch settings defined in dependencies will be ignored.

[Source](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section) (last sentence)

.gitmodules
lnicola: It feels like cloning could be more problematic for distros and people who want to build crates without network access.

Having a separate crate only for the source code was less familiar to me, but I guess it makes sense.

gdal-src/Cargo.toml
weiznich: Will do that.

.gitmodules
weiznich: My reasoning for using a submodule here is that this makes it much easier to test these changes locally as everything is already there. It also makes it easier to checkout potential problems and pull in additional fixes as that just need to change the commit of the submodule. 
A published version of `gdal-src` would then include a bundled version of the gdal source code inside the tar file uploaded to crates.io. That would allow other to build that package without having a git setup that allows the build script to clone gdal from somewhere. It also would prevent the build script from downloading things from the internet, which might be problematic in some situations.

.gitmodules
ChristianBeilschmidt: Maybe we do not need to use git then. They provide bundled dependencies as zip/tar.gz.

weiznich: The problem with not using git would be that it's much harder to update the bundled version (need to download a new tar file instead of just bumping the submodule) and that it would be much harder to pull in additional fixed (need to create your own tar instead of just bumping the submodule).

gdal-src/Cargo.toml
weiznich: I just tried doing that, but there is currently no workspace wide version defined. It also seems like the `gdal` crate and the `gdal-sys` crate use different versions

Cargo.toml
weiznich: I've cut a 0.1.1 release of pq-sys and removed the patch section here afterwards.

weiznich: Patches for pq-sys and libsqlite3-sys are removed as the relevant releases are published to crates.io. For proj-sys is still wait for feedback from the maintainers.

.gitmodules
ChristianBeilschmidt: Well, does it make a difference if we pull a subrepo or download a tar?
Moreover, we shouldn't do any fixes to GDAL that are not part of GDAL, I guess. Otherwise, we will ship a non-existing version.

gdal-src/Cargo.toml
ChristianBeilschmidt: I guess NetCDF and Zarr could be interesting.

weiznich: The zarr driver so currently not not supported by the bundling. It would require figuring out how to bundle all the dependencies for that driver as well. 

Netcdf is supported, as there is already a rust crate which provides bundling support and exposes the relevant information. Therefore it should be possible to enable it by default.

.gitmodules
weiznich: For the end user it does not make a difference. For a person that actively works on the code it makes a huge difference. By having a sub repository it's simple to pull in fixes from upstream and in turn also submit fixes to upstream. Both are things that already happened several times while working on that code. 

That written: If you really prefer having a static tar in your gut repository I can certainly change to that, but I highly doubt that this is a good solution.

ChristianBeilschmidt: Can we use the tag instead of the commit hash for the submodule?
If not, can we add a comment for the tag that is represented by the commit hash?

weiznich: Neither a tag nor a commit has is meaningful in that file according to the documentation linked above. That information in tracked in git itself. So I do not see adding a comment there, especially as it's likely to become outdated.

ChristianBeilschmidt: Well, you don't want to use a release but rather the repo. Then, there is just a commit hash and not a tag name (that would refer to a release). So as a developer, I have to backtrack the commit hash to a tag and then to a GDAL version. If at least there is a comment that tells me "This hash refers to tag 3.8.4" and this was checked at the PR that changed it, it could be helpful.

weiznich: @ChristianBeilschmidt I really feel that there got something lost here, so let me write that out explicitly again:

* There is no `rev` nor a `tag` entry specified for `.gitmodules` files. 
* Information about which commit/tag is referenced for the given submodule is stored directly in git itself (essentially like the content of other files as well).
* (The `rev` specifier will be removed from this file with the next commit, I just wait for a full review of the whole PR instead of trying to address minor feedback again and again)

Therefore could you clarify where exactly you would like to see that comment about which commit hash exactly?

Documentation for submodules: https://git-scm.com/docs/gitmodules

weiznich: I'm not aware of any way to change how this is displayed in the github user interface. If you checkout the repo locally it will correctly show that there is a tag that corresponds to this commit.

weiznich:
Is there any interest in moving this PR forward from the relevant maintainers?  I'm still waiting on a full review

lnicola:
Yeah, I even realize you've added features for the default drivers. I want to take a better look, but it would be very nice to have this, even with the format limitations.

Cargo.toml
```diff
@@ -36,11 +36,15 @@ tempfile = "3.8"
...
+proj-sys = { git = "https://github.com/GiGainfosystems/proj", rev = "44a503869709ea793a5f8452bfb9e2a99b7160fc" }
```
lnicola, line 50: Is this temporary until your three `proj` PRs get merged? Do you want to merge this before them?

gdal-src/Cargo.toml
```diff
@@ -0,0 +1,329 @@
...
+geos-src = { version = "0.2.1", optional = true }
+geos-sys = { version = "2", optional = true }
+
+
+[build-dependencies]
+cmake = "0.1.50"
+
+[features]
+default = []
+enable_all_drivers = [
```
lnicola, line 27: Should these be `all_drivers` / `internal_drivers`? :bike: :house:

gdal-src/Cargo.toml
```diff
@@ -0,0 +1,329 @@
...
+        "enable_all_internal",
+        "DRIVER_SQLITE",
```
lnicola, line 29: I don't feel too strongly about it, but I'd prefer lower-case here.

gdal-src/Cargo.toml
```diff
@@ -0,0 +1,329 @@
...
+DRIVER_HDF5 = ["dep:hdf5-src", "dep:libz-sys"]
+
+# libnetcdf
+DRIVER_NETCDF = ["dep:netcdf-src", "DRIVER_HDF5", "dep:hdf5-sys"]
+
+# poppler
+#DRIVER_PDF = []
+
+# combined
+# these miss at least one dependency
```
lnicola, line 318:
```suggestion
# these are missing at least one dependency
```

gdal-src/build.rs
```diff
@@ -0,0 +1,351 @@
...
+        .define("BUILD_APPS", "OFF")
+        .define("BUILD_DOCS", "OFF")
+        .define("BUILD_TESTING", "OFF")
+        .define("BUILD_GMOCK", "OFF")
+        .define(
+            "PROJ_INCLUDE_DIR",
+            format!("{}/include", proj_root.display()),
+        )
+        .define("PROJ_LIBRARY", proj_library)
+        .pic(true)
```
lnicola, line 78: Is this needed for a static lib?

lnicola (approved):
The `build.rs` changes are pretty hairy, so I'm unlikely to spot any issues. But I guess we'll find out in time.

Before merging, I think we should add:

 - a CI job for the bundled GDAL, with no and with all drivers enabled
 - some instructions for upgrading `gdal-src` to a newer version of GDAL

metasim (commented):
I'm sympathetic to the need for a fully statically linked version of GDAL, but gosh, this feels like a lot of new maintenance for a small team of maintainers to take on, especially without additional documentation on the design decisions made, liabilities, assumptions, etc. Maybe I'm being too quick to react, or being myopic in my assessment, but it feels like a significant amount of complexity for an unquantified subset of GDAL users who need it. 

So I must ask: Will the PR author be committed to support this capability moving forward for a period of time, or is it a "throw it over the fence" type of PR? 

Personally, I'm going to have a lot less time in the months ahead to help out with the project, and I suspect @lnicola has a lot on his plate, so I think it's important we get CI working rock solid and monitor this carefully through several iterations of GDAL updates. Maybe I'd feel better about it if we compared this against other crates "vendoring" for the purposes of static compilation, to ensure ourselves this is the optimal build architecture.

Apologies if I'm coming across as unenthusiastic. This capability truly could significantly improve deployment for many users. But I'd suggest we do so in a way that ensures its maintainable by volunteers that come and go, and ensure the crate doesn't collapse under its own complexity.


Specific note:

Is `gdal-src/source` a git submodule? Will all users end up triggering a clone of GDAL whether or not the users need a statically linked version? I know from experience that git submodules cause a lot of people problems because they are not widely understood, and we don't want to add any additional friction to new users of the crate.

metasim:
PS: We'll need documentation on how to update the git submodule whenever a new point release of GDAL appears.

Cargo.toml
weiznich: > Is this temporary until your three proj PRs get merged? Do you want to merge this before them?

It only depends on https://github.com/georust/proj/pull/190. The other two PR's are not needed for this feature, although I obviously would prefer getting all of them merged.

> Definitely don't think this should be merged like this. Need to be pointing to the official version.

This is unfortunately still waiting on a review from the relevant maintainers of the georust project.

> Why are we interacting directly with proj-sys anyway? Should we not be going through the C++ gdal wrappers around this? 

There are two reasons for depending on `proj-sys`:

* Statically linking gdal does not link libproj. By depending on `proj-sys` we add a dependency on libproj that is resolved by rustc while linking the final binary.
* gdal requires access to the proj header files for building. By enabling the bundled feature of `proj-sys` we can access the relevant include directory there to get access to the header files

> What happens if the user links in georust/proj into their application? Do we end up with multiple copies of proj?

As it's a `-sys` crate, cargo ensures that there will be only one version of this crate in your dependency tree. Also keep in mind that this all is only relevant if you enable the `bundled` feature for gdal and also depend on `proj-sys`.

weiznich:
Thanks for the reviews.

> a CI job for the bundled GDAL, with no and with all drivers enabled

Which platforms should be covered by that job? All that are supported by github actions, so windows, ubuntu, mac os (x64) and mac os (aarch64)?

> some instructions for upgrading gdal-src to a newer version of GDAL

Assuming that there are no fundamental changes to the gdal build system and that they did not add new dependencies it should be as simple as:

```
cd gdal/gdal-src/source
git fetch origin
git checkout v{whatever-is-the-latest-gdal-version}
cd ../..
git add gdal-src/source
git commit -m "Update gdal to the {whatver-is-the-latest-gdal-version}
```

I'm happy to write that down somewhere in the repository. Should I just put a `Upgrading.md` into the `gdal-src` folder, or do you have any other preferences for this?


> I'm sympathetic to the need for a fully statically linked version of GDAL, but gosh, this feels like a lot of new maintenance for a small team of maintainers to take on, especially without additional documentation on the design decisions made, liabilities, assumptions, etc. Maybe I'm being too quick to react, or being myopic in my assessment, but it feels like a significant amount of complexity for an unquantified subset of GDAL users who need it.

It's quite hard to me to understand on which topic you would like to have additional documentation. The only actively made design decision I made is that different drivers can be enabled via feature flags. There are certainly other solutions there, I just went with the first thing that came into my mind. Otherwise, at least I feel that most of the code is actually a straight forward rust port of what you would do to build gdal locally from source + the required rust linker flags for linking gdal and all required dependencies. 

> So I must ask: Will the PR author be committed to support this capability moving forward for a period of time, or is it a "throw it over the fence" type of PR?

Given that I have more than enough work on my table with diesel, I cannot promise any sort of active maintenance. Given that I plan to actively use at least a subset of this feature (only certain drivers), I will likely submit a PR from time to time to update to a new version (as required by my needs) or to bugs that I encounter. 

> Maybe I'd feel better about it if we compared this against other crates "vendoring" for the purposes of static compilation, to ensure ourselves this is the optimal build architecture.

See the following examples:

* https://github.com/rusqlite/rusqlite/blob/master/libsqlite3-sys/build.rs#L100-L305 (does not use cmake)
* https://github.com/sgrif/pq-sys/blob/master/pq-src/build.rs (does not use cmake)
* https://github.com/georust/proj/blob/main/proj-sys/build.rs#L73-L171(uses cmake + pkg_config)
* https://github.com/georust/netcdf/blob/master/netcdf-src/build.rs (uses cmake)
* https://github.com/aldanor/hdf5-rust/blob/master/hdf5-src/build.rs (uses cmake)
* https://github.com/alexcrichton/openssl-src-rs/blob/main/src/lib.rs (uses the openssl custom build scripts)
* https://github.com/alexcrichton/curl-rust/blob/main/curl-sys/build.rs#L57-L451 (does not use cmake)


> Is gdal-src/source a git submodule? Will all users end up triggering a clone of GDAL whether or not the users need a statically linked version? I know from experience that git submodules cause a lot of people problems because they are not widely understood, and we don't want to add any additional friction to new users of the crate.

Yes it's a submodule. I really depends on what you understand by "user" here. 

If that refers to someone that gets `gdal-src` from crates.io and builds it locally: No they wouldn't need to care about the submodule at all. `cargo publish` puts the source code into an archive which is hosted at `crates.io`. It doesn't interact with your git repository at all for using a published version. (As side note: The current setup would include all of the `gdal` source code into the `gdal-src` package. By using the [`include`/`exclude`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields) `Cargo.toml` fields, it would likely be possible to reduce the package size quite a bit.

If "user" refers to someone that uses this git repository as direct dependency (`gdal = { git = "https://github.com/georust/gdal"}` ) via cargo: They don't need to handle the subrepository. Cargo does automatically handle that for you. (They will need to wait some time for the download to finish, as the `gdal` repo is quite large)

If "user" refers to a potential contributor: Yes, they might need to care about the submodule, but only if the actually want to build `gdal-src` locally. If they contribute to a different part of the code base it might be fine.

lnicola:
> Which platforms should be covered by that job? All that are supported by github actions, so windows, ubuntu, mac os (x64) and mac os (aarch64)?

Linux x64 since that's what we're testing today. Windows might be nice, but there are already 4 different old PRs for that, so we probably shouldn't block on it.

> Should I just put a Upgrading.md into the gdal-src folder, or do you have any other preferences for this?

I think that's usually for telling users how to upgrade. Let's call it `DEVELOPMENT.md` or something like that and put it in the repo root. We (I) also need to document how to generate new bindings, so that will be a good place for it.

> it should be as simple as [...]

LGTM, but let's have it written down.

> if we compared this against other crates "vendoring"

I'm comfortable with the repo structure, less so with the `build.rs` changes. But I don't think there's any way to simplify it.

gdal-src/Cargo.toml
```diff
@@ -0,0 +1,331 @@
...
+DRIVER_NETCDF = ["dep:netcdf-src", "DRIVER_HDF5", "dep:hdf5-sys"]
+
+# poppler
+#DRIVER_PDF = []
+
+# combined
+# these are missing at least one dependency
+# DRIVER_MVT = [] # requires sqlite + geos
+# DRIVER_OSM = [] # requires sqlite + libexpat
+# DRIVER_ZARR = [] # requires liblz4, libxz, libzstd and libblocs
```
lnicola, line 323:
```suggestion
# DRIVER_ZARR = [] # requires liblz4, libxz, libzstd and libblosc
```

gdal-src/build.rs
weiznich: That's not required, I've removed it.

Cargo.toml
```diff
@@ -36,11 +37,16 @@ tempfile = "3.8"
...
+pq-src = { git = "https://github.com/sgrif/pq-sys", rev = "7300e96574cc0eeb95985881fbbc2cf95886cca2" }
```
weiznich, line 52: I found another bug in the pg-src crate while setting up the CI here. I will issue an release in the next few days and update this PR afterwards to depend on the fixed version instead

weiznich:
> Linux x64 since that's what we're testing today. Windows might be nice, but there are already 4 different old PRs for that, so we probably shouldn't block on it.

I've added a CI job for linux (x64), windows (x64) and macos (x64 and aarch64) for the bundled build. It takes quite a bit of time to build everything there but it seems to work. These jobs will try to build with the minimal feature set enabled first and then run tests with all drivers enabled. Running tests with the minimal feature set enabled seems to be not possible as that is missing quite a lot of drivers to execute all tests.

Cargo.toml
weiznich: Switched to a release version in https://github.com/georust/gdal/pull/517/commits/cf39e1ff1e9fce2a59df334b9e4344e0231ee9f7

weiznich:
It might sometimes be not possible to statically link both libpq and libnetcdf because both include a vendored copy of `strlcat`. There is the following upstream issue for this https://github.com/Unidata/netcdf-c/issues/927

Cargo.toml
```diff
@@ -36,11 +37,16 @@ tempfile = "3.8"
...
 [package.metadata.docs.rs]
 # include `array` feature in documentation
 features = ["array"]
 # define attribute `docsrs` for feature badges
 rustdoc-args = ["--cfg", "docsrs"]
+
+
+[patch.crates-io]
+proj-sys = { git = "https://github.com/GiGainfosystems/proj", rev = "44a503869709ea793a5f8452bfb9e2a99b7160fc" }
+netcdf-src = { git = "https://github.com/GiGainfosystems/netcdf", rev = "292309979f533f71d542f4fcf8fdac8830f0e22b" }
```
weiznich, line 52: I've opened https://github.com/Unidata/netcdf-c/pull/2906 to fix the netcdf duplicated symbol issue. Let's see if upstream accepts that PR, after that we need to merge the fix into the rust bindings crate. (Which is under the georust github organization)

metasim:
@lnicola Before this PR is merged, could we consider cutting a [0.17 Release](https://github.com/georust/gdal/issues/484)?. I say we go with what we have. (I'm not going to have time to wrap up #508 before then. :/) It's been in the wild for a while, and it would be nice to close that chapter before a big change like this one.

Cargo.toml
weiznich: The upstream PR to netcdf itself is accepted and merged :tada:

I've opened one against the rust bindings here to include the fix there as well: https://github.com/georust/gdal/pull/517

weiznich: I've pushed https://github.com/georust/gdal/pull/517/commits/50dd40516c85fbf26d46632fe3f51c1c61e42141 which pulls in netcdf-src 0.3.3 which contains the relevant fix.

Cargo.toml
weiznich: I pushed https://github.com/georust/gdal/pull/517/commits/27c2a5e396089f57bdd7a64f81c20183245b44bf which changes this to use the master branch of proj-sys instead as the relevant patch is now merged there. There is still no release that contains this change yet.

weiznich: Seems like we still need https://github.com/georust/proj/pull/192/files

weiznich:
@lnicola @metasim With gdal 0.17 now being released: Is there any chance to land this now? What is still missing from your point of view.

lnicola:
I'm away from the computer until Monday, but I've been meaning to get back to this afterwards, especially at the latest changes.

But I think we can merge it soon. We can also fix things in later PRs if anything comes up.

weiznich:
No worries, take your time off. It's not that important to merge it literally in this moment. The comment was mostly meant to bring this PR back to your attention.

lnicola:
I'm still a bit worried about the massive complexity here, but there's not much we can do about it. So LGTM and thanks for the massive effort you've put into this!

Looks like this still needs a rebase, and there's a pretty critical fix in #550 which I hope we can release once someone reviews it. Then I think we should be good to go, unless anyone speaks out.

weiznich:
> I'm still a bit worried about the massive complexity here, but there's not much we can do about it.

If there is anything I can do to lower the complexity, just let me know. That written: Most of the changes are just in the new `build.rs` file and most of that is just setting the right config flags for the cmake build system. It looks like a lot and it looks complex, but its mostly repetitive code.

> Looks like this still needs a rebase,

I'm not sure about that. Github doesn't show any conflicts so it should be possible to merge it directly? Or is there any policy to have everything always on top of the current master state?

There is one other thing that you might want to consider doing before merging: This currently relies on a unpublished proj4-sys version. It might be meaningful to cut a release there first and then change the dependencies here to the released version?

lnicola:
> If there is anything I can do to lower the complexity, just let me know.

I don't think so, something has to drive the 10KLOC of CMake that GDAL has :slightly_smiling_face:.

> Github doesn't show any conflicts so it should be possible to merge it directly?

You're right, I assumed it wasn't going to work because I had Rebase selected in the GitHub UI:

image

> This currently relies on a unpublished proj4-sys version. It might be meaningful to cut a release there first and then change the dependencies here to the released version?

I don't expect us to publish 0.18 too soon, we're doing one release per year or so. But yeah, ideally we'd use a published version there.

weiznich:
This is another bump. Would it be possible to make finally some progress here?

lnicola:
I wonder what we can do to get that `proj-sys` version published...

Cargo.toml
weiznich: That's finally released with proj-sys 0.24 :tada:

weiznich:
@lnicola I've updated the PR to use the newly published proj-sys version

lnicola:
Yup, that's great. I guess this is now blocked on:

 - #550 - I'm not a fan of deprecating that impl, and we can't `#[deprecate]` an `impl Trait` anyway IIRC
 - someone else approving this

weiznich:
@ChristianBeilschmidt @metasim Are there any chances to get another review on this?

jdroenner:
I will try to test it this weekend.

weiznich:
The current CI failures are very likely caused by the rust 1.81 release, which brings new clippy lints. I might have a look at them tomorrow.

lnicola:
https://github.com/georust/gdal/pull/552 fixes the lint.

weiznich:
@jdroenner Do you have any other comments here? It is really frustrating that people just seem to come around, put a quick comment (often about already discussed topics) at the PR and then vanish again for weeks. I can fully understand that this is a free time project for all of you and you are short on time for anything. As a maintainer of a larger project I only can give you the advice to communicate the lack of time openly instead of putting up such low effort responses. I mean it's totally fine to say: No we won't want to have this, although I personally would have really appreciated such a response already last year as I opened #465 which asked if you were interested in that or at least half a year ago at the time I've put up the initial PR. 

Anyway: I would kindly ask for a final decision if you want to merge this or not.

jdroenner:
@weiznich again sorry for a late reply. I already commented in another issue that i don't have much time at the moment.

Regarding the PR: Since it adds an optional feature that no one has to enable i think we should merge it. I know there are situations where i want to use it. However, I am really worried about maintaining it. If the submodule is the only viable way to got then let's try it. We can still change it if something goes wrong.

jdroenner (approved):

weiznich:
@jdroenner Thanks for the response :heart: 

> However, I am really worried about maintaining it. If the submodule is the only viable way to got then let's try it. 

I still would like to hear some details on this. 
Independently from your answer: I can offer to do a submodule update from time to time after this is merged. I personally wouldn't expect much more trouble there than bumping the commit from time to time, at least as long as nobody finds bugs in gdal itself.

lnicola:
IMO, the submodule is the most boring part of this. I don't like them, but the alternatives are worse (as you've clearly shown above).

I'm more worried about the CMake config and proliferation of features, but we'll see how problematic they are.

Isaac-Leonard:
I'm attempting to use this PR in a project that also uses the geo package with the "use-proj" feature and getting version confilicts for proj_sys with the key part of the error being:
the package `proj-sys` links to the native library `proj`, but it conflicts with a previous package which links to `proj` as well:

Is there a way to specify which proj version to link to?

weiznich:
@Isaac-Leonard cargo only allows one version of a `*-sys` tree in your dependency tree as these crates always link to native dependencies and those cannot appear multiple times. This change needs a very recent proj-sys version as we needed to adjust the build script there as well to work correctly. It seems like geo just depends on an older version, so they would need to update to the new version to be compatible. Maybe open a PR there to bump the version?

Isaac-Leonard:
I was able to fix it by patching dependencies to the proj repo and it works grate.
Another question I had  though I'm not sure if this is the best place to ask but don't know anywhere better, is if it is possible to build the apps as well, currently they're disabled in the build.rs file but if enabled is there a reasonably easy way to access them?
I've got an app I'm trying to package up and would like to make it so people don't need to install gdal themselves but I'm using the commandline tools where the bindings are lacking.
I'm not familiar enough with the build process to know if I can extract files from dependancies build outputs though I suspect I can't.

weiznich:
@Isaac-Leonard It might be possible to do something like that, but at least for me that's clearly out of scope for this PR as it is additional functionality. I suggest that you check with the rust-gdal maintainers if they are comfortable with adding a new feature flag for building the binaries as well. After that you would need to change the `BUILD_APPS` flag here to `ON` depending on whether the feature flag is set or not. Afterwards you would need to reexport a environment variable containing the build dir root so that other cares can consume this that and find the binaries.

lnicola:
Regarding the apps, I think we should instead offer bindings for them. It's not too complicated, just that there's a ton of API surface to bikeshed. And, as weiznich said, it's a bit unusual to build apps together with a library crate.

That said, can we merge this now?

jdroenner:
yes lets merge it

weiznich:
I'm naturally interested in seeing a release that contains this new feature. Is there anything that I could do to make this happen faster?

lnicola:
Let's pick a reasonable cut-off date (say 2-4 weeks from now), in case we can get in any other PRs until then. Would that time frame work for you?

weiznich:
That would be really nice. It's also fine for me if it takes a few weeks longer, but I would prefer not to have wait months for this, as cloning the submodule everytime in CI is a possible source of network failures.

lnicola:
Or let's just publish it as -alpha.0 or something.

PR #519, opened 2024-01-31
Unified raster creation options over `CplStringList`.

metasim:
---

This:

```rust
let options = [
    RasterCreationOption {
        key: "TILED",
        value: "YES",
    },
    RasterCreationOption {
        key: "BLOCKXSIZE",
        value: "16",
    },
    RasterCreationOption {
        key: "BLOCKYSIZE",
        value: "16",
    },
];

```

is now this:

```rust
let options = RasterCreationOptions::from_iter([
   "TILED=YES", "BLOCKXSIZE=16", "BLOCKYSIZE=16"
]);
```

### DIFF
--- a/src/raster/create_options.rs
+++ b/src/raster/create_options.rs
@@ -1,0 +1,7 @@
+use crate::cpl::CslStringList;
+
+/// Key/value pairs of options for passing driver-specific creation flags to
+/// [`Driver::create_with_band_type_with_options`](crate::Driver::create_with_band_type_with_options`).
+///
+/// See `papszOptions` in [GDAL's `Create(...)` API documentation](https://gdal.org/api/gdaldriver_cpp.html#_CPPv4N10GDALDriver6CreateEPKciii12GDALDataType12CSLConstList).
+pub type RasterCreationOptions = CslStringList;
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -578,23 +578,12 @@ impl<'a> RasterBand<'a> {
     /// ```rust, no_run
     /// # fn main() -> gdal::errors::Result<()> {
     /// use gdal::DriverManager;
-    /// use gdal::raster::{Buffer, RasterCreationOption};
+    /// use gdal::raster::{Buffer, RasterCreationOptions };
     ///
     /// let driver = DriverManager::get_driver_by_name("GTiff").unwrap();
-    /// let options = [
-    ///     RasterCreationOption {
-    ///         key: "TILED",
-    ///         value: "YES",
-    ///     },
-    ///     RasterCreationOption {
-    ///         key: "BLOCKXSIZE",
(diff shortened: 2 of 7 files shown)


src/raster/create_options.rs
```diff
@@ -0,0 +1,7 @@
+use crate::cpl::CslStringList;
+
+/// Key/value pairs of options for passing driver-specific creation flags to
+/// [`Driver::create_with_band_type_with_options`](crate::Driver::create_with_band_type_with_options`).
+///
+/// See `papszOptions` in [GDAL's `Create(...)` API documentation](https://gdal.org/api/gdaldriver_cpp.html#_CPPv4N10GDALDriver6CreateEPKciii12GDALDataType12CSLConstList).
+pub type RasterCreationOptions = CslStringList;
```
metasim, line 7: Not 100% sure if

a) A type alias is a good idea
b) It should be in a separate file

Rationale:

a) If we ever needed to convert it into a newtype (to add creation-specific methods like `validate`, it's less breaking.
b) Putting a single type in `mod.rs` feels unbalanced.

metasim:
Question: Should we also update DatasetOptions::open_options to do something similar?

https://github.com/georust/gdal/blob/1b8ba56e5c9264435b5f29ac518561fb58f739de/src/options.rs#L7-L12

src/raster/create_options.rs
lnicola: > a) If we ever needed to convert it into a newtype (to add creation-specific methods like validate, it's less breaking. Also, it's easy to get create options and other options confused, so this makes it a bit more clear.

We should probably re-export it from the parent module, so not breaking.

> b) Putting a single type in mod.rs feels unbalanced.

There might be more :slightly_smiling_face:.

lnicola (approved):
LGTM, but I secretly wish we just took `&[&str]` and built those under the scenes. :sweat_smile:

src/raster/create_options.rs
metasim: > We should probably re-export it from the parent module, so not breaking.

It is: 

https://github.com/georust/gdal/pull/519/files#diff-eb172c9758a6ca0ee77f152d6a8c864de3a68ea11439bb26fa33ea60d3490155R78

src/raster/rasterband.rs
```diff
@@ -578,23 +578,12 @@ impl<'a> RasterBand<'a> {
     /// ```rust, no_run
     /// # fn main() -> gdal::errors::Result<()> {
     /// use gdal::DriverManager;
-    /// use gdal::raster::{Buffer, RasterCreationOption};
+    /// use gdal::raster::{Buffer, RasterCreationOptions };
```
lnicola, line 581: Nit:

`/// use gdal::raster::{Buffer, RasterCreationOptions };`

PR #520, opened 2024-02-01
Added `no_data_value_u64`, `set_no_data_value_u64`, `no_data_value_i64` and `set_no_data_value_i64` to `RasterBand`

metasim:
---

Adds `RasterBand::no_data_value_u64` and `RasterBand::no_data_value_i64`, etc.


lnicola:
Can you resample those to 1x1 and maybe merge them into a single file?

src/raster/rasterband.rs
```diff
@@ -696,7 +696,61 @@ impl<'a> RasterBand<'a> {
         None
     }
 
-    /// Set the no data value of this band.
+    /// Fetch the no-data value for this band.
+    ///
+    /// This method should ONLY be called on rasters whose data type is `UInt64`.
+    ///
+    /// If there is no no-data value, an out of range value will generally be returned.
```
lnicola, line 703: I don't think this makes sense.

src/raster/rasterband.rs
```diff
@@ -696,7 +696,61 @@ impl<'a> RasterBand<'a> {
...
+            return Some(no_data);
+        }
+        None
+    }
+
+    /// Fetch the no data value for this band.
+    ///
+    /// This method should ONLY be called on rasters whose data type is `Int64`.
+    ///
+    /// If there is no out of data value, an out of range value will generally be returned.
```
lnicola, line 730: Same, but worse.

metasim: It's what the docs say 🤷

metasim: Should I ask `gdal-dev`? Or just strike it?

metasim:
Just realized I should add the setters too. Not enough coffee this morning...

src/raster/rasterband.rs
lnicola: We return `None`, right?

lnicola:
Good idea, I was actually wondering if there's any way to set it. Might make sense to use a larger value then, like 2^62.

metasim:
> Can you resample those to 1x1 and maybe merge them into a single file?

How about a 1x2 so we can have a data value and a no-data value?

CHANGES.md
```diff
@@ -1,6 +1,9 @@
 # Changes
 
 ## Unreleased
+- Added bindings for `GetNoDataValueAsInt64` and `GetNoDataValueAsUInt64`
```
lnicola, line 4: Let's spell out the method names here.

metasim:
> Good idea, I was actually wondering if there's any way to set it. Might make sense to use a larger value then, like 2^62.

I basically neglected to see these

* https://gdal.org/api/gdalrasterband_cpp.html#_CPPv4N14GDALRasterBand21SetNoDataValueAsInt64E7int64_t
* https://gdal.org/api/gdalrasterband_cpp.html#_CPPv4N14GDALRasterBand22SetNoDataValueAsUInt64E8uint64_t
*

metasim:
> Doesn't seem necessary to me...

What are you referring to here? The comment sequence is confusing.

lnicola:
I meant that about the second pixel with a valid value. I don't think it's needed to test these functions, but they could be useful in a simple warping/resampling test.

src/raster/rasterband.rs
```diff
@@ -713,6 +713,101 @@ impl<'a> RasterBand<'a> {
...
+    ///
+    /// If `no_data` is `None`, any existing no-data value is deleted.
+    ///
+    /// # Notes
+    /// See also:
+    /// [`GDALSetRasterNoDataValueAsInt64`](https://gdal.org/api/raster_c_api.html#_CPPv431GDALSetRasterNoDataValueAsInt6415GDALRasterBandH7int64_t),
+    /// [`GDALDeleteRasterNoDataValue`](https://gdal.org/api/raster_c_api.html#_CPPv427GDALDeleteRasterNoDataValue15GDALRasterBandH)
+    #[cfg(all(major_ge_3, minor_ge_5))]
+    pub fn set_no_data_value_i64(&mut self, no_data: Option<i64>) -> Result<()> {
+        let rv = if let Some(no_data) = no_data {
```
lnicola, line 799: Well, isn't [this](https://github.com/georust/gdal/pull/308#discussion_r964887637) awkward...

Taking an `Option` seemed like a good idea at the time...

src/raster/tests.rs
```diff
@@ -457,16 +458,57 @@ fn test_get_rasterband() {
...
+        let ds = driver.create_with_band_type::<i64, _>(&path, 1, 1, 1)?;
+        let mut rasterband = ds.rasterband(1)?;
+        assert_eq!(rasterband.no_data_value_i64(), None);
+        rasterband.set_no_data_value_i64(Some(255i64))?;
+    }
+
+    let ds = Dataset::open(&path)?;
+    let rasterband = ds.rasterband(1)?;
+
+    assert_eq!(rasterband.no_data_value_i64(), Some(255i64));
```
lnicola, line 489: Should we use something larger here? I don't care that much.

src/raster/rasterband.rs
metasim: I still think there's value in keeping them consistent...

The alternative is to add [GDALDeleteRasterNoDataValue](https://gdal.org/api/raster_c_api.html#_CPPv427GDALDeleteRasterNoDataValue15GDALRasterBandH)

lnicola: And now I think we should go for that (but in a separate PR).

src/raster/tests.rs
metasim: Sure, why not.

Are you ok with basically synthesizing the test data on the fly, rather than checking in files?

lnicola: Using MEM? Sure, that's a good idea. We're not testing GDAL here,.we just want to make sure we don't crash.

metasim: It's using `TempFixture`, which does actually write to disk, but the rational is the same. Just testing bindings.

lnicola (approved):

lnicola:
Please squash at the end!

PR #523, opened 2024-02-22
Add Dataset::rasterbands

### LINKED ISSUE #521: No iterator support for rasterbands
There is no way to use an iterator over the rasterbands of a dataset, the only way to access them is through either indexing using the `.rasterband()` method, or indirectly specifying a range and iterating over that.
For instance: `(a..b).into_iter().for_each(|idx| { ds.rasterband(idx); });`
Is there any way to do this directly? The dataset layers has a method `.layers()` that enables iterating over the layers of it. Anything similar for rasterbands would be helpful.

cc: @rkshthrmsh

lnicola:
---

Closes #521

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Added `Dataset::rasterbands`.
+  - <https://github.com/georust/gdal/pull/523>
+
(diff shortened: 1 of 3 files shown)


metasim (approved):
Nice!

PR #527, opened 2024-04-11
build: no bindings: make error more explicit

Frankkkkk:
Hi,

The panic that arises when there are no pre-build bindings available is not that explicit and may confuse some users into thinking there's something wrong with GDAL.

Thanks!

---

### DIFF
--- a/gdal-sys/build.rs
+++ b/gdal-sys/build.rs
@@ -247,7 +247,7 @@ fn main() {
                 version.major, version.minor
             ));
             if !binding_path.exists() {
-                panic!("No pre-built bindings available for GDAL version {}.{}. Use `--features bindgen` to generate your own bindings.", version.major, version.minor);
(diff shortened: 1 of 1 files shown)


lnicola:
Thanks! Maybe we should go for something simpler like "Enable the `bindgen` feature of `gdal` or `gdal-sys` to generate them during build."?

Frankkkkk:
Disclaimer: I know nothing about gdal-rs ;-)

I think it'd be great to distinguish GDAL from the rust crate. To be honest I thought the message was related to the gdal library. Maybe something like "Enable the `bindgen` feature of the `gdal` or `gdal-sys` crate to generate them during build."?

If that's okay with you I can change the MR.
Thanks!

lnicola:
I mean sure, but GDAL is GDAL and `gdal` is this crate :-).

PR #528, opened 2024-04-18
Add `Rasterband::fill`

netthier:
---
This PR adds safe bindings for https://docs.rs/gdal-sys/latest/gdal_sys/fn.GDALFillRaster.html (https://gdal.org/api/gdalrasterband_cpp.html#classGDALRasterBand_1a55bf20527df638dc48bf25e2ff26f353)
The C++ API has a default value for the `dfImaginaryValue` param which is missing from the C API, so I reintroduced via an `Option`+`.unwrap_or`. Let me know if this isn't acceptable.
I also haven't tested that parameter, as I'm unsure how to read it back.

### DIFF
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -815,6 +815,28 @@ impl<'a> RasterBand<'a> {
             Ok(())
         }
     }
+
+    /// Fill this band with a constant value.
+    ///
+    /// If `imaginary_value` is `None`, the imaginary component will be set to 0.
+    ///
+    /// # Notes
+    /// See also:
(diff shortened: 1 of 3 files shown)


src/raster/rasterband.rs
```diff
@@ -815,6 +815,16 @@ impl<'a> RasterBand<'a> {
             Ok(())
         }
     }
+
+    /// Fill this band with a constant value.
```
lnicola, line 819: Can you add a link to https://gdal.org/api/gdalrasterband_cpp.html#classGDALRasterBand_1a55bf20527df638dc48bf25e2ff26f353?

See the method above for an example.

lnicola (approved):

netthier:
I should probably add docs about the default imaginary value too...

netthier:
I ran `cargo fmt` against it, CI shouldn't fail anymore (I hope).

PR #529, opened 2024-04-19
Add `Rasterband::set_statistics`

netthier:
---
This PR adds safe bindings for https://docs.rs/gdal-sys/latest/gdal_sys/fn.GDALSetRasterStatistics.html (https://gdal.org/api/gdalrasterband_cpp.html#_CPPv4N14GDALRasterBand13SetStatisticsEdddd).


netthier:
It could maybe make sense to have the method take `StatisticsAll` instead.

src/raster/rasterband.rs
```diff
@@ -1033,6 +1033,26 @@ impl<'a> RasterBand<'a> {
...
+    ///
+    /// This method can be used to store min/max/mean/standard deviation statistics on a raster band.
+    ///
+    /// The default implementation stores them as metadata, and will only work on formats that can save arbitrary metadata.
+    /// This method cannot detect whether metadata will be properly saved and so may return `Ok(())` even if the statistics will never be saved.
+    ///
+    /// # Notes
+    /// See also:
+    /// [`GDALSetRasterStatistics`](https://gdal.org/api/gdalrasterband_cpp.html#_CPPv4N14GDALRasterBand13SetStatisticsEdddd)
+    pub fn set_statistics(&mut self, min: f64, max: f64, mean: f64, std_dev: f64) -> Result<()> {
```
ChristianBeilschmidt, line 1046: We could use `StatisticsAll` so that we can avoid having for floats being submitted. It is more verbose.

ChristianBeilschmidt (changes requested):
Looks good, maybe you can add your suggestion.

PR #536, opened 2024-05-06
Add Wrapper for GDALVectorTranslate(ogr2ogr)

zwishing:
---
This PR add Wrapper for GDALVectorTranslate which is the equivalent of the ogr2ogr (https://gdal.org/programs/ogr2ogr.html) utility(https://gdal.org/api/gdal_utils.html#_CPPv419GDALVectorTranslatePKc12GDALDatasetHiP12GDALDatasetHPK26GDALVectorTranslateOptionsPi).
Also, I changed MultiDimTranslateDestination enum to DatasetDestination


src/programs/destination.rs
```diff
@@ -0,0 +1,83 @@
+use std::ffi::CString;
+use std::mem::ManuallyDrop;
+use std::path::{Path, PathBuf};
+use crate::Dataset;
+use crate::errors::GdalError;
+use crate::utils::_path_to_c_string;
+
+/// Path or Dataset to store Vectors or Raster
+pub enum DatasetDestination {
```
lnicola, line 9: Not sure, but maybe `DestinationDataset` sounds a little better?

src/programs/destination.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    pub unsafe fn do_no_drop_dataset(&mut self) {
+        match self {
+            Self::Path(_) => {}
+            Self::Dataset { dataset: _, drop } => {
+                *drop = false;
+            }
+        }
+    }
+}
```
lnicola, line 83: Nit: missing trailing newline.

src/programs/vector/mod.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg(all(major_ge_3, minor_ge_1))]
+mod vector_translate;
```
lnicola, line 2: This is available since 2.1, not 3.1, isn't it?

src/programs/vector/vector_translate.rs
```diff
@@ -0,0 +1,150 @@
...
+use std::ptr::{null, null_mut};
+
+use crate::{
+    errors::*,
+    utils::{_last_null_pointer_err},
+    Dataset,
+};
+use gdal_sys::{GDALVectorTranslate, GDALVectorTranslateOptions, GDALVectorTranslateOptionsFree};
+use libc::c_char;
+use crate::programs::destination::DatasetDestination;
```
lnicola, line 12: This should be `pub use`, otherwise you can't call `vector_translate`.

src/programs/vector/vector_translate.rs
```diff
@@ -0,0 +1,150 @@
...
+/// [GDALVectorTranslateOptions]: https://gdal.org/api/gdal_utils.html#_CPPv426GDALVectorTranslateOptions
+pub struct VectorTranslateOptions{
+    c_options: *mut GDALVectorTranslateOptions
+}
+
+impl VectorTranslateOptions{
+    /// See [GDALVectorTranslateOptionsNew].
+    ///
+    /// [GDALVectorTranslateOptionsNew]: https://gdal.org/api/gdal_utils.html#_CPPv429GDALVectorTranslateOptionsNewPPcP35GDALVectorTranslateOptionsForBinary
+    pub fn new<S:Into<Vec<u8>>,I:IntoIterator<Item=S>>(args:I)->Result<Self>{
```
lnicola, line 25: Please run `cargo fmt`.

Zwishing: It‘s excellent advice to use `CslString `instead of `fn new<S:Into<Vec<u8>>,I:IntoIterator<Item=S>>(args:I)->Result<Self> `, and I will make some modifications to this section.

PR #540, opened 2024-05-21
Catch potential null pointers returned by `Dataset::spatial_ref`

weiznich:
This otherwise hits an assertion in gdal debug builds for the OSRClone method

---

See https://github.com/OSGeo/gdal/blob/master/ogr/ogrspatialreference.cpp#L1406 for the relevant assertion in gdal

### DIFF
--- a/src/dataset.rs
+++ b/src/dataset.rs
@@ -221,7 +221,17 @@ impl Dataset {
     #[cfg(major_ge_3)]
     /// Get the spatial reference system for this dataset.
     pub fn spatial_ref(&self) -> Result<SpatialRef> {
-        unsafe { SpatialRef::from_c_obj(gdal_sys::GDALGetSpatialRef(self.c_dataset)) }
+        unsafe {
+            let spatial_ref = gdal_sys::GDALGetSpatialRef(self.c_dataset);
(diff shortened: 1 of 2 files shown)


lnicola (approved):

lnicola:
Can you also add an `assert!` to `SpatialRef::from_c_obj`, since its argument must be a valid pointer?

src/dataset.rs
```diff
@@ -221,7 +221,21 @@ impl Dataset {
...
-        unsafe { SpatialRef::from_c_obj(gdal_sys::GDALGetSpatialRef(self.c_dataset)) }
+        unsafe {
+            let spatial_ref = gdal_sys::GDALGetSpatialRef(self.c_dataset);
+            if spatial_ref.is_null() {
+                Err(GdalError::NullPointer {
+                    method_name: "GDALGetSpatialRef",
+                    msg: "Unable to get a spatial reference".to_string(),
+                })
+            } else {
+                assert!(
```
lnicola, line 232: Not here (we clearly can't hit this one), in `from_c_obj`.

lnicola:
Thanks, will merge once GHA comes back up.

PR #548, opened 2024-08-01
Fix/cfg lint

weiznich:
---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,3 +44,20 @@ members = ["gdal-sys"]
 features = ["array"]
 # define attribute `docsrs` for feature badges
 rustdoc-args = ["--cfg", "docsrs"]
+
+[lints.rust.unexpected_cfgs]
+level = "warn"
+check-cfg = [
+    'cfg(major_is_3)',
+    'cfg(major_ge_3)',
+    'cfg(major_ge_4)',
+    'cfg(minor_ge_1)',
+    'cfg(minor_ge_2)',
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -44,3 +44,20 @@ members = ["gdal-sys"]
...
+    'cfg(major_ge_4)',
+    'cfg(minor_ge_1)',
+    'cfg(minor_ge_2)',
+    'cfg(minor_ge_3)',
+    'cfg(minor_ge_4)',
+    'cfg(minor_ge_5)',
+    'cfg(minor_ge_6)',
+    'cfg(minor_ge_7)',
+    'cfg(minor_ge_8)',
+    'cfg(minor_ge_9)',
```
lnicola, line 62: I wonder if it wouldn't be better to emit these from the build script. We'll have to add `minor_ge_10` soon.

Cargo.toml
```diff
@@ -44,3 +44,20 @@ members = ["gdal-sys"]
 features = ["array"]
 # define attribute `docsrs` for feature badges
 rustdoc-args = ["--cfg", "docsrs"]
+
+[lints.rust.unexpected_cfgs]
+level = "warn"
```
lnicola, line 49: Is this required if the lint is warn-by-default?

lnicola (approved):

lnicola:
Thanks, I've been meaning to take a look at this one.

PR #550, opened 2024-08-04
Fix raster buffer indexing

CyclotomicDev:
---
Changed to the correct indexing when converting from a row-major 2D array to vector index.
Previous test only really tested if `std::ops::Index` works, and not indexing the buffer; in the example the index of `b[(3,5)] = 3 * 5 + 5` would be the same as `b[(4,0)] = 4 * 5 + 0` . New test ensures all buffer indices covered.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@
 
 ## 0.17.1
 
+### Added
+ - Added `Buffer<T>::width` and `Buffer<T>::height` ([#550](https://github.com/georust/gdal/pull/550))
+
+### Fixed
  - Fixed incorrect bounds check in `Buffer<T>` indexing ([#550](https://github.com/georust/gdal/pull/550))
 
 ## 0.17
--- a/src/raster/buffer.rs
+++ b/src/raster/buffer.rs
@@ -91,8 +91,20 @@ impl<T: GdalType> Buffer<T> {
         (self.shape, self.data)
     }
 
+    /// Returns the width (number of columns) of the buffer.
+    #[doc(alias = "columns")]
+    pub fn width(&self) -> usize {
+        self.shape.0
+    }
+
+    /// Returns the height (number of rows) of the buffer.
+    #[doc(alias = "rows")]
+    pub fn height(&self) -> usize {
+        self.shape.1
+    }
+
     /// Gets the 2-d shape of the buffer.
-    //
+    ///
     /// Returns `(cols, rows)`
     ///
     /// # Notes


lnicola:
I think `Buffer` is supposed to be row-major, so the computation is correct. How did you come across this?

lnicola:
That's because band sizes are (width, height), but that doesn't mean the data is structured the same way. Compare to GDAL's [`RasterIO`](https://gdal.org/api/gdalrasterband_cpp.html#classGDALRasterBand_1a10d99d9e1f8b7c77fa48c0c648720070):

> The buffer into which the data should be read, or from which it should be written. This buffer must contain at least nBufXSize * nBufYSize words of type eBufType. It is organized in left to right, top to bottom pixel order. Spacing is controlled by the nPixelSpace, and nLineSpace parameters.

lnicola:
>  I don't understand why coord.0 * self.shape.0 + coord.1 is used as it is not the formula for dealing with row-major (or column-major) index representation, and the test ,index, would fail if some of the numbers were changed.

Because `coord` is `(y, x)` and `shape` is `(width, height)`, as documented on `fn shape()`.

lnicola:
> Currently it returns data at (y, x), but has bounds for (x, y),

Oh, I didn't spot the bounds check. Yeah, I think the formula is correct, but the bounds check is wrong.

lnicola:
CC @metasim my intention was to index by `(row, col)`, but I don't remember if that matches `ndarray` or other libraries.

jdroenner:
I think ndarray  can use both (https://docs.rs/ndarray/latest/ndarray/enum.Order.html) and there are good arguments for both options.  I prefer [y,x] which is what numpy (python) use. And if i remember correct this is also the default in ndarray.

lnicola:
@CyclotomicDev can you update this PR to switch around the bounds check?

lnicola:
I've dropped the dependency version change and added a warning to the docs:

image

metasim:
> CC @metasim my intention was to index by `(row, col)`, but I don't remember if that matches `ndarray` or other libraries.

There's a lot of historical legacy here, but my memory is fuzzy on why it is the way it is.  Every time I would iterate over cells, I'd double check the docs, for both correctness and cache coherence (an iterator can take care of that part).  

In my own personal libraries I never return tuples, but structs with labeled fields because it's too easy for me to mess up. But `gdal` doesn't do that (e.g. window coordinates to `read`), so I don't think there's any way around having to manually check the docs.

As you note, `ndarray` supports either (as does `numpy`) due to Fortran vs C traditions, so that part of the API could abstract this out.

lnicola:
I was asking more about what `buf[(y, x)]` should do, but good point, I've added a pair of getters to make things easier for the users.

jdroenner:
To have been thinking about the Buffer. To be honest i think we should deprecate all methods with tuple params as well as the shape method. We should simply add new methods with row and column as parameters. Who ever needs ndarray style access can transform the Buffer into an Ndarray...

lnicola:
@jdroenner we should probably discuss the deprecation of `Index` separately, but maybe let's publish a patch version now, to unblock the other PRs.

jdroenner:
good idea. some lint broke the CI. I guess we should fix that first.

lnicola:
I have an open PR for that.

jdroenner:
merged! thx!

lnicola:
Gentle nudge about publishing a patch, @jdroenner.

jdroenner:
i will do it later today or tomorrow

PR #564, opened 2024-09-29
Add script/fix_typos.sh to help fix typos

rouault:
---
Script which was used to help fix typos of https://github.com/georust/gdal/pull/563 (borrowed from the one of OSGeo/GDAL)

### DIFF
--- a/script/fix_typos.sh
+++ b/script/fix_typos.sh
@@ -1,0 +1,42 @@
+#!/bin/sh
+# -*- coding: utf-8 -*-
+
+set -eu
+
+SCRIPT_DIR=$(dirname "$0")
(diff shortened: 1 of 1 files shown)


lnicola:
Thanks! I've seen https://github.com/marketplace/actions/typos-action in action before, I will give that one a try later today and see how it compares. Probably about the same, I guess.

rouault:
> https://github.com/marketplace/actions/typos-action i

thanks for the hint!

lnicola:
I'm inclined to merge #565 instead of this. I think it finds everything in #563, plus an extra typo ("auto-dectect") and requires less configuration.

PR #569, opened 2024-10-06
upgrade ndarray to 0.16

nmandery:
This PR upgrades the ndarray dependency from 0.15 to 0.16

---

### DIFF
--- a/src/raster/buffer.rs
+++ b/src/raster/buffer.rs
@@ -220,7 +220,11 @@ impl<T: GdalType + Copy> From<Array2<T>> for Buffer<T> {
         let shape = value.shape();
         let (cols, rows) = (shape[1], shape[0]);
         let data: Vec<T> = if value.is_standard_layout() {
-            value.into_raw_vec()
+            let (data, offset) = value.into_raw_vec_and_offset();
+            match offset {
+                None | Some(0) => data,
+                Some(offset) => data[offset..].to_vec(),
+            }
         } else {
             value.iter().copied().collect()
         };
@@ -233,7 +237,7 @@ impl<T: GdalType + Copy> From<Array2<T>> for Buffer<T> {
 #[cfg(test)]
 mod tests {
     use crate::raster::Buffer;
-    use ndarray::{Array2, ShapeBuilder};
+    use ndarray::{arr2, s, Array2, ShapeBuilder};
 
     #[test]
     fn convert_to() {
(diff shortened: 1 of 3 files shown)


lnicola:
Thanks, but I don't think this is right (just as the previous code wasn't right). For reference, I saw this example in https://github.com/rust-ndarray/ndarray/pull/994:

```rust
let a = Array2::<f64>::zeros([3, 4]).slice_move(s![1.., ..]);
let b = Array2::<f64>::zeros([2, 4]);
assert_eq!(a.shape(), b.shape());
assert_eq!(a.strides(), b.strides());
assert_ne!(
    unsafe { a.as_ptr().offset_from(a.into_raw_vec().as_ptr()) },
    unsafe { b.as_ptr().offset_from(b.into_raw_vec().as_ptr()) },
);
```

Which implies that slicing sometimes avoids copying the data by bumping an offset from the start of the storage (the example strips off the first row).

So the `offset` is not always 0 (of course, otherwise it wouldn't be there), and we should handle it, either by copying the data to a new `Vec`, or by introducing a similar `offset` concept, which would probably be quite confusing to use.

nmandery:
I understood having a standard layout means also having an offset of 0, but I am afraid you are right. So we need to copy the data (like it is currently done for the non-standard-layout) in case of an offset.

I will add this to the PR later today.

lnicola:
I'm again wondering if we should fully switch to `ndarray` (see #570).

src/raster/buffer.rs
```diff
@@ -220,7 +220,12 @@ impl<T: GdalType + Copy> From<Array2<T>> for Buffer<T> {
         let shape = value.shape();
         let (cols, rows) = (shape[1], shape[0]);
         let data: Vec<T> = if value.is_standard_layout() {
-            value.into_raw_vec()
+            let (data, offset) = value.into_raw_vec_and_offset();
+            if let Some(offset) = offset {
+                data.into_iter().skip(offset).collect()
+            } else {
+                data
```
lnicola, line 227: Let's also handle `offset == 0` the same way.

nmandery: Done - I removed the if

src/raster/buffer.rs
```diff
@@ -220,7 +220,12 @@ impl<T: GdalType + Copy> From<Array2<T>> for Buffer<T> {
         let shape = value.shape();
         let (cols, rows) = (shape[1], shape[0]);
         let data: Vec<T> = if value.is_standard_layout() {
-            value.into_raw_vec()
+            let (data, offset) = value.into_raw_vec_and_offset();
+            if let Some(offset) = offset {
+                data.into_iter().skip(offset).collect()
```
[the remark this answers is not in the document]
nmandery: Yepp, done

src/raster/buffer.rs
lnicola: That might do an extra copy in the zero offset case.

lnicola:
@rouault looks like I've just won the lottery in https://github.com/georust/gdal/actions/runs/11372075639/job/31635584318:

> OGR2SQLITE_Setup() failed due to sqlite3_api == nullptr

PR #575, opened 2024-10-19
Relax libsqlite3-sys version requirements

lnicola:
---

Not sure if it's actually a good idea, but it will match https://github.com/georust/proj/blob/main/proj-sys/Cargo.toml.


lnicola:
Probably blocked on a `proj-sys` release.

ChristianBeilschmidt:
Is there a way to test the minimum versions in the CI? Max versions are checked by the current CI.

Nevertheless, this should be fine to merge.

lnicola:
There's a cargo flag to pick the minimum versions, but I don't think we call anything from that crate, so we'll be testing GDAL rather than our crate.

I think we have to wait until proj-sys gets a new release, because cargo seems to complain otherwise.

PR #579, opened 2024-10-23
Add an explicit include to gdal-src Cargo.toml

weiznich:
This commit adds an explicit include to the `Cargo.toml` file of gdal-src to minimize the amount of code uploaded to crates.io. This reduces the uncompressed package size from >100MB to ~50MB. This is still larger than what's allowed by crates.io by default, but I know that it is possible to get an exception from these default limits by just asking with a good reason. Source code from an existing established open source project counts as good reason, so that's just something that needs to be done before publishing the first version.

The last time I had this problem the workflow was as follows:

1. Publish an empty version
2. Ask the crates.io team to bump the limit, e.g. via zulip
3. Publish the real code.

---

### DIFF
--- a/gdal-src/Cargo.toml
+++ b/gdal-src/Cargo.toml
@@ -3,8 +3,21 @@ name = "gdal-src"
 version = "0.1.0"
 edition = "2021"
 links = "gdal_src"
+description ="Build script for compiling GDAL from source."
+license-file = "source/LICENSE.TXT"
+repository = "https://github.com/georust/gdal/"
+include = [
+    "/src/lib.rs",
+    "/build.rs",
+    "/Cargo.toml",
+    "/source/**/*.cpp",
+    "/source/**/*.h",
+    "/source/**/CMakeLists.txt",
+    "/source/LICENSE.TXT",
+]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
 
 [dependencies]
 link-cplusplus = "1.0"


lnicola:
Ouch, that's still a lot of MB, thanks! I wonder if we can ask for the size increase before publishing the crate (probably not?). I'll send an email right away to ask.

lnicola:
I can merge this, but it will be up to @jdroenner to publish the initial version.

lnicola:
Sure, let's add those.

jdroenner:
I can do the empty release later today.  Anything i need to be aware of?

weiznich:
> I can do the empty release later today. Anything i need to be aware of?

No, not really. It's just that the database entry must exist, so that they can just add an exception manually.

jdroenner:
> `cargo publish --dry-run` also reports that at least the following `Cargo.toml` fields need to be added:
> 
> ```
> warning: manifest has no description, license, license-file, documentation, homepage or repository.
> ```
> 
> I can push a commit for that as well if you want.

what should we fill in for homepage and documentation? A link to georust/gdal on github?

lnicola:
Same as the other crates, I guess?

jdroenner:
i created the empty crate: https://crates.io/crates/gdal-src

lnicola (approved):

lnicola:
Turns out this was much too optimistic :pensive:.

lnicola:
@jdroenner @weiznich did any of you write to the crates.io team? Should I send an email?

weiznich:
I did not write them yet. In my experience the fastest way to reach them is via Zulip: https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io

lnicola:
I've just sent an email, that's the preferred option.

lnicola:
Looks like we got the limit bumped to 50 MB compressed, which is hopefully enough (I think we're at 12 MB, even after #580).

PR #581, opened 2024-10-29
Take the field index, not name, in the feature getters and setters

lnicola:
---

Field access by name tends to be very slow (I've seen it take 30% of the time in some programs), but beginners are unlikely to realize that they're doing a number of expensive lookups per feature.

Summary of changes:

 - drop field getters that take a field name
 - change the field setters to take an index instead of a name (the index versions were missing)
 - make `field_count` return `usize` for consistency
 - add `Defn::field_index` and make `Feature::field_index` public
 - drop `LayerAccess::create_feature_fields` (it could be updated, but it's a bit weird)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,11 @@
   - Include OS-specific versions of the pre-built bindings ([#574](https://github.com/georust/gdal/pull/574))
   - Upgrade `ndarray` dependency to 0.16 ([#569](https://github.com/georust/gdal/pull/569))
   - Upgrade `thiserror` dependency to 2.0 ([#586](https://github.com/georust/gdal/pull/586))
+  - Update `Feature::field`, `Feature::set_field`, `Feature::set_field_string`, `Feature::set_field_string_list`, `Feature::set_field_double`, `Feature::set_field_double_list`, `Feature::set_field_integer`, `Feature::set_field_integer_list`, `Feature::set_field_integer64`, `Feature::set_field_integer64_list`, `Feature::set_field_datetime`, `Feature::set_field_null`, `Feature::unset_field` to take a field index, not a name ([#581](https://github.com/georust/gdal/pull/581))
+  - Drop `Feature::field_as_integer_by_name`, `Feature::field_as_integer64_by_name`, `Feature::field_as_double_by_name`, `Feature::field_as_string_by_name`, `Feature::field_as_datetime_by_name` ([#581](https://github.com/georust/gdal/pull/581))
+  - Update `Feature::field_count` to return `usize` instead of `i32` ([#581](https://github.com/georust/gdal/pull/581))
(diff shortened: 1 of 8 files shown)


lnicola:
r? @metasim, @ChristianBeilschmidt, @jdroenner since this is a relatively big change
r? @latot since you said you wanted to get involved

Hopefully I'll be able to get rid of this thing soon:

image

(it goes on and on)

latot:
I think this is nice, some questions... can we actually change the fields in any way that could broke the field id?

From where is the image? I can see some transformations from ```isize``` to ```i32``` with ```as```.

lnicola:
> can we actually change the fields in any way that could broke the field id?

You can delete fields from the `Defn` of a layer, but it won't reflect in the fields of the features, and I don't think we even expose that. 

> From where is the image? I can see some transformations from isize to i32 with as.

From `usize`, maybe, but most of those are checked (I think layer iteration is the exception). GDAL fields are 32-bit and positive (-1 means a missing field), but Rust uses probably expect `usize`. We do a similar thing for raster dimensions. But it's a small change and can be reverted easily (either now or in the future).

Or we could go the other way around and introduce a `FieldIdx` type that wraps an `i32`.

lnicola:
Gentle nudge @jdroenner :sweat_smile:.

latot:
I thinking the next thing.

Actually you said there is no function exposed that would be able to change the columns (remove/add), at some point I think would be good to have some tests that checks the actual idx do not change in any of this operations, just to be sure this is going to work fine in the future too.

No idea if rn is the best moment.

lnicola:
@latot if we don't expose `OGR_L_DeleteField`, how could we test that it doesn't affect the other indexes? Note that the GDAL docs explicitly say:

> This function should not be called while there are feature objects in existence that were obtained or created with the previous layer definition.

Keep in mind that the intended usage is something like:

```rust
let id_idx = layer.defn().field_index("id")?;
let name_idx = layer.defn().field_index("name")?;
for feature in layer.features() {
    let id = feature.field_as_integer(id_idx)?.expect("id is set");
    let name = feature.field_as_string(id_idx)?.expect("name is set");
    println!("Feature {id}, name: {name}");
}
```

latot:
Yes, the would be the usage, is more like a note to when something that could affect the id is exposed to the user, also, the only side I'm seeing, some ppl, could try to call that function directly even if is not exposed, would be the last concern from me.

lnicola:
We can't prevent every bug. You can always pass in the wrong field index or name, or even delete the whole layer. But if you do, you'll notice pretty quickly once you run your program.

What's harder to notice is that your program is running twice as slow because you've used the wrong field access methods.

lnicola:
Heads-up: I'll probably merge this tonight if there's no more opposition.

PR #583, opened 2024-11-01
Add delete layer method and tests in doc

Atreyagaurav:
---

I noticed this was not present, so I just added a crude implementation. Might be nice to add `delete_layer_by_name` option as well.

If there are more unimplemented methods, we can add as well.

### DIFF
--- a/src/vector/layer.rs
+++ b/src/vector/layer.rs
@@ -728,6 +728,30 @@ impl Dataset {
         };
         Ok(self.child_layer(c_layer))
     }
+
+    /// Deletes the layer at given index
+    ///
+    /// ```
+    /// # use gdal::DriverManager;
+    /// # let driver = DriverManager::get_driver_by_name("GPKG").unwrap();
+    /// # let mut dataset = driver.create_vector_only("/vsimem/example.gpkg").unwrap();
+    /// let blank_layer = dataset.create_layer(Default::default()).unwrap();
+    /// assert!(dataset.delete_layer(1).is_err());
+    /// dataset.delete_layer(0).unwrap();
+    /// assert_eq!(dataset.layers().count(), 0);
+    /// ```
+    pub fn delete_layer(&mut self, idx: usize) -> Result<()> {
+        let idx = c_int::try_from(idx)?;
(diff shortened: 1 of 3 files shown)


lnicola (approved):

Atreyagaurav:
I don't think the CI fail is from my changes. Any idea what to do?

Atreyagaurav:
The CI error says this: 
     --- stderr
      thread 'main' panicked at gdal-sys/build.rs:250:17:
      No pre-built bindings available for GDAL version 3.11. Enable the `bindgen` feature of the `gdal` or `gdal-sys` crate to generate them during build.

lnicola:
You can just comment out https://github.com/georust/gdal/blob/master/.github/workflows/ci.yml#L28. It looks like `-latest` is now 3.11.0, but 3.10 hasn't been published yet :unamused:.

Atreyagaurav:
IDK why, commenting the line got me `Invalid workflow file` error, and CI didn't run. I tried a few alternative way to comment, thinking indent might have done it. I don't understand it, so for now I just removed the `-latest` line, Please remember to add it and `3.10.0` later.

Atreyagaurav:
I squashed all commits, so it's easier to merge. This should be done now (once CI passes).

src/vector/layer.rs
```diff
@@ -728,6 +728,30 @@ impl Dataset {
...
+    /// # let mut dataset = driver.create_vector_only("/vsimem/example.gpkg").unwrap();
+    /// let blank_layer = dataset.create_layer(Default::default()).unwrap();
+    /// assert!(dataset.delete_layer(1).is_err());
+    /// dataset.delete_layer(0).unwrap();
+    /// assert_eq!(dataset.layers().count(), 0);
+    /// ```
+    pub fn delete_layer(&mut self, idx: usize) -> Result<()> {
+        let idx = c_int::try_from(idx)?;
+        let err = unsafe { gdal_sys::GDALDatasetDeleteLayer(self.c_dataset(), idx) };
+        if err != gdal_sys::OGRErr::OGRERR_NONE {
```
lnicola, line 746:
```suggestion
        if err != OGRErr::OGRERR_NONE {
```

lnicola:
I'll merge this after you squash.

PR #584, opened 2024-11-02
Add MaybeTransaction for better writing speed when supported

### LINKED ISSUE #582: [Feature Req] Optionally use Transaction if supported
Would it be possible to use transaction if supported, else use normal dataset.

Considering the output dataset being able to use any driver user wants, I want to speed up the writing process when I can. If the output is to `gpkg` file I'd like to use transaction, if it's to `shp` then the code I have will fail. So I'm hoping for something that can work on both cases.

The Idea I have is, considering the code for `Transaction::commit` uses the internal `Dataset`, we can probably add the `commit` method on the `Dataset` and do something like this:
```rust
let trans = d.start_transaction().is_ok();
// do stuffs with d: Dataset
if trans {
    d.commit()?;
}
```

This just needs the `commit` method added to `Dataset` which makes sense: 

https://github.com/georust/gdal/blob/5d886abec30bb1666750d9f20b9860b361e39a5a/src/vector/transaction.rs#L48-L58

Currently, You have to do something like this:
```rust
let mut trans = false;
if let Some(txn) = d.start_transaction() {
    // do stuffs with txn 
    trans = true;
}

if !trans {
    // do stuffs with d
}
```

You can't use `if let ... {} else {}` pattern, as you can't borrow `d` mutably on `start_transaction()` and on the `else` block at the same time. And due to rust's type system, you can't replace `d: Dataset` with `txn: Transaction` to do the same thing (although you can with deref). 

But basically to write a code that works with drivers that support transaction and drivers that don't, I think the first pattern I suggested feels like the best way to do it, as the code is internally calling `commit` on the `Dataset` anyway.

Atreyagaurav:
---

Fixes #582


src/vector/transaction.rs
```diff
@@ -179,9 +179,93 @@ impl Dataset {
...
+
+impl<'a> DerefMut for MaybeTransaction<'a> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.dataset
+    }
+}
+
+impl Dataset {
+    pub fn maybe_start_transaction(&mut self) -> MaybeTransaction<'_> {
+        let force = 1;
```
lnicola, line 249: We should add at some point that parameter for `force`, but if you want this for better performance, you absolutely don't want to enable it.

I think it only works for FileGDB, and it's implemented by making a file system-level copy of the whole dataset, and restoring it on rollback.

Atreyagaurav: Yeah that won't help with performance, I think if we want transaction for rollback purposes vs want transaction for speed purposes, I think we should have different methods.

The `MaybeTransaction` doesn't have `rollback` method at all, that'd help at compile time to let people know they can't rollback. Since it could be working on the database directly without transaction as well.

That's why I was thinking maybe some other name would be better. The current interpretation of `MaybeTransaction` is it is maybe a transaction or not, hence it has to be committed on drop, and no rollback. To make the Transaction and non transaction database do same thing.

On that note I think we can add the force thing to normal `Transaction` or add another `ForceTransaction` that does filesystem copy ourselves when Transaction fails.

src/vector/transaction.rs
```diff
@@ -179,9 +179,93 @@ impl Dataset {
...
+}
+
+impl<'a> DerefMut for MaybeTransaction<'a> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.dataset
+    }
+}
+
+impl Dataset {
+    pub fn maybe_start_transaction(&mut self) -> MaybeTransaction<'_> {
```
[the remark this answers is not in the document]
Atreyagaurav: Sorry for late response. I made it quickly, just copying the mechanism that is being already used by the transaction. It gives uniformity to the transaction API.

lnicola:
I can see why the callback approach is better: a `commit` method doesn't really make sense, because the transaction will get committed anyway (we could offer an API where `rollback` can fail and still commit, but that seems harder to understand), and we can't expose. On the other hand, the callback API is much harder to use:

```rust
// Nb.: untested, and I'm probably misremembering the API

let mut tx = ds.maybe_transaction()?;
let mut batch_len = 0; // slice::chunks would probably be even nicer
for row in df {
    let feature = write_feature(&tx, row)?;
    batch_len += 1;
    if batch_len == BATCH_SIZE {
        tx.commit()?; // just to check for errors
        tx = ds.maybe_transaction()?;
        batch_len = 0;
    }
}

if batch_len > 0 {
    tx.commit()?;
}
```

For me, at least, this feels pretty natural to write. With a callback, I think it would look like this:

```rust
// Nb.: even more wrong, probably

let mut it = df.into_iter().peekable();
while it.peek().is_some() {
    ds.maybe_transaction(|| {
        for row in it.take(BATCH_LIMIT) {
            write_feature(&tx, row)?;
        }
        Ok(())
    })?;
}
```

Which isn't _that_ bad, but seems more tricky to write.

And if we use a callback approach here, we should probably change `start_transaction` for symmetry. @weiznich I'm pretty sure you've spent a lot of time on this API question, you might have more arguments in favor of the callback approach.

---

Then there's the question of naming, `MaybeTransaction` isn't that bad, but feels a bit unserious. I don't have a better alternative, though.

Atreyagaurav:
I haven't thought that much into it. But I think, looking at your examples, the callback does make sense. Although, I don't know if there is going to be any problem with the borrow checker for that example. But making it take a closure and then calling that either on transaction or dataset seems like a good idea, but yeah it'll be harder to write than running things on a `MaybeTransaction`.

What do you think about this implementation?
```diff
@@ -177,6 +177,22 @@ impl Dataset {
         }
         Ok(Transaction::new(self))
     }
+
+    pub fn maybe_transaction(&mut self, func: impl Fn(&Dataset) -> Result<()>) -> Result<()> {
+	let force = 1;
+	let rv = unsafe { gdal_sys::GDALDatasetStartTransaction(self.c_dataset(), force) };
+	let res = func(self);
+        if rv == OGRErr::OGRERR_NONE {
+	    let rv = unsafe { gdal_sys::GDALDatasetCommitTransaction(self.c_dataset()) };
+            if rv != OGRErr::OGRERR_NONE {
+		return Err(GdalError::OgrError {
+                    err: rv,
+                    method_name: "GDALDatasetCommitTransaction",
+		});
+            }
+	}
+	res
+    }
 }
```

Atreyagaurav:
@lnicola  what do you think of this:
https://github.com/Atreyagaurav/gdal/commit/87bfc391db1650616247e336af2bfa72554eb75c

It is the above implementation with tests, I think this is good.

lnicola:
I think it's fine. Feel free to force-push to this branch.

Atreyagaurav:
I thought renaming the branches would be a fun way to work around the branch swap, but it closed it.

Please refer to #591  for the new implementation based on closure.

PR #585, opened 2024-11-03
Add `Dataset::has_capability`

Atreyagaurav:
---

This is for doing things like `Dataset::has_capability(DatasetCaps::ODsCTransactions)` to check if dataset supports Transaction. 

I added the Capabilities from:
- https://gdal.org/en/latest/doxygen/ogr__api_8h.html#a9010219bbc2e32627064ed860048d979
- https://gdal.org/en/latest/doxygen/gdal_8h.html#a8dc635a536f123e1b03004dc74b4da86

I followed the example of `Layer::has_capability` for both of them.

This will help use the `if {} else {}` block for capability based logic like the one mentioned in #582 

Basically it'll make it so that @lnicola 's Suggestion given below can work:
```rust
let stuffs = |d: &Dataset| -> Result<()> { ... };
if /* transactions are supported */ {
    let txn = d.start_transaction()?;
    stuffs(&txn)?;
    txn.commit()?;
} else {
    stuffs(&d)?;
}
```

Currently, without this function, we have to wait for the result of `Dataset::start_transaction` to know if it supports it, in which case we cannot use `if () {} else {}` block due to Rust's borrow rules as the `Dataset::start_transaction` will take `&mut Dataset`.

### DIFF
--- a/src/dataset.rs
+++ b/src/dataset.rs
@@ -1,4 +1,8 @@
-use std::{ffi::CString, ffi::NulError, path::Path, ptr};
+use std::{
+    ffi::{CStr, CString, NulError},
+    path::Path,
+    ptr,
+};
 
 use gdal_sys::{CPLErr, GDALDatasetH, GDALMajorObjectH};
 
@@ -10,6 +14,29 @@ use crate::utils::{_last_cpl_err, _last_null_pointer_err, _path_to_c_string, _st
 use crate::{
     gdal_major_object::MajorObject, spatial_ref::SpatialRef, Driver, GeoTransform, Metadata,
 };
+
+pub struct DatasetCapability(&'static CStr);
+
+/// Dataset capabilities
+impl DatasetCapability {
+    /// Dataset can create new layers.
+    pub const CREATE_LAYER: DatasetCapability = DatasetCapability(c"CreateLayer");
+    /// Dataset can delete existing layers.
+    pub const DELETE_LAYER: DatasetCapability = DatasetCapability(c"DeleteLayer");
+    /// Layers of this datasource support CreateGeomField() just after layer creation.
+    pub const CREATE_GEOM_FIELD_AFTER_CREATE_LAYER: DatasetCapability =
+        DatasetCapability(c"CreateGeomFieldAfterCreateLayer");
+    /// Dataset supports curve geometries.
+    pub const CURVE_GEOMETRIES: DatasetCapability = DatasetCapability(c"CurveGeometries");
+    /// Dataset supports (efficient) transactions.
+    pub const TRANSACTIONS: DatasetCapability = DatasetCapability(c"Transactions");
+    /// Dataset supports transactions through emulation.
+    pub const EMULATED_TRANSACTIONS: DatasetCapability = DatasetCapability(c"EmulatedTransactions");
+    /// Dataset has a dedicated GetNextFeature() implementation, potentially returning features from layers in a non sequential way.
+    pub const RANDOM_LAYER_READ: DatasetCapability = DatasetCapability(c"RandomLayerRead");
+    /// Dataset supports calling CreateFeature() on layers in a non sequential way.
+    pub const RANDOM_LAYER_WRITE: DatasetCapability = DatasetCapability(c"RandomLayerWrite");
+}
 
 /// Wrapper around a [`GDALDataset`][GDALDataset] object.
 ///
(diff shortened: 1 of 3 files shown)


Atreyagaurav:
About the Enum variants, since `clippy` warns about the Prefix, should we just remove them? Since these are rust enum, I think we don't need the prefixes like `ODrC`, `ODsC`, `OLC` etc. on the enum itself as the Enum names (`LayerCaps`, `DriverCaps`, `DatasetCaps` etc.) are enough to distinguish them unlike in C.

I just followed what `LayerCaps` was doing for the other two. Changing that would be breaking change, but at least for the new ones we could not add Prefixes.

src/driver.rs
```diff
@@ -301,6 +321,13 @@ impl Driver {
     pub unsafe fn c_driver(&self) -> GDALDriverH {
         self.c_driver
     }
+
+    pub fn has_capability(&self, capability: DriverCaps) -> bool {
+        unsafe {
+            gdal_sys::OGR_Dr_TestCapability(self.c_driver(), capability.into_cstring().as_ptr())
```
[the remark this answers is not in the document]
Atreyagaurav: Maybe I should remove `Driver::has_capability` in that case. And we can add the new one later. Or accept the legacy API for now. But considering there are only two capability to check, I think it is fine to not have this method for now.

What do you think?

Atreyagaurav:
I'm waiting on whether to remove `Driver::has_capability` since it is based on legacy API, or rewrite it to use the `DCAP` metadata instead. Or maybe just adding a metadata checking function (if it doesn't exist already) is sufficient.

lnicola:
I think we can do something like below, what do you say?

```rust
struct DatasetCapability(&'static CStr); // we could also support non-'static ones, but probably not worth it

impl DatasetCapability {
    pub const CREATE_LAYER: DatasetCapability = DatasetCapability(c"CreateLayer");
    pub const DELETE_LAYER: DatasetCapability = DatasetCapability(c"DeleteLayer");
}

struct Dataset;
impl Dataset {
    pub fn test_capability(&self, cap: DatasetCapability) -> bool {
        _ = cap.0;

        // this is the best dataset
        true
    }
}
```

This is a little shorter, while still allowing custom caps if the user ever needs one.

lnicola:
> The DatasetCapability is not deprecated, and there are a fixed list on gdal side so I think it's fine.

My version has much less code, runs slightly faster, and can still be extended if GDAL adds a new capability, so I still think it's an improvement.

> Thank you, can you merge with failed tests? or should I do something?

I disabled the test, you can rebase.

lnicola:
That wasn't a clean rebase or squash but.. whatever, I guess. Anyway, please capitalize `datasource` in the doc-comments, and add a `.` at the end, they're a bit hard to read right now.

Afterwards I think we can merge this and use it in #591?

Atreyagaurav:
Sorry, I was struggling with rebase since the Valgrind was in the middle of two commits I made. It should be clean one, I undid everything and made a new commit on top.

lnicola (approved):

lnicola:
Yeah, it's clean now, thanks!

Atreyagaurav:
> Afterwards I think we can merge this and use it in https://github.com/georust/gdal/pull/591?

About that, this one is so that we can manually do the "if supports transaction" part by ourselves.

I think the #591 is a bit different because even if starting transaction fails we want to continue writing to database as the transaction is optional and for speedup process only.

src/dataset.rs
```diff
@@ -1,4 +1,9 @@
-use std::{ffi::CString, ffi::NulError, path::Path, ptr};
+use std::{
+    ffi::NulError,
+    ffi::{CStr, CString},
```
lnicola, line 3: Nit: let's merge these:

```suggestion
    ffi::{CStr, CString, NulError},
```

and add a changelog entry.

src/dataset.rs
```diff
@@ -327,6 +354,10 @@ impl Dataset {
         }
         Ok(transformation)
     }
+
+    pub fn has_capability(&self, capability: DatasetCapability) -> bool {
+        unsafe { gdal_sys::GDALDatasetTestCapability(self.c_dataset(), capability.0.as_ptr()) == 1 }
```
lnicola, line 359: Well, I do hope that `TRUE` is 1 and not `0xFFFF_FFFF` :sweat_smile:.

Atreyagaurav: That's what test cases are for. It's working for now.

PR #587, opened 2024-11-07
Replace OGR_DS_xxxx() by GDALDatasetxxxx()

rouault:
The OGR_DS_ API is legacy since GDAL 2.0

---


lnicola (approved):
Thanks

PR #589, opened 2024-11-27
Make string conversion fallible

### LINKED ISSUE #588: Unsoundness in "_string" and "_pathbuf"
**Describe the bug**
The _string and _pathbuf functions use CStr::from_ptr to convert a raw pointer (*const c_char) into a CStr. However, the functions do not validate that the pointer meets the necessary safety requirements. This may lead to Undefined Behavior (UB) if the pointer is invalid, null, or if the data it references is not null-terminated.
https://github.com/georust/gdal/blob/2c673d6a885d1299b8c7ce8cb0818e7e6dac43e6/src/utils.rs#L8
https://github.com/georust/gdal/blob/2c673d6a885d1299b8c7ce8cb0818e7e6dac43e6/src/utils.rs#L25
```
pub fn _string(raw_ptr: *const c_char) -> String {
    let c_str = unsafe { CStr::from_ptr(raw_ptr) };
    c_str.to_string_lossy().into_owned()
}
pub fn _pathbuf(raw_ptr: *const c_char) -> PathBuf {
    let c_str = unsafe { CStr::from_ptr(raw_ptr) };
    c_str.to_string_lossy().into_owned().into()
}
```
**To Reproduce**
Provide a null pointer:
```
let null_ptr: *const c_char = std::ptr::null();
let result = _string(null_ptr); // UB: Null pointer
```
**Expected behavior**
The functions should validate the input pointer to ensure that:

It is not null.
It points to valid memory.
It references a null-terminated C string.
The functions should gracefully handle invalid inputs by returning an appropriate error instead of invoking Undefined Behavior.
**Additional context**
Expected behavior includes:

Validating the raw pointer before dereferencing it.
Ensuring the memory it points to is a valid, null-terminated C string.

lnicola:
---

Closes #588

This makes `_string` and `_pathbuf` fallible and updates the callers to some sensible defaults (by returning empty strings in many places, which seems similar in spirit to what GDAL does). There's an API change on `SpatialRef::authority_name` and some friends, which used to return `Result`, but where `Option` feels more appropriate for what they're doing. It might be a little annoying for the users, though.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -55,7 +55,7 @@ pub fn get_config_option(key: &str, default: &str) -> Result<String> {
     let c_key = CString::new(key.as_bytes())?;
     let c_default = CString::new(default.as_bytes())?;
     let rv = unsafe { gdal_sys::CPLGetConfigOption(c_key.as_ptr(), c_default.as_ptr()) };
-    Ok(_string(rv))
+    Ok(_string(rv).unwrap_or_else(|| default.to_string()))
 }
 
 /// Clear the value of a GDAL library configuration option
@@ -96,7 +96,7 @@ pub fn get_thread_local_config_option(key: &str, default: &str) -> Result<String
     let c_key = CString::new(key.as_bytes())?;
     let c_default = CString::new(default.as_bytes())?;
     let rv = unsafe { gdal_sys::CPLGetThreadLocalConfigOption(c_key.as_ptr(), c_default.as_ptr()) };
-    Ok(_string(rv))
+    Ok(_string(rv).unwrap_or_else(|| default.to_string()))
 }
 
 /// Clear the value of a GDAL library configuration option
@@ -141,7 +141,7 @@ where
         error_num: CPLErrorNum,
         error_msg_ptr: *const c_char,
     ) {
-        let error_msg = _string(error_msg_ptr);
+        let error_msg = _string(error_msg_ptr).unwrap_or_default();
         let error_type: CplErrType = error_type.into();
 
         // reconstruct callback from user data pointer
--- a/src/cpl.rs
+++ b/src/cpl.rs
@@ -6,7 +6,6 @@
 use std::ffi::{c_char, c_int, CString};
 use std::fmt::{Debug, Display, Formatter};
 use std::mem::ManuallyDrop;
-use std::ops::Deref;
 use std::ptr;
 use std::str::FromStr;
 
@@ -178,11 +177,7 @@ impl CslStringList {
         // we know already `name` will never exist in a valid CslStringList.
         let key = CString::new(name).ok()?;
         let c_value = unsafe { CSLFetchNameValue(self.as_ptr(), key.as_ptr()) };
-        if c_value.is_null() {
-            None
-        } else {
-            Some(_string(c_value))
-        }
+        _string(c_value)
     }
 
     /// Perform a case <u>insensitive</u> search for the given string
@@ -240,20 +235,13 @@ impl CslStringList {
         if index > c_int::MAX as usize {
             return None;
         }
+
         // In the C++ implementation, an index-out-of-bounds returns an empty string, not an error.
         // We don't want to check against `len` because that scans the list.
         // See: https://github.com/OSGeo/gdal/blob/fada29feb681e97f0fc4e8861e07f86b16855681/port/cpl_string.cpp#L181-L182
         let field = unsafe { CSLGetField(self.as_ptr(), index as c_int) };
-        if field.is_null() {
-            return None;
-        }
 
-        let field = _string(field);
-        if field.is_empty() {
-            None
-        } else {
-            Some(field.deref().into())
-        }
+        _string(field).filter(|s| !s.is_empty()).map(|s| s.into())
     }
 
     /// Determine the number of entries in the list.
@@ -505,12 +493,7 @@ impl Iterator for CslStringListIterator<'_> {
             slice[self.idx]
         };
         self.idx += 1;
-        if field.is_null() {
-            None
-        } else {
-            let entry = _string(field);
-            Some(entry.deref().into())
-        }
+        _string(field).map(|s| s.into())
     }
 }
 
--- a/src/dataset.rs
+++ b/src/dataset.rs
@@ -208,7 +208,7 @@ impl Dataset {
     /// Fetch the projection definition string for this dataset.
     pub fn projection(&self) -> String {
         let rv = unsafe { gdal_sys::GDALGetProjectionRef(self.c_dataset) };
-        _string(rv)
+        _string(rv).unwrap_or_default()
     }
 
     /// Set the projection reference string for this dataset.
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -66,7 +66,7 @@ impl Driver {
     /// See also: [`long_name`](Self::long_name).
     pub fn short_name(&self) -> String {
         let rv = unsafe { gdal_sys::GDALGetDriverShortName(self.c_driver) };
-        _string(rv)
+        _string(rv).unwrap_or_default()
     }
 
     /// Return the short name of a driver.
@@ -76,7 +76,7 @@ impl Driver {
     /// See also: [`short_name`](Self::short_name`).
     pub fn long_name(&self) -> String {
         let rv = unsafe { gdal_sys::GDALGetDriverLongName(self.c_driver) };
-        _string(rv)
+        _string(rv).unwrap_or_default()
     }
 
     /// Create a new dataset of size (`size_x`, `size_y`) and `bands` band count,
--- a/src/gcp.rs
+++ b/src/gcp.rs
@@ -81,8 +81,10 @@ impl GcpRef<'_> {
 impl From<&gdal_sys::GDAL_GCP> for Gcp {
     fn from(gcp: &gdal_sys::GDAL_GCP) -> Self {
         Gcp {
-            id: _string(gcp.pszId),
-            info: _string(gcp.pszId),
+            // This seems to never be NULL
+            id: _string(gcp.pszId).unwrap_or_default(),
+            // GDAL docs state that `pszInfo` is filled in or `""`
+            info: _string(gcp.pszInfo).unwrap_or_default(),
             pixel: gcp.dfGCPPixel,
             line: gcp.dfGCPLine,
             x: gcp.dfGCPX,
@@ -133,10 +135,7 @@ impl Dataset {
     ///  See: [`GDALGetGCPProjection`](https://gdal.org/api/raster_c_api.html#gdal_8h_1a85ffa184d3ecb7c0a59a66096b22b2ec)
     pub fn gcp_projection(&self) -> Option<String> {
         let cc_ptr = unsafe { gdal_sys::GDALGetGCPProjection(self.c_dataset()) };
-        if cc_ptr.is_null() {
-            return None;
-        }
-        Some(_string(cc_ptr))
+        _string(cc_ptr)
     }
 
     /// Fetch GCPs.
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -59,10 +59,7 @@ pub trait Metadata: MajorObject {
     /// ```
     fn description(&self) -> Result<String> {
         let c_res = unsafe { gdal_sys::GDALGetDescription(self.gdal_object_ptr()) };
-        if c_res.is_null() {
-            return Err(_last_null_pointer_err("GDALGetDescription"));
-        }
-        Ok(_string(c_res))
+        _string(c_res).ok_or_else(|| _last_null_pointer_err("GDALGetDescription"))
     }
 
     /// Metadata in GDAL is partitioned into namespaces, knows as "domains" in the
@@ -150,9 +147,7 @@ pub trait Metadata: MajorObject {
                         c_domain.as_ptr(),
                     )
                 };
-                if !c_res.is_null() {
-                    return Some(_string(c_res));
-                }
+                return _string(c_res);
             }
         }
         None
--- a/src/raster/mdarray.rs
+++ b/src/raster/mdarray.rs
@@ -331,16 +331,11 @@ impl<'a> MDArray<'a> {
 
             let strings = string_pointers
                 .into_iter()
-                .map(|string_ptr| {
-                    if string_ptr.is_null() {
-                        String::new()
-                    } else {
-                        let string = _string(string_ptr);
-
-                        VSIFree(string_ptr as *mut c_void);
-
-                        string
-                    }
+                .filter_map(|ptr| {
+                    let s = _string(ptr);
+                    // GDAL allows `VSIFree(nullptr)`
+                    VSIFree(ptr as *mut c_void);
+                    s
                 })
                 .collect();
 
@@ -376,9 +371,8 @@ impl<'a> MDArray<'a> {
     pub fn unit(&self) -> String {
         unsafe {
             // should not be freed
-            let c_unit = GDALMDArrayGetUnit(self.c_mdarray);
-
-            _string(c_unit)
+            let c_ptr = GDALMDArrayGetUnit(self.c_mdarray);
+            _string(c_ptr).unwrap_or_default()
         }
     }
 
@@ -488,7 +482,8 @@ impl<'a> Group<'a> {
     }
 
     pub fn name(&self) -> String {
-        _string(unsafe { GDALGroupGetName(self.c_group) })
+        let c_ptr = unsafe { GDALGroupGetName(self.c_group) };
+        _string(c_ptr).unwrap_or_default()
     }
 
     pub fn group_names(&self, options: CslStringList) -> Vec<String> {
@@ -624,7 +619,8 @@ impl<'a> Dimension<'a> {
     }
 
     pub fn name(&self) -> String {
-        _string(unsafe { GDALDimensionGetName(self.c_dimension) })
+        let c_ptr = unsafe { GDALDimensionGetName(self.c_dimension) };
+        _string(c_ptr).unwrap_or_default()
     }
 
     pub fn indexing_variable(&self) -> MDArray {
@@ -712,7 +708,8 @@ impl ExtendedDataType {
     }
 
     pub fn name(&self) -> String {
-        _string(unsafe { GDALExtendedDataTypeGetName(self.c_data_type) })
+        let c_ptr = unsafe { GDALExtendedDataTypeGetName(self.c_data_type) };
+        _string(c_ptr).unwrap_or_default()
     }
 }
 
@@ -767,12 +764,9 @@ impl Attribute {
     }
 
     pub fn read_as_string(&self) -> String {
-        unsafe {
-            // SAFETY: should no be freed
-            let c_string = GDALAttributeReadAsString(self.c_attribute);
-
-            _string(c_string)
-        }
+        // should not be freed
+        let c_ptr = unsafe { GDALAttributeReadAsString(self.c_attribute) };
+        _string(c_ptr).unwrap_or_default()
     }
 
     pub fn read_as_string_array(&self) -> Vec<String> {
--- a/src/raster/rasterband.rs
+++ b/src/raster/rasterband.rs
@@ -950,13 +950,8 @@ impl<'a> RasterBand<'a> {
     /// Return the unit of the rasterband.
     /// If there is no unit, the empty string is returned.
     pub fn unit(&self) -> String {
-        let str_ptr = unsafe { gdal_sys::GDALGetRasterUnitType(self.c_rasterband) };
-
-        if str_ptr.is_null() {
-            return String::new();
-        }
-
-        _string(str_ptr)
+        let c_ptr = unsafe { gdal_sys::GDALGetRasterUnitType(self.c_rasterband) };
+        _string(c_ptr).unwrap_or_default()
     }
 
     /// Read the band mask flags for a GDAL `RasterBand`.
@@ -1362,8 +1357,8 @@ impl ColorInterpretation {
 
     /// Returns the name of this color interpretation.
     pub fn name(&self) -> String {
-        let rv = unsafe { gdal_sys::GDALGetColorInterpretationName(self.c_int()) };
-        _string(rv)
+        let c_ptr = unsafe { gdal_sys::GDALGetColorInterpretationName(self.c_int()) };
+        _string(c_ptr).unwrap_or_default()
     }
 }
 
--- a/src/raster/types.rs
+++ b/src/raster/types.rs
@@ -103,15 +103,7 @@ impl GdalDataType {
     /// ```
     pub fn name(&self) -> String {
         let c_str = unsafe { GDALGetDataTypeName(self.gdal_ordinal()) };
-        if c_str.is_null() {
-            // This case shouldn't happen, because `self` only exists for valid
-            // GDALDataType ordinals.
-            panic!(
-                "GDALGetDataTypeName unexpectedly returned an empty name for {:?}",
-                &self
-            );
-        }
-        _string(c_str)
+        _string(c_str).unwrap_or_default()
     }
 
     /// Get the [`GDALDataType`] size in **bits**.
--- a/src/spatial_ref/srs.rs
+++ b/src/spatial_ref/srs.rs
@@ -154,7 +154,7 @@ impl SpatialRef {
                 method_name: "OSRExportToWkt",
             })
         } else {
-            Ok(_string(c_wkt))
+            Ok(_string(c_wkt).unwrap_or_default())
         };
         unsafe { gdal_sys::VSIFree(c_wkt.cast::<std::ffi::c_void>()) };
         res
@@ -180,7 +180,7 @@ impl SpatialRef {
                 method_name: "OSRExportToPrettyWkt",
             })
         } else {
-            Ok(_string(c_wkt))
+            Ok(_string(c_wkt).unwrap_or_default())
         };
         unsafe { gdal_sys::VSIFree(c_wkt.cast::<std::ffi::c_void>()) };
         res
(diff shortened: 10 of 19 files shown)


src/vector/ops/conversions/formats.rs
```diff
@@ -92,8 +92,8 @@ impl Geometry {
                 method_name: "OGR_G_ExportToWkt",
             });
         }
-        let wkt = _string(c_wkt);
-        unsafe { gdal_sys::OGRFree(c_wkt as *mut c_void) };
+        let wkt = _string(c_wkt).unwrap_or_default();
+        unsafe { gdal_sys::VSIFree(c_wkt as *mut c_void) };
```
lnicola, line 96: `OGRFree` is deprecated.

src/gcp.rs
```diff
@@ -81,8 +81,8 @@ impl<'p> GcpRef<'p> {
 impl From<&gdal_sys::GDAL_GCP> for Gcp {
     fn from(gcp: &gdal_sys::GDAL_GCP) -> Self {
         Gcp {
-            id: _string(gcp.pszId),
-            info: _string(gcp.pszId),
+            id: _string(gcp.pszId).unwrap_or_default(),
+            info: _string(gcp.pszInfo).unwrap_or_default(),
```
lnicola, line 85: :see_no_evil:

src/version.rs
```diff
@@ -54,7 +54,8 @@ use std::fmt::Write;
 /// Details: [`const char *GDALVersionInfo(const char*)`](https://gdal.org/api/raster_c_api.html#_CPPv415GDALVersionInfoPKc)
 pub fn version_info(key: &str) -> String {
     let c_key = CString::new(key.as_bytes()).unwrap();
-    _string(unsafe { gdal_sys::GDALVersionInfo(c_key.as_ptr()) })
+    let rv = unsafe { gdal_sys::GDALVersionInfo(c_key.as_ptr()) };
```
weiznich, line 57: Do we need to free `rv` here? I assume that this is the equivalent of a `&'static str` on the c side, right?

src/vector/layer.rs
```diff
@@ -284,7 +284,7 @@ pub trait LayerAccess: Sized {
     /// Get the name of this layer.
     fn name(&self) -> String {
         let rv = unsafe { gdal_sys::OGR_L_GetName(self.c_layer()) };
```
weiznich, line 286: We don't need to free `rv` here because it's owned by the layer?

src/vector/geometry.rs
```diff
@@ -358,9 +354,7 @@ impl Eq for Geometry {}
 
 pub fn geometry_type_to_name(ty: OGRwkbGeometryType::Type) -> String {
     let rv = unsafe { gdal_sys::OGRGeometryTypeToName(ty) };
```
weiznich, line 356: We don't need to free `rv` here as it's essentially a static str, right?

src/vector/geometry.rs
```diff
@@ -170,11 +170,7 @@ impl Geometry {
         // Note: C API makes no statements about this possibly returning null.
         // So we don't have to result wrap this,
         let c_str = unsafe { gdal_sys::OGR_G_GetGeometryName(self.c_geometry()) };
```
weiznich, line 172: This is essentially a static string, right?

src/vector/feature.rs
```diff
@@ -835,7 +834,7 @@ impl FieldValue {
 
 pub fn field_type_to_name(ty: OGRFieldType::Type) -> String {
     let rv = unsafe { gdal_sys::OGR_GetFieldTypeName(ty) };
-    _string(rv)
+    _string(rv).unwrap_or_default()
```
weiznich, line 837: `rv` is a static string here, right?

src/vector/feature.rs
```diff
@@ -625,13 +624,13 @@ impl<'a> Iterator for FieldValueIterator<'a> {
             let field_defn =
                 unsafe { gdal_sys::OGR_F_GetFieldDefnRef(self.feature.c_feature, idx as c_int) };
             let field_name = unsafe { gdal_sys::OGR_Fld_GetNameRef(field_defn) };
```
weiznich, line 626: `field_name` is owned by `field_defn` so we wont need to free it, right?

src/vector/feature.rs
```diff
@@ -90,7 +90,7 @@ impl<'a> Feature<'a> {
         match field_type {
             OGRFieldType::OFTString => {
                 let rv = unsafe { gdal_sys::OGR_F_GetFieldAsString(self.c_feature, field_idx) };
```
weiznich, line 92: `rv` is owned by `c_feature` so we don't need to free it?

src/vector/defn.rs
```diff
@@ -206,7 +203,7 @@ impl<'a> GeomField<'a> {
     /// Get the name of this field.
     pub fn name(&'a self) -> String {
         let rv = unsafe { gdal_sys::OGR_GFld_GetNameRef(self.c_field_defn) };
```
weiznich, line 205: `rv` is owned by `c_field_defn` so we don't need to free it?

src/vector/defn.rs
```diff
@@ -162,11 +162,8 @@ impl<'a> Field<'a> {
 
     /// Get default field value.
     pub fn default_value(&'a self) -> Option<String> {
-        let rv = unsafe { gdal_sys::OGR_Fld_GetDefault(self.c_field_defn) };
-        if rv.is_null() {
-            return None;
-        }
-        Some(_string(rv))
+        let c_ptr = unsafe { gdal_sys::OGR_Fld_GetDefault(self.c_field_defn) };
```
weiznich, line 165: `c_ptr` is owned by `c_field_defn` so we don't need to free it here?

src/vector/defn.rs
```diff
@@ -122,13 +122,13 @@ impl<'a> Field<'a> {
     /// Get the name of this field.
     pub fn name(&'a self) -> String {
         let rv = unsafe { gdal_sys::OGR_Fld_GetNameRef(self.c_field_defn) };
-        _string(rv)
+        _string(rv).unwrap_or_default()
     }
 
     /// Get the alternative name (alias) of this field.
     pub fn alternative_name(&'a self) -> String {
         let rv = unsafe { gdal_sys::OGR_Fld_GetAlternativeNameRef(self.c_field_defn) };
```
weiznich, line 130: `rv` is owned by `c_field_defn` here so we don't need to free it?

src/utils.rs
```diff
@@ -5,9 +5,13 @@ use gdal_sys::CPLErr;
 
 use crate::errors::*;
 
-pub fn _string(raw_ptr: *const c_char) -> String {
-    let c_str = unsafe { CStr::from_ptr(raw_ptr) };
-    c_str.to_string_lossy().into_owned()
+pub fn _string(raw_ptr: *const c_char) -> Option<String> {
```
weiznich, line 8: It might be worth to add a documentation comment to this function and to `__pathbuf` to point out that they perform null checks internally.

src/spatial_ref/srs.rs
```diff
@@ -580,12 +567,8 @@ impl SpatialRef {
         let child = child.try_into().expect("`child` must fit in `c_int`");
 
         let c_node_path = CString::new(node_path)?;
-        let rv = unsafe { gdal_sys::OSRGetAttrValue(self.0, c_node_path.as_ptr(), child) };
-        if rv.is_null() {
-            Ok(None)
-        } else {
-            Ok(Some(_string(rv)))
-        }
+        let c_ptr = unsafe { gdal_sys::OSRGetAttrValue(self.0, c_node_path.as_ptr(), child) };
```
weiznich, line 570: `c_ptr` is owned by `c_node_path` here, so we don't need to free it?

src/spatial_ref/srs.rs
```diff
@@ -379,7 +370,7 @@ impl SpatialRef {
     }
 
     pub fn axis_name(&self, target_key: &str, axis: i32) -> Result<String> {
-        // See get_axis_orientation
+        // See axis_orientation
         let c_ptr = unsafe {
```
weiznich, line 374: axis names are assumed to be static, so no need to free it?

src/spatial_ref/srs.rs
```diff
@@ -284,34 +284,25 @@ impl SpatialRef {
...
+        _string(c_ptr)
     }
 
     pub fn angular_units(&self) -> f64 {
         unsafe { gdal_sys::OSRGetAngularUnits(self.0, ptr::null_mut()) }
     }
 
-    pub fn linear_units_name(&self) -> Result<String> {
+    pub fn linear_units_name(&self) -> Option<String> {
         let mut c_ptr = ptr::null_mut();
```
weiznich, line 303: unit names are assumed to be static, so again no need to free it?

src/spatial_ref/srs.rs
```diff
@@ -284,34 +284,25 @@ impl SpatialRef {
...
-            return Err(_last_null_pointer_err("OSRGetName"));
-        }
-        Ok(_string(c_ptr))
+        _string(c_ptr)
     }
 
-    pub fn angular_units_name(&self) -> Result<String> {
+    pub fn angular_units_name(&self) -> Option<String> {
         let mut c_ptr = ptr::null_mut();
         unsafe { gdal_sys::OSRGetAngularUnits(self.0, &mut c_ptr) };
```
weiznich, line 294: Angular units are assumed to be static, so no need to free them?

src/spatial_ref/srs.rs
```diff
@@ -284,34 +284,25 @@ impl SpatialRef {
     }
 
     #[cfg(major_ge_3)]
-    pub fn name(&self) -> Result<String> {
+    pub fn name(&self) -> Option<String> {
         let c_ptr = unsafe { gdal_sys::OSRGetName(self.0) };
```
weiznich, line 288: `c_ptr` is onwed by self, so we don't need to free it?

src/spatial_ref/srs.rs
```diff
@@ -226,41 +226,41 @@ impl SpatialRef {
...
-            Ok(_string(c_projjsonstr))
+            Ok(_string(c_projjsonstr).unwrap_or_default())
         };
         unsafe { gdal_sys::VSIFree(c_projjsonstr.cast::<std::ffi::c_void>()) };
         res
     }
 
-    pub fn auth_name(&self) -> Result<String> {
+    pub fn auth_name(&self) -> Option<String> {
         let c_ptr = unsafe { gdal_sys::OSRGetAuthorityName(self.0, ptr::null()) };
```
weiznich, line 236: The same is assumed to be owned by `self` so there is no need to free `c_ptr`?

src/raster/types.rs
```diff
@@ -103,15 +103,7 @@ impl GdalDataType {
     /// ```
     pub fn name(&self) -> String {
         let c_str = unsafe { GDALGetDataTypeName(self.gdal_ordinal()) };
```
weiznich, line 105: `c_str` is owned by `self` so we don't need to free it?

src/raster/rasterband.rs
```diff
@@ -1362,8 +1357,8 @@ impl ColorInterpretation {
 
     /// Returns the name of this color interpretation.
     pub fn name(&self) -> String {
-        let rv = unsafe { gdal_sys::GDALGetColorInterpretationName(self.c_int()) };
-        _string(rv)
+        let c_ptr = unsafe { gdal_sys::GDALGetColorInterpretationName(self.c_int()) };
```
weiznich, line 1360: `c_ptr` is owned by `self` so we don't need to free it?

src/raster/rasterband.rs
```diff
@@ -950,13 +950,8 @@ impl<'a> RasterBand<'a> {
...
     /// If there is no unit, the empty string is returned.
     pub fn unit(&self) -> String {
-        let str_ptr = unsafe { gdal_sys::GDALGetRasterUnitType(self.c_rasterband) };
-
-        if str_ptr.is_null() {
-            return String::new();
-        }
-
-        _string(str_ptr)
+        let c_ptr = unsafe { gdal_sys::GDALGetRasterUnitType(self.c_rasterband) };
```
weiznich, line 953: `c_ptr` is owned by `self` so we don't need to free it?

src/raster/mdarray.rs
```diff
@@ -767,12 +763,9 @@ impl Attribute {
...
 
     pub fn read_as_string(&self) -> String {
-        unsafe {
-            // SAFETY: should no be freed
-            let c_string = GDALAttributeReadAsString(self.c_attribute);
-
-            _string(c_string)
-        }
+        // should no be freed
+        let c_ptr = unsafe { GDALAttributeReadAsString(self.c_attribute) };
```
weiznich, line 767: `c_ptr` is owned by `self` so we don't need to free it?

src/raster/mdarray.rs
```diff
@@ -712,7 +707,8 @@ impl ExtendedDataType {
     }
 
     pub fn name(&self) -> String {
-        _string(unsafe { GDALExtendedDataTypeGetName(self.c_data_type) })
+        let c_ptr = unsafe { GDALExtendedDataTypeGetName(self.c_data_type) };
```
weiznich, line 710: `c_ptr` is owned by `self` so we don't need to free it?

src/raster/mdarray.rs
```diff
@@ -624,7 +618,8 @@ impl<'a> Dimension<'a> {
     }
 
     pub fn name(&self) -> String {
-        _string(unsafe { GDALDimensionGetName(self.c_dimension) })
+        let c_ptr = unsafe { GDALDimensionGetName(self.c_dimension) };
```
weiznich, line 621: `c_ptr` is owned by `self` so we don't need to free it?

src/raster/mdarray.rs
```diff
@@ -488,7 +481,8 @@ impl<'a> Group<'a> {
     }
 
     pub fn name(&self) -> String {
-        _string(unsafe { GDALGroupGetName(self.c_group) })
+        let c_ptr = unsafe { GDALGroupGetName(self.c_group) };
```
weiznich, line 484: `c_ptr` is owned by `self` so we don't need to free it?

src/raster/mdarray.rs
```diff
@@ -376,9 +370,8 @@ impl<'a> MDArray<'a> {
     pub fn unit(&self) -> String {
         unsafe {
             // should not be freed
-            let c_unit = GDALMDArrayGetUnit(self.c_mdarray);
-
-            _string(c_unit)
+            let c_ptr = GDALMDArrayGetUnit(self.c_mdarray);
```
weiznich, line 373: `c_ptr` is owned by `self` so we don't need to free it?

src/raster/mdarray.rs
```diff
@@ -331,16 +331,10 @@ impl<'a> MDArray<'a> {
...
-                    } else {
-                        let string = _string(string_ptr);
-
-                        VSIFree(string_ptr as *mut c_void);
-
-                        string
-                    }
+                .filter_map(|ptr| {
+                    let s = _string(ptr);
+                    VSIFree(ptr as *mut c_void);
```
weiznich, line 336: ptr might here be actually null, free functions usually do not like to get a `nullptr` passed into them. Maybe better write

```
let s = _string(ptr)?;
VSIFree(ptr as *mut c_void);
Some(s)
```

src/metadata.rs
```diff
@@ -59,10 +59,7 @@ pub trait Metadata: MajorObject {
     /// ```
     fn description(&self) -> Result<String> {
         let c_res = unsafe { gdal_sys::GDALGetDescription(self.gdal_object_ptr()) };
```
weiznich, line 61: `c_res` is owned by self and we don't need to free it?

src/gcp.rs
```diff
@@ -133,10 +133,7 @@ impl Dataset {
     ///  See: [`GDALGetGCPProjection`](https://gdal.org/api/raster_c_api.html#gdal_8h_1a85ffa184d3ecb7c0a59a66096b22b2ec)
     pub fn gcp_projection(&self) -> Option<String> {
         let cc_ptr = unsafe { gdal_sys::GDALGetGCPProjection(self.c_dataset()) };
```
weiznich, line 135: `cc_ptr` is owned by self and we don't need to free it?

src/gcp.rs
weiznich: Maybe rather replace this function with a `Gcp::from_ptr(&gdal_sys::GDAL_GCP) -> Option<Gcp>` function?

src/driver.rs
```diff
@@ -66,7 +66,7 @@ impl Driver {
     /// See also: [`long_name`](Self::long_name).
     pub fn short_name(&self) -> String {
         let rv = unsafe { gdal_sys::GDALGetDriverShortName(self.c_driver) };
```
weiznich, line 68: `rv` is owned by self, so we don't need to free it here?

src/driver.rs
```diff
@@ -76,7 +76,7 @@ impl Driver {
     /// See also: [`short_name`](Self::short_name`).
     pub fn long_name(&self) -> String {
         let rv = unsafe { gdal_sys::GDALGetDriverLongName(self.c_driver) };
```
weiznich, line 78: `rv` is owned by self, so we don't need to free it here?

src/dataset.rs
```diff
@@ -208,7 +208,7 @@ impl Dataset {
     /// Fetch the projection definition string for this dataset.
     pub fn projection(&self) -> String {
         let rv = unsafe { gdal_sys::GDALGetProjectionRef(self.c_dataset) };
```
weiznich, line 210: `rv` is owned by self, so we don't need to free it here?

src/cpl.rs
```diff
@@ -240,20 +235,13 @@ impl CslStringList {
         if index > c_int::MAX as usize {
             return None;
         }
+
         // In the C++ implementation, an index-out-of-bounds returns an empty string, not an error.
         // We don't want to check against `len` because that scans the list.
         // See: https://github.com/OSGeo/gdal/blob/fada29feb681e97f0fc4e8861e07f86b16855681/port/cpl_string.cpp#L181-L182
         let field = unsafe { CSLGetField(self.as_ptr(), index as c_int) };
```
weiznich, line 242: `field` is owned by self, so we don't need to free it here?

src/cpl.rs
```diff
@@ -505,12 +493,7 @@ impl<'a> Iterator for CslStringListIterator<'a> {
             slice[self.idx]
         };
         self.idx += 1;
-        if field.is_null() {
-            None
-        } else {
-            let entry = _string(field);
-            Some(entry.deref().into())
-        }
+        _string(field).map(|s| s.into())
```
weiznich, line 496: `field` is owned by `self.list.list_ptr`, so we don't need to free it here?

src/cpl.rs
```diff
@@ -178,11 +177,7 @@ impl CslStringList {
         // we know already `name` will never exist in a valid CslStringList.
         let key = CString::new(name).ok()?;
         let c_value = unsafe { CSLFetchNameValue(self.as_ptr(), key.as_ptr()) };
```
weiznich, line 179: `c_value` is owned by self, so we don't need to free it here?

src/config.rs
```diff
@@ -96,7 +96,7 @@ pub fn get_thread_local_config_option(key: &str, default: &str) -> Result<String
     let c_key = CString::new(key.as_bytes())?;
     let c_default = CString::new(default.as_bytes())?;
     let rv = unsafe { gdal_sys::CPLGetThreadLocalConfigOption(c_key.as_ptr(), c_default.as_ptr()) };
```
weiznich, line 98: `rv` is points to a thread local, so we don't need to free it here?

src/config.rs
```diff
@@ -55,7 +55,7 @@ pub fn get_config_option(key: &str, default: &str) -> Result<String> {
     let c_key = CString::new(key.as_bytes())?;
     let c_default = CString::new(default.as_bytes())?;
     let rv = unsafe { gdal_sys::CPLGetConfigOption(c_key.as_ptr(), c_default.as_ptr()) };
```
weiznich, line 57: `rv` is points to a global variable so we don't need to free it here?

weiznich (commented):
I think overall that is a good change.

I'm relatively new to the low level gdal interfaces, so I added "some" questions around which pointer comes from where. Feel free to dismiss them in batch if the assumptions written down there are correct, although it might be good to just put that as comments into the code.

I think in one location I might have discovered a problematic free call.

src/version.rs
lnicola: I think it's more complicated (it's a string that GDAL formats and stores in TLS IIRC), but according to the docs:

> an internal string containing the requested information

So we can treat it as static.

src/vector/layer.rs
lnicola: > the layer name (must not been freed)

src/vector/geometry.rs
lnicola: > internal human readable string, or NULL on failure.

(but it's never NULL in the current implementation, it returns something like "Unrecognized: XXXX" in a static buffer)

src/vector/geometry.rs
lnicola: Apparently so, it returns a static string.

src/vector/feature.rs
lnicola: Yes, returns a static string, never NULL.

src/vector/feature.rs
lnicola: Yes, it's owned by the field (I think it's a general pattern with `GetFooRef()`).

src/vector/feature.rs
lnicola: Yes:

> This string is internal, and should not be modified, or freed. Its lifetime may be very brief.

src/vector/defn.rs
lnicola: Yes, owned by `OGRGeomFieldDefn`.

src/vector/defn.rs
lnicola: Owned by `OGRFieldDefn`.

src/raster/mdarray.rs
lnicola: @rouault would you take a GDAL doc PR to explicitly mention that `VSIFree` can be called with a null pointer?

src/vector/defn.rs
lnicola: Yes, owned by `OGRFieldDefn`.

src/spatial_ref/srs.rs
lnicola: Yes, owned by the `OGR_SRSNode`.

src/spatial_ref/srs.rs
lnicola: Comes from `OGR_SRSNode`.

src/spatial_ref/srs.rs
lnicola: Owned by `OGRSpatialReference`.

src/spatial_ref/srs.rs
lnicola: Owned by `OGRSpatialReference`.

src/spatial_ref/srs.rs
lnicola: Owned by PROJ via `OGRSpatialReference`.

src/spatial_ref/srs.rs
lnicola: Owned by PROJ via `OGRSpatialReference`.

src/raster/types.rs
lnicola: `GDALGetDataTypeName` returns static strings.

src/raster/rasterband.rs
lnicola: `GDALGetColorInterpretationName` returns static strings.

src/raster/rasterband.rs
lnicola: Owned by `GDALRasterBand`.

> The returned string should not be modified, nor freed by the calling application.

src/version.rs
rouault: > Do we need to free `rv` here?

As a rule of thumb, that part of the GDAL C API should be relatively easy to get:
- For C/C++ functions that return a "const char*", as a "const char*" cannot be passed directly to C free(),  that indicates that the returned value should not be freed
- And the documentation of functions where the return value must be freed indicates so, and generally will specify which free function is expected. (like OGR_G_DestroyGeometry() for anything OGRGeometryH related, etc)

src/gcp.rs
lnicola: I think it's fine, they docs say about `pszInfo`:

> Informational message or "". 

GDAL calls `json_object_new_string` on them [here](https://github.com/OSGeo/gdal/blob/0caa252cda18ce1e68028403b9719e3ca898ae50/apps/gdalinfo_lib.cpp#L806-L807), which [assumes](https://github.com/OSGeo/gdal/blob/0caa252cda18ce1e68028403b9719e3ca898ae50/ogr/ogrsf_frmts/geojson/libjson/json_object.c#L1307) that the pointer is non-null by calling `strlen` on it.

And in most places `pszInfo` is initialized with `CSLStrDup("")`, which matches the docs. As for `pszid`, that's usually initialized with a number, I don't think it can be null.

src/utils.rs
```diff
@@ -5,35 +5,40 @@ use gdal_sys::CPLErr;
 
 use crate::errors::*;
 
-pub fn _string(raw_ptr: *const c_char) -> String {
-    let c_str = unsafe { CStr::from_ptr(raw_ptr) };
-    c_str.to_string_lossy().into_owned()
-}
-
-pub(crate) fn _string_tuple(raw_ptr: *const c_char, delim: char) -> Option<(String, String)> {
```
lnicola, line 13: This was also unsound, but never used.

PR #594, opened 2024-11-29
Add Defn::geometry_field_index

lnicola:
---


Atreyagaurav:
I don't know if I'm qualified to review... It looks like minor changes, looks good to me.

Btw, I see the  `Feature::field` now takes index, and you use `Feature::field_index` to get the index, but wouldn't it still be a good idea to have `Feature::field_by_name` since it used to be present, and `geometry_by_name` is present.

lnicola:
> wouldn't it still be a good idea to have `Feature::field_by_name` since it used to be present

No, since it can be very slow and it's really easy to use it by mistake. The index-based API takes one more line to use, but it mirrors the GDAL API, and you don't risk making your code 1.5x slower.

> `geometry_by_name` is present

Good point, this.

Atreyagaurav:
I meant: keep the function, but use `field_index` and `field` internally. Shouldn't be slow, just convenient.

lnicola:
No, because if you have 1M features you can get the index once and reuse it. Otherwise you need to look the field up once for every feature.

Atreyagaurav:
Ohk. So basically this API will force people to make hash map if they have to get field index by name in a loop, and it'll be lot faster, makes sense.

lnicola:
Not really. Most of the time you either:

 - have a relatively small, known set of fields: you can just query them into variables before iterating over the layer
 - don't know what fields to expect (maybe you're writing `ogr2ogr`): you might as well query them by index

lnicola:
Updated with a new `Feature::geometry_field_index` to match `Feature::field_index`. It's only used in a test and the documentation suggests using the `Defn` methods instead.

Atreyagaurav:
Probably irrelevant, but in case it helps, this is what I had to do previously when I had dynamic fields to write, but they could be null:

```rust
for n in nodes {
    // ...
    let feat_fields: Vec<(&str, FieldValue)> = fields
        .iter()
        .filter_map(|(k, (_, func))| Some((k.as_str(), func(n.attr(k)?))))
        .collect();
    
    let field_names: Vec<&str> = feat_fields.iter().map(|(k, _)| *k).collect();
    let field_vals: Vec<FieldValue> = feat_fields.into_iter().map(|(_, v)| v).collect();
    layer.create_feature_fields(node_geom, &field_names, &field_vals)?;
}
```
Which will prob become something like this now:
```rust
fields_ind: HashMap<&str, usize> = // ...
for n in nodes {
    // ...
    let mut feat = layer.create_feature(node_geom)?;
    let feat_fields: Vec<(&str, FieldValue)> = fields
        .iter()
        .filter_map(|(k, (_, func))| Some((k.as_str(), func(n.attr(k)?))))
        .collect();
    for (field, value) in feat_fields {
        // can do it with a for_each above but separating it to show the difference
        feat.set_field(fields_ind[field], value)?;
    }
}
```

Where `func` converts attribute values to FieldValue, but there is no Null on FieldValue enum, so I had to filter fields each time. Would be nice to not have to do that, kinda just `create_feature_fields(Geometry, &[Option<FieldValue>])`

Atreyagaurav (approved):

lnicola:
> Which will prob become something like this now:

I don't know how you create the nodes and the fields, but maybe you could create them in the same order, then just do:

```rust
for idx in 0..num_attrs {
    feat.set_field(idx, func(n.attr_by_idx(k)))?;
}
```

If not, you can still build that hashmap (or even an array) once, before iterating over `nodes`, assuming your nodes have their attributes in the same order.

Atreyagaurav:
The features to export to GIS will be a subset of features in the nodes, so even if it's ordered, we can't use their index.

Please ignore this, no need to spend time, I just wanted to show/verify what I think I would I have to do based on this API change, in case it is a use case you hadn't thought of. Caching the fields' indices outside the loop should be a good solution.

lnicola:
But you can still do:

```rust
let attrs_to_export = todo!();
for n in nodes {
    // ...
    for (field_idx, attr_idx) in attrs_to_export {
        feat.set_field(field_idx, func(n.attr_by_idx(attr_idx)))?;
    }
}
```

PR #600, opened 2025-01-03
Add M-dim bindings

JmsPae:
---
This PR seeks to add bindings for GDAL functions relating to higher-dimension geometry, in particular Measured geometry, with some other minor fixes/additions.

Added methods for `Geometry`:
- `add_point_zm`
- `add_point_m`
- `set_point_zm`
- `set_point_m`
- `get_point_zm`
- `get_point_vec_zm`
- `iso_wkt`
- `iso_wkb`

Considering renaming `get_point_vec_zm` to `get_point_zm_vec`, though neither exactly roll off the tongue.

Also modified the Debug impl. for `Geometry` to use iso_wkt in cases of measured geometry.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,10 +28,13 @@
   - Add `Defn::field_index` and `Feature::field_index` ([#581](https://github.com/georust/gdal/pull/581))
   - Add `Defn::geometry_field_index` and `Feature::geometry_field_index` ([#594](https://github.com/georust/gdal/pull/594))
   - Add `Dataset::has_capability` for dataset capability check ([#581](https://github.com/georust/gdal/pull/585))
+  - Add methods `add_point_zm`, `add_point_m`, `set_point_zm`, `set_point_m`, `get_point_zm`, `get_point_vec_zm`, `iso_wkt` and `iso_wkb` to `Geometry` ([#600](https://github.com/georust/gdal/pull/600))
+  - Add functions `geometry_type_flatten`, `geometry_type_set_z`, `geometry_type_set_m`, `geometry_type_set_modifier`, `geometry_type_has_z` and `geometry_type_has_m` to `vector::geometry` ([#600](https://github.com/georust/gdal/pull/600))
 
 ### Fixed
 
   - Fix conversion from `ndarray` when the data is offsetted from the start of the buffer ([#569](https://github.com/georust/gdal/pull/569))
+  - `Debug` trait for `Geometry` now prints ISO WKT when the M-dim is present ([#600](https://github.com/georust/gdal/pull/600))
 
 ### Removed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,7 +34,7 @@
 ### Fixed
 
   - Fix conversion from `ndarray` when the data is offsetted from the start of the buffer ([#569](https://github.com/georust/gdal/pull/569))
-  - `Debug` trait for `Geometry` now prints ISO WKT when the M-dim is present ([#600](https://github.com/georust/gdal/pull/600))
+  - use ISO WKT for the `Debug` implementation of `Geometry`, in order to properly display measure values,  ([#600](https://github.com/georust/gdal/pull/600))
 
 ### Removed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@
   - Drop `LayerAccess::create_feature_fields` ([#581](https://github.com/georust/gdal/pull/581))
   - Drop `Feature::geometry_by_name` ([#594](https://github.com/georust/gdal/pull/594))
   - Update `SpatialRef::auth_name`, `SpatialRef::name`, `SpatialRef::angular_units_name`, `SpatialRef::linear_units_name` to return `Option<String>` instead of `Result<String>` ([#589](https://github.com/georust/gdal/pull/589))
+  - Update `Geometry::get_point_vec` to modify a `&mut Vec` as opposed to new allocations, returning an `i32` of points added ([#600](https://github.com/georust/gdal/pull/600))
 
 ### Added
 
--- a/src/vector/geometry.rs
+++ b/src/vector/geometry.rs
@@ -457,7 +457,7 @@ pub fn geometry_type_set_modifier(ty: OGRwkbGeometryType::Type, set_z: bool, set
     unsafe { gdal_sys::OGR_GT_SetModifier(ty, set_z as i32, set_m as i32) }
 }
 
-/// Return if the geometry type is a 3D geometry type. 
+/// Returns `true` if the geometry type is a 3D geometry type. 
 pub fn geometry_type_has_z(ty: OGRwkbGeometryType::Type) -> bool {
     unsafe { gdal_sys::OGR_GT_HasZ(ty) != 0 }
 }
--- a/src/vector/geometry.rs
+++ b/src/vector/geometry.rs
@@ -408,13 +408,7 @@ impl Clone for Geometry {
 
 impl Debug for Geometry {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let should_iso = geometry_type_has_m(self.geometry_type());
-
-        // .wkt() does not return the M coordinate regardless or whether or not the Z coordinate is present.
-        let wkt = match should_iso {
-            true => self.iso_wkt(),
-            false => self.wkt()
-        };
+        let wkt = self.iso_wkt();
 
         match wkt {
             Ok(wkt) => f.write_str(wkt.as_str()),
--- a/src/vector/geometry.rs
+++ b/src/vector/geometry.rs
@@ -217,14 +217,28 @@ impl Geometry {
         (x, y, z, m)
     }
 
-    pub fn get_point_vec(&self) -> Vec<(f64, f64, f64)> {
+    /// Appends all points of a line string to `out_points`. 
+    ///
+    /// Only wkbPoint[X] or wkbLineString[X] may alter `out_points`. Other geometry types will silently do nothing, see
+    /// [`OGR_G_GetPointCount`](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv419OGR_G_GetPointCount12OGRGeometryH)
+    pub fn get_point_vec(&self, out_points: &mut Vec<(f64, f64, f64)>) {
+        // Consider replacing logic with
+        // [OGR_G_GetPoints](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv415OGR_G_GetPoints12OGRGeometryHPviPviPvi)
         let length = unsafe { gdal_sys::OGR_G_GetPointCount(self.c_geometry()) };
-        (0..length).map(|i| self.get_point(i)).collect()
+        out_points.extend(
+            (0..length).map(|i| self.get_point(i))
+        );
     }
 
-    pub fn get_point_vec_zm(&self) -> Vec<(f64, f64, f64, f64)> {
+    /// Appends all points of a line string to `out_points`. 
+    ///
+    /// Only wkbPoint[X] or wkbLineString[X] may alter `out_points`. Other geometry types will silently do nothing, see
+    /// [`OGR_G_GetPointCount`](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv419OGR_G_GetPointCount12OGRGeometryH)
+    pub fn get_point_vec_zm(&self, out_points: &mut Vec<(f64, f64, f64, f64)>) {
         let length = unsafe { gdal_sys::OGR_G_GetPointCount(self.c_geometry()) };
-        (0..length).map(|i| self.get_point_zm(i)).collect()
+        out_points.extend(
+            (0..length).map(|i| self.get_point_zm(i))
+        );
     }
 
     /// Get the geometry type ordinal
@@ -489,7 +503,7 @@ impl Debug for GeometryRef<'_> {
 
 #[cfg(test)]
 mod tests {
-    use self::OGRwkbGeometryType::{wkbMultiPointZM, wkbPointZM};
+    use self::OGRwkbGeometryType::{wkbLineStringZM, wkbMultiPointZM, wkbPointZM};
 
     use super::*;
     use crate::spatial_ref::SpatialRef;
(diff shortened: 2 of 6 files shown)


JmsPae:
`update changelog` also adds `set_point_m` and `add_point_m` to `Geometry`.

src/vector/ops/conversions/formats.rs
```diff
@@ -117,6 +132,26 @@ impl Geometry {
...
+            unsafe { gdal_sys::OGR_G_ExportToIsoWkb(self.c_geometry(), byte_order, wkb.as_mut_ptr()) };
+        if rv != gdal_sys::OGRErr::OGRERR_NONE {
+            return Err(GdalError::OgrError {
+                err: rv,
+                method_name: "OGR_G_ExportToIsoWkb",
+            });
+        }
+        Ok(wkb)
+    }
+
```
JmsPae, line 154: Maybe instead of a separate function, modifying the original `wkb`/`wkt` methods would be nicer, though that would be a breaking change.

src/vector/geometry.rs
```diff
@@ -357,6 +436,37 @@ pub fn geometry_type_to_name(ty: OGRwkbGeometryType::Type) -> String {
...
+/// Return if the geometry type is a 3D geometry type. 
+pub fn geometry_type_has_z(ty: OGRwkbGeometryType::Type) -> bool {
+    unsafe { gdal_sys::OGR_GT_HasZ(ty) != 0 }
+}
+
+/// Return if the geometry type is a measured type.
+pub fn geometry_type_has_m(ty: OGRwkbGeometryType::Type) -> bool {
+    unsafe { gdal_sys::OGR_GT_HasM(ty) != 0 }
+}
+
```
JmsPae, line 469: Not sure about adding a bunch of stand-alone functions like this when they only take an `OGRwkbGeometryType` though since that's generated, this might just be good enough?

src/vector/geometry.rs
```diff
@@ -151,11 +203,30 @@ impl Geometry {
...
+        unsafe { gdal_sys::OGR_G_GetPointZM(self.c_geometry(), index, &mut x, &mut y, &mut z, &mut m) };
+        (x, y, z, m)
+    }
+
     pub fn get_point_vec(&self) -> Vec<(f64, f64, f64)> {
         let length = unsafe { gdal_sys::OGR_G_GetPointCount(self.c_geometry()) };
         (0..length).map(|i| self.get_point(i)).collect()
     }
 
+    pub fn get_point_vec_zm(&self) -> Vec<(f64, f64, f64, f64)> {
```
lnicola, line 225: I know it's for symmetry with `get_point_vec`, but I don't like this API because allocates a new `Vec`. It's easy for the user to implement this properly, but having the `get_point_vec` will nudge them towards the inefficient approach.

Are you planning to use it?

src/vector/geometry.rs
```diff
@@ -357,6 +436,37 @@ pub fn geometry_type_to_name(ty: OGRwkbGeometryType::Type) -> String {
...
+pub fn geometry_type_set_m(ty: OGRwkbGeometryType::Type) -> OGRwkbGeometryType::Type {
+    unsafe { gdal_sys::OGR_GT_SetM(ty) }
+}
+
+/// Returns a XY, XYZ, XYM or XYZM geometry type depending on parameter.
+pub fn geometry_type_set_modifier(ty: OGRwkbGeometryType::Type, set_z: bool, set_m: bool) -> OGRwkbGeometryType::Type {
+    unsafe { gdal_sys::OGR_GT_SetModifier(ty, set_z as i32, set_m as i32) }
+}
+
+/// Return if the geometry type is a 3D geometry type.
```
lnicola, line 460:
```suggestion
/// Returns `true` if the geometry type is a 3D geometry type. 
```

src/vector/geometry.rs
```diff
@@ -357,6 +436,37 @@ pub fn geometry_type_to_name(ty: OGRwkbGeometryType::Type) -> String {
...
+pub fn geometry_type_set_modifier(ty: OGRwkbGeometryType::Type, set_z: bool, set_m: bool) -> OGRwkbGeometryType::Type {
+    unsafe { gdal_sys::OGR_GT_SetModifier(ty, set_z as i32, set_m as i32) }
+}
+
+/// Return if the geometry type is a 3D geometry type. 
+pub fn geometry_type_has_z(ty: OGRwkbGeometryType::Type) -> bool {
+    unsafe { gdal_sys::OGR_GT_HasZ(ty) != 0 }
+}
+
+/// Return if the geometry type is a measured type.
```
lnicola, line 465:
```suggestion
/// Returns `true` if the geometry type is a measured type.
```

src/vector/ops/conversions/formats.rs
```diff
@@ -117,6 +132,26 @@ impl Geometry {
...
 
+    /// Serializes the geometry to SFSQL 1.2 / ISO SQL / MM Part 3
+    /// [WKB](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary)
+    /// (Well-Known Binary) format.
+    pub fn iso_wkb(&self) -> Result<Vec<u8>> {
+        let wkb_size = unsafe { gdal_sys::OGR_G_WkbSize(self.c_geometry()) as usize };
+        // We default to little-endian for now. A WKB string explicitly indicates the byte
+        // order, so this is not a problem for interoperability.
+        let byte_order = gdal_sys::OGRwkbByteOrder::wkbNDR;
+        let mut wkb = vec![0; wkb_size];
```
lnicola, line 143: We could use `MaybeUninit` here, but feel free to just add a comment instead.

lnicola:
Quick r? @weiznich if you have time.

src/vector/geometry.rs
```diff
@@ -114,6 +114,33 @@ impl Geometry {
         };
     }
 
+    pub fn set_point_zm(&mut self, i: usize, point: (f64, f64, f64, f64)) {
```
weiznich, line 117: Given that the underlying gdal function only accepts a `c_int` (which should be 32bit on most platforms?) we might want to change the index to `u32` instead?

src/vector/geometry.rs
```diff
@@ -151,11 +203,30 @@ impl Geometry {
         (x, y, z)
     }
 
+    /// Get point coordinates from a line string or a point geometry.
+    ///
+    /// `index` is the line string vertex index, from 0 to `point_count()-1`, or `0` when a point.
+    ///
+    /// See: [`OGR_G_GetPointZM`](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv416OGR_G_GetPointZM12OGRGeometryHiPdPdPdPd)
+    pub fn get_point_zm(&self, index: i32) -> (f64, f64, f64, f64) {
```
weiznich, line 211: The index type should be consistent with the add method?

weiznich (commented):
Looks ok for me beside the inconsistencies around the index type. Those are not critical, but it would be nice to see them addressed.

lnicola:
I noticed those, but they match the existing `add_point` and `set_point`.

But in https://github.com/georust/gdal/pull/581 I went the other way around (`i32` to `usize`)..

src/vector/geometry.rs
```diff
@@ -337,7 +408,15 @@ impl Clone for Geometry {
 
 impl Debug for Geometry {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self.wkt() {
+        let should_iso = geometry_type_has_m(self.geometry_type());
```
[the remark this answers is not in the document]
lnicola: From Even:

> identical for non-curve 2D geometries ie Point, LineString, Polygon, their multi version and GeometryCollection
> classic one is mostly of historic "interest". ISO WKT and WKB should be preferred

CHANGES.md
```diff
@@ -28,10 +28,13 @@
   - Add `Defn::field_index` and `Feature::field_index` ([#581](https://github.com/georust/gdal/pull/581))
   - Add `Defn::geometry_field_index` and `Feature::geometry_field_index` ([#594](https://github.com/georust/gdal/pull/594))
   - Add `Dataset::has_capability` for dataset capability check ([#581](https://github.com/georust/gdal/pull/585))
+  - Add methods `add_point_zm`, `add_point_m`, `set_point_zm`, `set_point_m`, `get_point_zm`, `get_point_vec_zm`, `iso_wkt` and `iso_wkb` to `Geometry` ([#600](https://github.com/georust/gdal/pull/600))
+  - Add functions `geometry_type_flatten`, `geometry_type_set_z`, `geometry_type_set_m`, `geometry_type_set_modifier`, `geometry_type_has_z` and `geometry_type_has_m` to `vector::geometry` ([#600](https://github.com/georust/gdal/pull/600))
 
 ### Fixed
 
   - Fix conversion from `ndarray` when the data is offsetted from the start of the buffer ([#569](https://github.com/georust/gdal/pull/569))
+  - `Debug` trait for `Geometry` now prints ISO WKT when the M-dim is present ([#600](https://github.com/georust/gdal/pull/600))
```
lnicola, line 37:
```suggestion
  - use ISO WKT for the `Debug` implementation of `Geometry`, in order to properly display measure values,  ([#600](https://github.com/georust/gdal/pull/600))
```

src/vector/geometry.rs
JmsPae: We could also alter these methods to use [OGR_G_GetPoints](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv415OGR_G_GetPoints12OGRGeometryHPviPviPvi). Not sure how much faster it'd be, though I feel that's somewhat out of scope for this PR but it might be worth nipping in the bud whilst we're adding this method.

src/vector/ops/conversions/formats.rs
JmsPae: I'm not terribly sure how to go about it, so I'll leave a comment for now.

src/vector/geometry.rs
JmsPae: Should I leave it as-is for now? Food for thought for whenever we get around to switching to OGR_G_GetPoints.

lnicola:
I mean, `u32` feels worse than both since it doesn't match the C type (and you can pass in a value larger than `i32::MAX` which overflows), and it doesn't play that well with idiomatic Rust code, which probably has an `usize` around.

src/vector/geometry.rs
lnicola: Nah, let's take `&mut Vec`.

JmsPae:
We could do a checked conversion from `usize` to `i32`/`c_int` (eg. try_from) but then we'd have to make the methods fallible, which also isn't that great.

src/vector/geometry.rs
```diff
@@ -151,9 +203,42 @@ impl Geometry {
...
+        out_points.extend(
+            (0..length).map(|i| self.get_point(i))
+        );
+    }
+
+    /// Appends all points of a line string to `out_points`. 
+    ///
+    /// Only wkbPoint[X] or wkbLineString[X] may alter `out_points`. Other geometry types will silently do nothing, see
+    /// [`OGR_G_GetPointCount`](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv419OGR_G_GetPointCount12OGRGeometryH)
+    pub fn get_point_vec_zm(&self, out_points: &mut Vec<(f64, f64, f64, f64)>) {
```
lnicola, line 237: I guess these should return an `usize` so it's easier to tell how many points they added.

The changelog needs to be updated too.

JmsPae:
> LGTM with one last nit. No strong opinion on the `i32` / `usize` thing though 🤷.

IMHO it makes more sense to go with `i32` if we're not (for now) making the methods fallible. I don't think the corresponding C functions do anything if you supply an index outside the range of points, which doesn't really give us a reason to throw any Err's unless we want to do any extra checks ourselves.

src/vector/geometry.rs
```diff
@@ -151,9 +203,42 @@ impl Geometry {
...
+        let mut z: c_double = 0.;
+        let mut m: c_double = 0.;
+        unsafe { gdal_sys::OGR_G_GetPointZM(self.c_geometry(), index, &mut x, &mut y, &mut z, &mut m) };
+        (x, y, z, m)
+    }
+
+    /// Appends all points of a line string to `out_points`. 
+    ///
+    /// Only wkbPoint[X] or wkbLineString[X] may alter `out_points`. Other geometry types will silently do nothing, see
+    /// [`OGR_G_GetPointCount`](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv419OGR_G_GetPointCount12OGRGeometryH)
```
lnicola, line 223: It's not documented, but these also work for `wkbCircularString`, see https://github.com/OSGeo/gdal/blob/38253288f042ee983ceb4a9bd78e30cfceacc082/ogr/ogr_api.cpp#L508 and https://github.com/OSGeo/gdal/blob/38253288f042ee983ceb4a9bd78e30cfceacc082/ogr/ogr_api.cpp#L107.

lnicola:
Yeah, the GDAL functions log an error and return 0 (for `GetPoint`), so we'd have to check. But you still have to cast:

```rust
let points = todo!().
for (idx, p) in points.into_iter().enumerate() {
    geom.set_point_2d(idx as i32, p); // ugly, annoying for a new Rust user, and Clippy complains on clippy::pedantic
}
```

It's not too bad, though, more of a paper cut.

JmsPae:
Something _should probably be done_, though it's starting to feel more out of scope for this PR. Otherwise, we can create a new Error type (the IntConversionError is too vague) and make all the get/set methods fallible.

lnicola:
Yeah, okay, let's leave them like this or use `i32` for both (updating the changelog). But then what to do about #581?

JmsPae:
This was probably a larger can of worms than we thought, eh?

I'm going to leave things as is for now and someone (maybe me in the not-so-distant future) can have a crack at a half-way decent solution.

src/vector/geometry.rs
JmsPae: This turned out to be a whole can of worms, see end of PR. Closing for this PR.

JmsPae:
I feel like I have nothing else to add, so feel free to send it whenever.

lnicola:
Yeah, can you please squash everything together?

JmsPae:
There, pretty sure I've squashed everything now

JmsPae:
> That was cute 😂, but for future reference: https://www.baeldung.com/ops/git-squash-commits.

Ye, realized that, apologies.

Pretty sure this was also my first PR so now I know, and knowing is half the battle etc.

lnicola:
> Pretty sure this was also my first PR merge

It was really well-done, I wouldn't have guessed, good job! :grinning:

PR #602, opened 2025-01-10
Allow newer libsqlite3-sys versions

weiznich:
---

Now that there is a `proj-sys` release that supports these newer versions we can support them in gdal-src as well.


lnicola:
Thanks, I've been meaning to do this!

lnicola:
Oh, right https://github.com/georust/gdal/pull/575.

PR #603, opened 2025-01-10
Unsquish after #600

lnicola:
---

### DIFF
--- a/src/vector/geometry.rs
+++ b/src/vector/geometry.rs
@@ -219,11 +219,10 @@ impl Geometry {
         (x, y, z, m)
     }
 
-    /// Appends all points of a line string to `out_points`.
+    /// Appends all points in the geometry to `out_points`, as XYZ.
     ///
-    /// Only wkbPoint[X], wkbLineString[X] or wkbCircularString[X] may alter `out_points`. Other geometry types will silently do nothing, see
-    /// [`OGR_G_GetPointCount`](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv419OGR_G_GetPointCount12OGRGeometryH)
-    pub fn get_point_vec(&self, out_points: &mut Vec<(f64, f64, f64)>) -> usize {
+    /// For some geometry types, like polygons, that don't consist of points, `out_points` will not be modified.
+    pub fn get_points(&self, out_points: &mut Vec<(f64, f64, f64)>) -> usize {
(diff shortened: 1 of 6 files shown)


JmsPae:
Also some minor errors in `CHANGELOG.md`;

- Line 19: get_point_vecx now returns `usize`, not `i32`
- Line 38: Redundant comma and space

lnicola:
I guess `get_point_vec` should be called `get_points` now?

src/vector/geometry.rs
```diff
@@ -213,27 +213,27 @@ impl Geometry {
...
         // [OGR_G_GetPoints](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv415OGR_G_GetPoints12OGRGeometryHPviPviPvi)
         let length = unsafe { gdal_sys::OGR_G_GetPointCount(self.c_geometry()) };
         out_points.extend((0..length).map(|i| self.get_point(i)));
         length as usize
     }
 
-    /// Appends all points of a line string to `out_points`. 
+    /// Appends all points in the geometry to `out_points`, as XYZM.
     ///
-    /// Only wkbPoint[X], wkbLineString[X] or wkbCircularString[X] may alter `out_points`. Other geometry types will silently do nothing, see
```
lnicola, line 234: I rephrased this, since it wasn't exactly correct (MultiPoint has points, for example).

JmsPae (approved):
No complaints here.

JmsPae:
Upon some reflection, however, it was weird for get_points to append to a vec instead of overwriting it, but that can be addressed elsewhere.

lnicola:
I thought so too, but it matches e.g. `BufRead::read_line` and `Read::read_to_end`, so at least there's some prior art.

JmsPae:
Thank you! 

I'd prefer if it follows the original lib's behavior and overwrites the vec, but whether or not you want to fix it here or want to keep it is up to you.

lnicola:
I think appending makes sense in some cases (maybe you're producing something like WKB or another compact format), so I wouldn't mind keeping it for now.

JmsPae:
> I think appending makes sense in some cases (maybe you're producing something like WKB or another compact format), so I wouldn't mind keeping it for now.

Yea, I suppose if we end up 'fixing' it back to overwriting the vec then we can always add another method for appends.

PR #604, opened 2025-01-11
get_points now uses OGR_G_GetPointsX

JmsPae:
---
Previously the binding has simply iterated and appended via `get_point`, which is all well and good, but OGR_G_GetPoints is likely the more performant performant way to go about it. This PR simply replaces the logic of `Geometry::get_pointx` to use OGR_G_GetPointsX.

However, the current logic is certainly 'safer.' Though I'm somewhat confident that this PR is safe across platforms, someone more familiar with the nitty-gritty should take a look. The vague [warning in the GDAL docs](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv415OGR_G_GetPoints12OGRGeometryHPviPviPvi) does not spark confidence.

Something else worth discussion is whether the function should resize & overwrite, or append to the `&mut Vec` input. There was some discussion in [#603](https://github.com/georust/gdal/pull/603#issuecomment-2585191701) and whilst I was personally pro-'do what the library does' and overwrite the `Vec`, this isn't particularly graceful in this PR since the Vec input is being resized anyway, so sticking with appending to the `&mut Vec` might just be more sound.

### DIFF
--- a/src/vector/geometry.rs
+++ b/src/vector/geometry.rs
@@ -1,6 +1,6 @@
 use std::{
     cell::RefCell,
-    ffi::{c_double, c_int},
+    ffi::{c_double, c_int, c_void},
     fmt::{self, Debug, Formatter},
     marker::PhantomData,
     mem::MaybeUninit,
@@ -219,26 +219,60 @@ impl Geometry {
         (x, y, z, m)
     }
 
-    /// Appends all points in the geometry to `out_points`, as XYZ.
+    /// Writes all points in the geometry to `out_points`, as XYZ.
     ///
-    /// For some geometry types, like polygons, that don't consist of points, `out_points` will not be modified.
+    /// For some geometry types, like polygons, that don't consist of points, `out_points` will only be resized, not modified.
     pub fn get_points(&self, out_points: &mut Vec<(f64, f64, f64)>) -> usize {
-        // Consider replacing logic with
-        // [OGR_G_GetPoints](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv415OGR_G_GetPoints12OGRGeometryHPviPviPvi)
-        let length = unsafe { gdal_sys::OGR_G_GetPointCount(self.c_geometry()) };
-        out_points.extend((0..length).map(|i| self.get_point(i)));
-        length as usize
+        let length = unsafe { gdal_sys::OGR_G_GetPointCount(self.c_geometry()) } as usize;
+        out_points.resize(length, Default::default());
+
+        let byte_offset = size_of::<f64>() as isize;
+        let stride = size_of::<(f64, f64, f64)>() as c_int;
+
+        unsafe {
+            let data = out_points.as_mut_ptr() as *mut c_void;
+
+            gdal_sys::OGR_G_GetPoints(
+                self.c_geometry(),
+                data,
+                stride,
+                data.byte_offset(byte_offset),
+                stride,
+                data.byte_offset(byte_offset * 2),
+                stride,
+            );
+        }
+
+        length
     }
 
-    /// Appends all points in the geometry to `out_points`, as XYZM.
+    /// Writes all points in the geometry to `out_points`, as XYZM.
     ///
-    /// For some geometry types, like polygons, that don't consist of points, `out_points` will not be modified.
+    /// For some geometry types, like polygons, that don't consist of points, `out_points` will only be resized, not modified.
     pub fn get_points_zm(&self, out_points: &mut Vec<(f64, f64, f64, f64)>) -> usize {
-        // Consider replacing logic with
-        // [OGR_G_GetPoints](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv415OGR_G_GetPoints12OGRGeometryHPviPviPvi)
-        let length = unsafe { gdal_sys::OGR_G_GetPointCount(self.c_geometry()) };
-        out_points.extend((0..length).map(|i| self.get_point_zm(i)));
-        length as usize
+        let length = unsafe { gdal_sys::OGR_G_GetPointCount(self.c_geometry()) } as usize;
+        out_points.resize(length, Default::default());
+
+        let byte_offset = size_of::<f64>() as isize;
+        let stride = size_of::<(f64, f64, f64, f64)>() as c_int;
+
+        unsafe {
+            let data = out_points.as_mut_ptr() as *mut c_void;
+
+            gdal_sys::OGR_G_GetPointsZM(
+                self.c_geometry(),
+                data,
+                stride,
+                data.byte_offset(byte_offset),
+                stride,
+                data.byte_offset(byte_offset * 2),
+                stride,
+                data.byte_offset(byte_offset * 3),
+                stride,
+            );
+        }
+
+        length
     }
 
     /// Get the geometry type ordinal


JmsPae:
This is also a small bit of groundwork to make harmonizing the index types (as discussed in #600 ) a little less annoying.

lnicola:
> 'do what the library does' and overwrite

I'm not sure I follow. GDAL doesn't know (or care) if you have a `Vec`-like growable buffer or not. You can pass in a pointer to the middle of a buffer and it will just write from that point. So technically, GDAL doesn't resize or clear the destination.

> The vague [warning in the GDAL docs](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv415OGR_G_GetPoints12OGRGeometryHPviPviPvi) does not spark confidence.

Slices are aligned, but the bigger issue is that `(f32, f32, f32)` is not `#[repr(C)]` so we can't pass a `&mut [(f32, f32, f32]` to GDAL. We could do the following and use `&mut [PointXyz]`:

```rust
#[repr(C)]
struct PointXyz(f32, f32, f32);
```

(`&mut [[f32; 3]]` might work too, but I'm actually not sure)

So the simplest would probably be to pass in a `&mut Vec<f32>` and document that the output will be laid out as `XYZXYZXYZXYZ`. But then some user will want `XXXXYYYYZZZZ`, which is only possible if we expose the stride parameters.

Can of :worm:s indeed :grinning:.

JmsPae:
> I'm not sure I follow. GDAL doesn't know (or care) if you have a Vec-like growable buffer or not. You can pass in a pointer to the middle of a buffer and it will just write from that point. So technically, GDAL doesn't resize or clear the destination.

Fair enough.

> Slices are aligned, but the bigger issue is that (f32, f32, f32) is not #[repr(C)] so we can't pass a &mut [(f32, f32, f32] to GDAL.

For example, if we were to implement `OGR_G_SetPoints`?

JmsPae:
We could start adding geometry types (shame `geo` is 2D-only) but that's... A lot :smile:

lnicola:
> For example, if we were to implement OGR_G_SetPoints?

I think both `OGR_G_GetPoints` and `OGR_G_SetPoints` wrappers would look the same, there's no difference between them.

Maybe the simplest is to offer something like:

```rust
geom.get_points(&mut Vec::new(), LayoutXyz::XyzXyz); // there's probably a better name
geom.get_points(&mut Vec::new(), LayoutXyz::XxYyZz);
```

It doesn't offer the full power of the C API, but it's still useful.

JmsPae:
This might actually be a neat way to go about it if we want to avoid adding our own geometry types.

Since `OGR_G_GetPoints` and `OGR_G_GetPointsZM` both simply ignore `null` data, we could use this enum to limit which dims we want to extract and skip a separate `zm` impl.

lnicola:
> this enum to limit which dims we want to extract and skip a separate zm impl

Yeah, I updated my comment about this exactly when you posted yours. We could even do `ZyxZyx` :grimacing:.

> It would, however, still be quite annoying for the end-user to chunk the output to extract individual points

Only a little, right? And there's also [`Itertools::tuples`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.tuples) if you don't mind an extra depedency.

JmsPae:
> Only a little, right?
It certainly beats having to cook up our own `geo_types`, so I guess I'm sold.

For all the `XyzXyz`, `XxYyZz`, `ZzXxYy` etc combinations, a macro might be warranted, but I'll have a gander with some basic variations for now.

lnicola:
In any case, I propose we postpone (no need to close) this for a later time. I'd like to get a new release out of the door soon (maybe after 3.10.1 comes out in the next couple of days).

JmsPae:
Very well, I'll be cooking with gas in the meantime.

JmsPae:
@lnicola reasonably happy with this as a first go at it. Not sure how I'll go about strangely ordered coordinates (if that's even necessary) but feel free to take a peek when you want to.

src/vector/geometry.rs
```diff
@@ -682,21 +899,126 @@ mod tests {
...
+            vec![0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.25, 0.5, 1.0, 2.0, 0.5, 1.0]
+        );
+
+        line.get_points(&mut line_points, CoordinateLayout::XxYyZzMm)
+            .unwrap();
+        assert_eq!(line_points.len(), 3 * 4);
+        assert_eq!(
+            line_points,
+            vec![0.0, 1.0, 1.0, 0.0, 0.0, 2.0, 0.0, 0.25, 0.5, 0.0, 0.5, 1.0]
+        );
```
JmsPae, line 1021: Reasonably sure I'm thorough enough with these test-cases? The pointer manipulation does get a little _freaky_, so idk if more are warranted.

src/vector/geometry.rs
```diff
@@ -9,10 +9,64 @@ use std::{
...
+#[derive(Debug, Clone, Copy)]
+#[repr(u8)]
+pub enum CoordinateComponent {
+    X = 0b00001,
+    Y = 0b00010,
+    Z = 0b00100,
+    M = 0b01000,
+    /// Whether the point data layout is of an AoS structure (Array of Structures, e.g. XyzXyz) as
+    /// opposed to a SoA layout (Structure of Arrays, e.g. XxYyZz.)
+    AoS = 0b10000,
```
JmsPae, line 26: 99% sure this is the right terminology?

src/vector/geometry.rs
```diff
@@ -9,10 +9,64 @@ use std::{
...
+    XxYy = 0b00011,
+
+    XyzXyz = 0b10111,
+    XxYyZz = 0b00111,
+
+    XymXym = 0b11011,
+    XxYyMm = 0b01011,
+
+    XyzmXyzm = 0b11111,
+    XxYyZzMm = 0b01111,
```
JmsPae, line 48: Pretty sure this selection of layouts is sufficient, though a `Custom` variant might be useful?

src/vector/geometry.rs
```diff
@@ -515,6 +713,24 @@ mod tests {
...
+
+        assert!(!CoordinateLayout::XxYy.has_component(CoordinateComponent::AoS));
+
+        assert!(!CoordinateLayout::XymXym.has_component(CoordinateComponent::Z));
+        assert!(CoordinateLayout::XymXym.has_component(CoordinateComponent::M));
+
+        assert_eq!(CoordinateLayout::XyXy.coordinate_size(), 2);
+        assert_eq!(CoordinateLayout::XxYyZz.coordinate_size(), 3);
+        assert_eq!(CoordinateLayout::XyzmXyzm.coordinate_size(), 4);
+    }
```
JmsPae, line 732: Perhaps a little too paranoid, but that's hardly a problem I suppose.

src/vector/geometry.rs
```diff
@@ -219,26 +273,170 @@ impl Geometry {
...
+            return Err(
+                GdalError::InvalidDataInput {
+                    msg: Some(format!(
+                        "in_points length is not a multiple of {coord_size} as determined by '{layout:?}'"
+                    )),
+                    method_name: "set_points"
+                }
+            );
+        }
+        let num_points = in_points.len() / coord_size;
```
JmsPae, line 295: Might want similar checks in `get_points`? GDAL probably checks itself, though.

src/vector/geometry.rs
```diff
@@ -219,26 +273,170 @@ impl Geometry {
...
-    /// For some geometry types, like polygons, that don't consist of points, `out_points` will not be modified.
-    pub fn get_points_zm(&self, out_points: &mut Vec<(f64, f64, f64, f64)>) -> usize {
-        // Consider replacing logic with
-        // [OGR_G_GetPoints](https://gdal.org/en/stable/api/vector_c_api.html#_CPPv415OGR_G_GetPoints12OGRGeometryHPviPviPvi)
-        let length = unsafe { gdal_sys::OGR_G_GetPointCount(self.c_geometry()) };
-        out_points.extend((0..length).map(|i| self.get_point_zm(i)));
-        length as usize
+    /// For some geometry types, like polygons, that don't consist of points, Err will be returned
+    /// and `out_points` will only be resized.
+    pub fn get_points(&self, out_points: &mut Vec<f64>, layout: CoordinateLayout) -> Result<usize> {
```
JmsPae, line 365: Perhaps take `&mut [f64]` instead? Would require additional checks to avoid segv, but maybe letting the user handle resize and whatnot would be better.

JmsPae: I went ahead and implemented this change, though dissenting opinions are welcome.

JmsPae:
Ok now I'm done fiddling (for real this time) so feel free to review @lnicola @weiznich whenever you have time / are feeling bored / 0.18 gets kicked through the door.

I'll update the readme after 0.18 is released.

PR #606, opened 2025-01-13
Bump bundled GDAL to 3.10.1

lnicola:
---

### DIFF
--- a/gdal-src/Cargo.toml
+++ b/gdal-src/Cargo.toml
@@ -1,9 +1,9 @@
 [package]
 name = "gdal-src"
-version = "0.1.0+3.10.0"
+version = "0.1.0+3.10.1"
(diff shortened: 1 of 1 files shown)


lnicola:
@jdroenner how would you feel about putting out a 0.18 release?

lnicola:
@justmytwospence can you say more about your build issues? I don't have any experience with serverless deployments.

The bundled version is really limited, it doesn't have GeoTIFF support. You could take a look at https://github.com/georust/geotiff or https://github.com/pka/georaster for a Rust-native implementation.

lnicola:
I think it would be good to have a "serverless" guide, I wonder if those have a free plan or something.

> only read from files, not memory

`tmpfs` goes brrr.

PR #609, opened 2025-01-20
Update libsqlite3-sys to allow 0.31 as well

weiznich:
---

It won't be used before a new proj release with https://github.com/georust/proj/pull/218 is issued but it shouldn't cause build errors


lnicola:
Thanks. This is actually the same as #608.

PR #611, opened 2025-03-01
Reset layer when droppping feature iterator

### LINKED ISSUE #507: FeatureIterator doesn't reset if the iterator isn't completely read
When looping through a feature iterator, an early break can cause problems later in the program run, because `reset_feature_reading` is never called. The workaround is, to call `reset_feature_reading` after breaking early from such a loop, but failing to do this can cause errors that are difficult to diagnose.

I don't know how this would be fixed, but I think this should at least be documented under `Layer::features()`.

The following is a test I built to reproduce the issue. It's a lot of code, but it takes a few operations before it gets into a situation which results in an error.

This test always fails on my machine, unless I uncomment the code where I've written "FIX IS HERE". Except for a few gdal errors (which are printed to stderr instead of returning results for some reason), the consequences don't occur until far later in the code where I've written "ERROR HAPPENS HERE.".

```rust

#[test]
#[should_panic(expected="create should not return an an error here, but it does for now: OgrError { err: 6, method_name: \"OGR_L_CreateFeature\" }")]
fn test_database_lock_issue() {
    use std::path::PathBuf;

    fn edit_dataset(test_file: PathBuf, finish_loop: bool) {
        let mut dataset = if (&test_file).exists() {
            gdal::Dataset::open_ex(&test_file, gdal::DatasetOptions { 
                open_flags: gdal::GdalOpenFlags::GDAL_OF_UPDATE, 
                ..Default::default()
            }).expect("open dataset")
        } else {
            let driver = gdal::DriverManager::get_driver_by_name("GPKG").expect("get driver");
            driver.create_vector_only(&test_file).expect("create dataset")
        };
    
        let mut transaction = dataset.start_transaction().expect("start transaction");
    
        const RIVERS: &str = "rivers";
        let mut rivers = transaction.create_layer(gdal::LayerOptions {
                    name: RIVERS,
                    ty: gdal_sys::OGRwkbGeometryType::wkbNone,
                    srs: None,
                    options: Some(&["OVERWRITE=YES"])
                }).expect("create layer");
        gdal::vector::LayerAccess::create_defn_fields(&rivers, &[]).expect("define fields");
    
        { // put in a block so we can borrow rivers as mutable again.
            let feature = gdal::vector::Feature::new(gdal::vector::LayerAccess::defn(&rivers)).expect("new feature");
            feature.create(&rivers).expect("create feature");
        }
    
        // I think this is where the problem is...
        for _ in gdal::vector::LayerAccess::features(&mut rivers) {
            // break early...
            if !finish_loop {
                break;
            }
        };
        //gdal::vector::LayerAccess::reset_feature_reading(&mut rivers); // FIX IS HERE!
    
    
        const COASTLINES: &str = "coastlines";
        let coastlines = transaction.create_layer(gdal::LayerOptions {
            name: COASTLINES,
            ty: gdal_sys::OGRwkbGeometryType::wkbPolygon,
            srs: Some(&gdal::spatial_ref::SpatialRef::from_epsg(4326).expect("srs")),
            options: Some(&["OVERWRITE=YES"])
        }).expect("create layer");
    
        gdal::vector::LayerAccess::create_defn_fields(&coastlines, &[]).expect("defn_fields");
    
        { // in a block so transaction can be borrowed again.
            let feature = gdal::vector::Feature::new(gdal::vector::LayerAccess::defn(&coastlines)).expect("new feature");
            // ERROR HAPPENS HERE.
            feature.create(&coastlines).expect("create should not return an an error here, but it does for now");
        }
    
        transaction.commit().expect("commit");
    
        dataset.flush_cache().expect("flush");
    }
    

    let test_file: PathBuf = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("target").join("tmp").join("test_database_locked.gpkg");
    
    // delete the file if it exists so I can rerun the test.
    _ = std::fs::remove_file(test_file.clone()); // ignore error
    edit_dataset(test_file.clone(),false);
    // this one doesn't cause the error, so it's not because the database already exists...
    edit_dataset(test_file.clone(),true);
    // this one will error
    edit_dataset(test_file,false)


}
```

lnicola:
---

Fixes #507

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,7 +36,8 @@
 ### Fixed
 
   - Fix conversion from `ndarray` when the data is offsetted from the start of the buffer ([#569](https://github.com/georust/gdal/pull/569))
-  - use ISO WKT for the `Debug` implementation of `Geometry`, in order to properly display measure values ([#600](https://github.com/georust/gdal/pull/600))
+  - Use ISO WKT for the `Debug` implementation of `Geometry`, in order to properly display measure values ([#600](https://github.com/georust/gdal/pull/600))
+  - Reset layer when dropping the feature iterator, not when starting ([#611](https://github.com/georust/gdal/pull/611))
 
 ### Removed
 
--- a/src/spatial_ref/transform.rs
+++ b/src/spatial_ref/transform.rs
@@ -70,8 +70,8 @@ impl CoordTransform {
     ///
     /// # Arguments
     /// * `bounds` - array of [axis0_min, axis1_min, axis0_max, axis1_max],
-    ///              interpreted in the axis order of the source SpatialRef,
-    ///              typically [xmin, ymin, xmax, ymax]
+    ///   interpreted in the axis order of the source SpatialRef,
+    ///   typically [xmin, ymin, xmax, ymax]
     /// * `densify_pts` - number of points per edge (recommended: 21)
     ///
     /// # Returns
--- a/src/vector/feature.rs
+++ b/src/vector/feature.rs
@@ -708,8 +708,16 @@ impl<'a> FeatureIterator<'a> {
     }
 }
 
+impl Drop for FeatureIterator<'_> {
+    fn drop(&mut self) {
+        unsafe {
+            gdal_sys::OGR_L_ResetReading(self.c_layer);
+        }
+    }
+}
+
 pub struct OwnedFeatureIterator {
-    pub(crate) layer: OwnedLayer,
+    pub(crate) layer: Option<OwnedLayer>,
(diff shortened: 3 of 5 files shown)


lnicola:
@weiznich would you happen to know anything about this `md5_int.h` header?

weiznich:
I've just published pq-src 0.3.4 to address the build failure, so [9099e5a](https://github.com/georust/gdal/pull/611/commits/9099e5ae0106e38ae8bcb837c3fa6ff6ef2fb3c8) can be dropped now.

weiznich (approved):

Issue #514, opened 2024-01-10
Random test suite crash


lnicola:
Just ran into this on #510 (but can't reproduce it easily):

```
Warning 1: GPKG: bad application_id=0x00000000 on '/tmp/.tmpgi6DfV.gpkg'
ERROR 1: Layer new already exists, CreateLayer failed.
Use the layer creation option OVERWRITE=YES to replace it.
test spatial_ref::srs::tests::predicats_epsg2154 ... ok
test spatial_ref::srs::tests::get_units_epsg4326 ... ok
test spatial_ref::srs::tests::non_existing_proj_param ... ok
test vector::layer::tests::test_create_layer_options ... ok
error: test failed, to rerun pass `--lib`

Caused by:
  process didn't exit successfully: `/home/grayshade/Projects/georust/gdal/target/debug/deps/gdal-0978520db0adc3dd` (signal: 11, SIGSEGV: invalid memory reference)
```

That looks like a temporary file name, not sure what's going on.

Atreyagaurav:
Oh yeah, I ran into it once as well. It passed the next time I ran it, though. Maybe, like it said, the temp filename was coincidently the same as last time the test was done, and the file isn't cleared?

lnicola:
No, they're checked for collisions.

Not saying it's because of your changes, I just didn't see it before.

metasim:
Been a while since I've seen this. I'd start with checking to ensure any test that mutates an on-disk fixture is using the `TempFixture` tooling, so as to work on separate copies of the file.

lnicola:
Hmm, I was able to reproduce the crash on `master`, so it's not #510. The "Layer new already exists" message points to the `poly.gpkg` tests, which we should fix.

However, running `cargo test vector` in a loop doesn't seem to crash, so it's still possible that the SIGSEGV is unrelated.

lnicola:
```
#0  0x00007ffff768dfcb in OGR2SQLITEModule::Setup(sqlite3*) () from /lib64/libgdal.so.33
#1  0x00007ffff768e7d5 in OGR2SQLITE_Setup(GDALDataset*, OGRSQLiteDataSource*) () from /lib64/libgdal.so.33
#2  0x00007ffff797d9fb in OGRSQLiteDataSource::OpenOrCreateDB(int, bool) [clone .constprop.0] () from /lib64/libgdal.so.33
[...]

PR #612, opened 2025-03-06
Bump GDAL to 3.10.2

lnicola:
---


lnicola:
```
called `Result::unwrap()` on an `Err` value: NullPointer { method_name: "GDALOpenEx", msg: "OGR2SQLITE_Setup() failed due to sqlite3_api == nullptr" }
```

#514 again, and it's in 3.8.5, so https://github.com/OSGeo/gdal/pull/9077 didn't fix it.

PR #613, opened 2025-03-07
Bump libsqlite3-sys to support version 0.32 as well

weiznich:
---


lnicola:
Thanks!

PR #614, opened 2025-03-25
Switch to doc_auto_cfg and fix some warnings

lnicola:
---

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,7 +1,7 @@
 #![crate_name = "gdal"]
 #![crate_type = "lib"]
-// Enable `doc_cfg` features when `docsrs` is defined by docs.rs config
-#![cfg_attr(docsrs, feature(doc_cfg))]
+// Document all features on docs.rs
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -1,7 +1,7 @@
 #![crate_name = "gdal"]
 #![crate_type = "lib"]
-// Enable `doc_cfg` features when `docsrs` is defined by docs.rs config
-#![cfg_attr(docsrs, feature(doc_cfg))]
+// Document all features on docs.rs
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
```
lnicola, line 4: Like this comment if you lost almost one hour because of `RUSTFLAGS` vs. `RUSTDOCFLAGS`.

lnicola:
Bah, https://github.com/georust/gdal/actions/runs/13726353606/job/38393586016 worked, https://github.com/georust/gdal/actions/runs/14066687910/job/39390852293 failed.

We should publish our lockfile, I guess.

PR #616, opened 2025-03-28
bump sys and src versions

jdroenner:
---

### DIFF
--- a/gdal-src/Cargo.toml
+++ b/gdal-src/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gdal-src"
-version = "0.2.0"
+version = "0.2.0+3.10.2"
 edition = "2021"
 links = "gdal_src"
 description = "Build script for compiling GDAL from source."
--- a/gdal-src/Cargo.toml
+++ b/gdal-src/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gdal-src"
-version = "0.1.0+3.10.2"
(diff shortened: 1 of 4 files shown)


gdal-src/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "gdal-src"
-version = "0.1.0+3.10.2"
+version = "0.2.0"
```
lnicola, line 3: I think we can keep this as `0.2.0+3.10.2`. I forgot that we already had (I think?) an empty version published.

lnicola:
Ouch, I forgot about `gdal-sys`!

gdal-src/Cargo.toml
jdroenner: the last gdal-src is just 0.1.0 see https://crates.io/crates/gdal-src/versions

lnicola: Yeah, you published it so we could ask for a crate size exemption. We can do `0.2.0+3.10.2`, `0.3.0+3.11.0` and so onn.

jdroenner: cargo release keeps complaining about the version but we can try it. Maybe it is just a dry-run issue.

lnicola: Does it warn that the extra part is useless when you try to publish the crate? I think that's right, but harmless.

lnicola (approved):

PR #632, opened 2025-04-09
Include required files in published gdal-src package

weiznich:
This commit includes the required files into the gdal-src package. It also adds a CI job step to verify that everything is there on all supported platforms.

---


lnicola:
Can we run this only on `macos-14`?

image

weiznich:
We can but we would loose CI checks for x86_64 macos as the macos-13 builders are the last x86 builders.

PR #645, opened 2025-05-30
Make DatasetCapability accessible

Atreyagaurav:
---

Sorry for not noticing this, I think I added this feature (?), but the struct is not accessible publicly. This fixes it

https://docs.rs/gdal/latest/gdal/struct.Dataset.html#method.has_capability

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -125,7 +125,7 @@ pub mod vector;
 pub mod version;
 pub mod vsi;
 
-pub use dataset::Dataset;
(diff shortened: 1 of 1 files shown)


Atreyagaurav:
@lnicola the test failure seems to be about the CI pipeline.

This is a tiny change. The tests should be fine, please review and merge when you can.

This is a fix related to #585

lnicola:
Looks like https://github.com/actions/runner-images/issues/5656?

lnicola (approved):

PR #646, opened 2025-06-02
Expose `bundled` feature on gdal crate

yutannihilation:
---

Hi! I was struggling to statically build a CLI using the gdal crate, and found the CI in this repo uses this trick:

https://github.com/georust/gdal/blob/e711d7665f25e7b3644086be66ddb21008d01cef/.github/workflows/ci.yml#L160-L163

However, this is possible only when gdal-sys and gdal-src are the direct dependency; using the gdal crate is not the case. I think it makes sense to expose `bundled` feature on the gdal crate as well.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+  - Add `bundled` and `bundled_all` feature to the gdal crate for building and statically linking a minimal bundled version of GDAL.
+
 ### Fixed
 
   - Bump referenced gdal version to 3.10.3


nokonoko1203:
@yutannihilation ありがとうございます！！！！！！

nokonoko1203 (approved):

weiznich:
I'm personally not a friend of exposing features from dependencies in such a way as it adds additional constraints on what you can/cannot change as crate without a major release. Users can always opt in such features by directly depending on the crate exposing the feature as cargo will unify these features across multiple sources. Yes that's a bit more work for users, but I think that is manageable on their and.
In this particular case users cannot have ever more than one version of `gdal-sys` in their dependency tree, so the worst thing that happens is that they get a compilation error stating that they have multiple versions of that crate in their dependency tree.

Then there are one general note about `bundled` features like this: While they are really nice for certain use cases, they are a really bad for other uses cases. It's nice to ship statically linked binaries if you provide pre-build binaries for whatever reason. On the other hand if your application get shipped by some linux distribution they will very likely hate you for unconditionally statically linking such dependencies as they want to be able to dynamically link them from their canonical distro-package instead.

Finally there is one gdal specific consideration here: I would guess that the `gdal-sys/bundled` feature on it's on is not useful for most of the users as it doesn't enable any import/export drivers. On the other hand the proposed `bundled_all` feature seems to enable really all features which is again something I would guess is not the thing most projects want, especially given the implications around licensing and linking to libgeos. For me that sounds like most users rather want to enable drivers they depend on explicitly to get a fitting configuration for their use-cases. If that all should be configurable from the `gdal` crate itself it would require exposing a lot of feature flags. I'm not a maintainer of `gdal`, but I wouldn't do that to keep the amount of required work on the maintainer side as low as possible. If you don't duplicate the feature flags users would need to depend on `gdal-sys`/`gdal-src` anyway so there is really not much sense in exposing these features at all from `gdal` itself.

What can and should be done is documenting all of that. Unfortunately I'm currently rather busy with a lot of other stuff so I'm not sure if I can promise contributions in the next weeks.

yutannihilation:
Ah, makes sense, thanks for the details.

> I would guess that the `gdal-sys/bundled` feature on it's on is not useful for most of the users as it doesn't enable any import/export drivers.

In our case, we don't need import/export, so that's totally fine! But, I agree it might be confusing to users. Actually, I wasn't aware of this.

> For me that sounds like most users rather want to enable drivers they depend on explicitly to get a fitting configuration for their use-cases. If that all should be configurable from the `gdal` crate itself it would require exposing a lot of feature flags. I'm not a maintainer of `gdal`, but I wouldn't do that to keep the amount of required work on the maintainer side as low as possible.

Agreed. As a user, I want such fine-grained feature flags, but, if I were a maintainer, I would feel it's a nightmare.

So, should I close this pull request?

weiznich:
@yutannihilation I'm not a maintainer here, but I personally would suggest to change the scope of the PR to add the relevant documentation instead. Maybe just a crate level documentation entry [here](https://github.com/georust/gdal/blob/e711d7665f25e7b3644086be66ddb21008d01cef/src/lib.rs#L7-L104) that points to the `bundled` feature of `gdal-sys` and mentions the considerations around enabling specific drivers? Possibly with a link to [the relevant section](https://github.com/georust/gdal/blob/e711d7665f25e7b3644086be66ddb21008d01cef/gdal-src/Cargo.toml#L77-L380) of the `gdal-src/Cargo.toml` file?

yutannihilation:
Hm, thanks for the suggestion, but I don't think I'm the right person to write the documents, considering my knowledge is very limited.

After some thinking, while I get your concern, I still feel this pull request might be valid to some extent. So, I'll leave this pull request and wait for the maintainer to decide.

PR #666, opened 2025-09-15
Stop linking to Wbemuuid to see what happens

lnicola:
---


weiznich:
We build a static library via `gdal-src`, which shouldn't have all these final link arguments as far as I understand. These need to be provided for the "final" linkage, so via rustc. That means we still need to have them explicitly either in the gdal-src build.rs file or pull them in via a dependency.

lnicola:
Right, that makes sense for static libraries, thanks!

PR #674, opened 2025-10-02
Cross-compile for x86_64-apple-darwin

lnicola:
---


lnicola:
r? @weiznich 

How attached are you to the `x86_64-apple-darwin` target?

This stops running the tests (e.g. `cargo test --features "gdal-sys/bundled gdal-src gdal-src/all_drivers gdal-src/geos_static"`) on `x86_64-apple-darwin` because of OpenSSL. We could probably add an `openssl` dependency and enable `vendored` but I'm not sure it's worth it.

lnicola:
Oh, great, I didn't realize they walked back on that.

PR #677, opened 2025-10-04
Add `Dataset::try_into_thread_safe`

lnicola:
---

### DIFF
--- a/src/dataset.rs
+++ b/src/dataset.rs
@@ -7,7 +7,7 @@ use std::{
 use gdal_sys::{CPLErr, GDALDatasetH, GDALMajorObjectH};
 
 use crate::cpl::CslStringList;
-use crate::errors::*;
+use crate::errors::{GdalError, Result};
 use crate::options::DatasetOptions;
 use crate::raster::RasterCreationOptions;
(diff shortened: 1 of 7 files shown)


src/dataset.rs
```diff
@@ -378,6 +446,83 @@ impl Drop for Dataset {
...
+}
+
+#[cfg(any(all(major_ge_3, minor_ge_10), major_ge_4))]
+impl AsRef<Dataset> for ThreadSafeDataset {
+    fn as_ref(&self) -> &Dataset {
+        &self.inner
+    }
+}
+
+#[cfg(any(all(major_ge_3, minor_ge_10), major_ge_4))]
```
ChristianBeilschmidt, line 517: For style purposes: Could we make a gated module and impl this in this module? Then, we would only have to import/re-export it once in `dataset.rs` and get rid of 90% of the version gates?

ChristianBeilschmidt (changes requested):
Great idea, I have just one style remark.

src/dataset.rs
lnicola: Done.

lnicola: @ChristianBeilschmidt can you take another look? Thanks!

ChristianBeilschmidt (approved):

PR #724, opened 2026-07-04
Update bundled gdal version to 3.13.1

lnicola:
---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,10 +121,11 @@ jobs:
         os:
           - ubuntu-latest
           - windows-latest
-          - macos-15-intel # x86_64
+#          - macos-15-intel # x86_64
           - macos-15 # aarch64
     env:
       CMAKE_POLICY_VERSION_MINIMUM: 3.5
+      MACOSX_DEPLOYMENT_TARGET: 15.0
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout code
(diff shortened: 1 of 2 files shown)


lnicola:
CC @rouault if you update your CI to `macos-15-intel` and get something like:

```
Undefined symbols for architecture x86_64:
  "___cpu_model", referenced from:
      CPLHaveRuntimeAVX2Initialize() in libgdal_src-debf9667c3c99f3b.rlib[83](cpl_cpu_features.cpp.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

It's probably a regression from GDAL 3.12 to 3.13, but I couldn't figure it out.

@metasim @weiznich if you care about this target, please try to bring it back.

rouault:
> It's probably a regression from GDAL 3.12 to 3.13, but I couldn't figure it out.

yes, that's new code of 3.13. We do have a Conda CI build for macos-15-intel.  In https://github.com/georust/gdal/actions/runs/28704314542/job/85127428904  , I see "built for newer 'macOS' version (15.5) than being linked (10.12)".   There might be a possibility that this mismatch is the cause: the compiler expects the ``___cpu_model`` symbol (likely used for the implementation of ``__builtin_cpu_supports("avx2")``) to be available in some standard library that might be missing in 10.12

lnicola:
I tried to set `MACOSX_DEPLOYMENT_TARGET` to 14 and it fixed those warnings about 10.12, but the `___cpu_model` one remained. Maybe some clang/gcc thing?

weiznich:
Thanks for the headsup. We don't use the MacOS x86_64 target anymore, so it's fine to drop support for it from our perspective.

PR #731, opened 2026-09-08
feat: safe wrappers for ignored fields and the PROJ search paths

PlenoraETL:
Two C functions that `gdal-sys` already binds have no safe wrapper in `gdal`. This adds them — **no new FFI**, the symbols are already in the pre-built bindings for every supported version.

## `Layer::set_ignored_fields`

Wraps `OGR_L_SetIgnoredFields`, which lets a driver skip columns it would otherwise read.

The test asserts the **effect** — the field is present, absent while ignored, and present again once the list is cleared — rather than the call returning `Ok`.

It builds a GeoPackage instead of using a checked-in fixture, and that choice is load-bearing: `SetIgnoredFields` is advisory and drivers may ignore it. The GeoJSON driver does. I wrote the test against `roads.geojson` first and it passed whether or not the wrapper did anything, which would have made it worthless.

## `set_proj_search_paths` / `get_proj_search_paths`

Wrap `OSRSetPROJSearchPaths` and `OSRGetPROJSearchPaths`.

These are not the same thing as setting the `PROJ_DATA` configuration option: GDAL does not forward that option to PROJ, so a relocatable distribution that only sets the option keeps reading whatever paths were compiled into PROJ at build time. That is the problem we hit, and this is what fixed it.

The getter matters on its own: with nothing configured it reports the defaults PROJ was built with, which is how you learn where the resource files actually live before overriding them.

The paths are process-global, so the test restores whatever it found and runs behind a mutex — two tests writing a global concurrently would make each other flaky rather than fail honestly.

## Provenance and testing

Both wrappers have been in use in production in [plenora-IO-tools](https://github.com/PlenoraETL/plenora-IO-tools), a GDAL-backed reader, for some months; this is the same code, ported to `master`.

Verified against GDAL 3.8.4 with the pre-built bindings (no `bindgen`): `cargo test -p gdal` passes — 281 / 1 / 1 / 57 — and clippy and `fmt --check` are clean.

One thing worth flagging: `get_proj_search_paths` needs `_string_array`, which `utils.rs` already exports but `srs.rs` did not import, so the `use` line grows by one name.

Happy to split this into two PRs if you'd rather review them separately — they're independent.

### DIFF
--- a/src/spatial_ref/srs.rs
+++ b/src/spatial_ref/srs.rs
@@ -7,7 +7,7 @@ use std::{
 use gdal_sys::{OGRErr, OSRAxisMappingStrategy};
 
 use crate::errors::*;
-use crate::utils::{_last_null_pointer_err, _string};
+use crate::utils::{_last_null_pointer_err, _string, _string_array};
 
 /// A OpenGIS Spatial Reference System definition.
 ///
@@ -636,6 +636,32 @@ impl TryFrom<u32> for AxisMappingStrategy {
 
 #[cfg(test)]
 mod tests {
+    /// Setting the PROJ search paths replaces them, and reading them back
+    /// reports what was set.
+    ///
+    /// The paths are **process-global**: the test restores whatever was there
+    /// before, so it does not move the ground under the other tests. It also
+    /// runs behind a mutex, because two tests writing a global at the same time
+    /// would make each other flaky rather than fail honestly.
+    #[test]
+    fn test_proj_search_paths() {
+        use std::sync::Mutex;
+        static GUARD: Mutex<()> = Mutex::new(());
+        let _held = GUARD.lock().unwrap_or_else(|e| e.into_inner());
+
+        let original = super::get_proj_search_paths();
+
+        super::set_proj_search_paths(&["/tmp/proj-one", "/tmp/proj-two"]).unwrap();
+        assert_eq!(
+            super::get_proj_search_paths(),
+            vec!["/tmp/proj-one".to_string(), "/tmp/proj-two".to_string()]
+        );
+
+        let restore: Vec<&str> = original.iter().map(String::as_str).collect();
+        super::set_proj_search_paths(&restore).unwrap();
+        assert_eq!(super::get_proj_search_paths(), original);
+    }
+
     use super::*;
     use crate::assert_almost_eq;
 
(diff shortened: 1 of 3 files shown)


src/vector/layer.rs
```diff
@@ -445,6 +445,34 @@ pub trait LayerAccess: Sized {
         }
     }
 
+    /// Select fields that the driver may omit while reading features.
+    ///
+    /// This wraps [`OGR_L_SetIgnoredFields`], which is already bound in
```
lnicola, line 450: Drop "This wraps [`OGR_L_SetIgnoredFields`], which is already bound in `gdal-sys` but has no safe wrapper here.  Field names are validated as C strings and retained until the call returns; GDAL copies the null-terminated list into the layer state.", follow the convention. "See: ..." should be enough.

src/spatial_ref/srs.rs
```diff
@@ -636,6 +636,32 @@ impl TryFrom<u32> for AxisMappingStrategy {
 
 #[cfg(test)]
 mod tests {
+    /// Setting the PROJ search paths replaces them, and reading them back
```
lnicola, line 678: These tests tend to interact badly with the others, we should probably drop it or turn it into an integration test.

src/vector/layer.rs
```diff
@@ -747,6 +775,73 @@ mod tests {
     use crate::{assert_almost_eq, Dataset, DriverManager, GdalOpenFlags};
     use gdal_sys::OGRwkbGeometryType;
 
+    /// Ignoring a field keeps it out of the features that are read.
```
lnicola: Drop the doc comment, move the test after the helper functions.

src/spatial_ref/srs.rs
```diff
@@ -972,3 +998,42 @@ mod tests {
...
+        .iter()
+        .map(|p| CString::new(p.as_bytes()))
+        .collect::<std::result::Result<Vec<_>, _>>()?;
+    let mut raw: Vec<*const std::ffi::c_char> = owned.iter().map(|s| s.as_ptr()).collect();
+    raw.push(std::ptr::null());
+    unsafe { gdal_sys::OSRSetPROJSearchPaths(raw.as_ptr()) };
+    Ok(())
+}
+
+/// The search path(s) PROJ currently uses, as GDAL sees them.
```
lnicola: Trim down the doc comment, add a "See: " remark with a link to the `OSRGetPROJSearchPaths` docs.

src/spatial_ref/srs.rs
```diff
@@ -972,3 +998,42 @@ mod tests {
         );
     }
 }
+
+/// Set the search path(s) PROJ uses to find its resource files.
```
lnicola: Same as with `set_ignored_fields`, trim down the comment, add "See".

And move the functions out of the `tests` submodule.

